PR #10, opened 2019-08-27
Add missing header

olemke:
While testing clang-format, which reorders includes alphabetically, it became apparent that several of our headers are not self-contained. Not sure how to test for this globally.

### DIFF
--- a/src/constants.h
+++ b/src/constants.h
@@ -54,6 +54,7 @@
 #ifndef CONSTANTS_IN_ARTS_H
 #define CONSTANTS_IN_ARTS_H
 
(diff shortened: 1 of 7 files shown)


riclarsson:
About "matpack.h" in constants.h.  Using any other source for Numeric and Complex than matpack to define Numeric and Complex might be a good idea. This is all that I used of matpack in those files.

olemke:
@riclarsson Since constants.h currently only uses Numeric, not Complex, I'd prefer to stick with the minimal approach to only include what's needed.

PR #11, opened 2019-08-28
Remove unused old code and replace with small optimization

riclarsson:
This makes better use of LineShape::Model features in xsec_species2 and zeeman.cc.  I do not want to touch xsec_species to use the optimization.

There is one problem here.  I am unsure if the lines where the old model is selected are taking copies.  It does not need to take a copy, so it would be good if it didn't.  These two lines are at 224 in zeeman.cc and at 2235 in absorption.cc.

Also, can someone explain why github puts my name on Oliver's code fixes?  It is very annoying that it does this and I would like it to stop acting this way.  I hope it does not affect the history...


olemke:
About the order of the commits: It looks like you're not specifying `--rebase` when you pull in changes from upstream. I checked out your master branch and ran `git pull --rebase upstream master` which brings the commits in the right order. So you can do that on your end as well and then do a force push `git push -f origin master`. See section "Update your working copy" in [CONTRIBUTING.md](https://github.com/atmtools/arts/blob/master/CONTRIBUTING.md).

olemke:
About the copying: Yes, in the current code `Line::Model` is copied. Although the return value of `LineRecord::GetLineShapeModel()` is a reference, the variable `line_shape_model` it is assigned to is not. This forces copying.

I propose to change `line_shape_model` to a reference. But then you also need to initialize it. Something like this:

```cpp
LineShape::Model line_shape_default_model;
LineShape::Model& line_shape_model{line_shape_default_model};
```

PR #16, opened 2019-08-30
Simplify quantum number encoding

riclarsson:
This makes the quantum number encoding easier to handle and constexpr.  As I have suggested elsewhere, it might be good to have a wider rework here.  It takes 1024 bits per line to just store the Quantum numbers presently.  This is the largest single part of a LineRecord.  If ArrayOfLineRecord was a class it would be possible to use a lot less data here, since almost none of the quantum numbers are used in deep calculations.


olemke:
We should pick up the discussion around ArrayOfLineRecord again as there seems to be room for improvement there. But let's not do it in this pull request. How about we start an issue to track this topic and list the advantages that are to gain by replacing ArrayOfLineRecord with a, let's call it ArtsCatalog, class. To get an overview it would help as a first step to list which data is currently stored redundant with each line although it only changes by band or by species.

PR #21, opened 2019-09-02
Monday morning ARTS week contribution

riclarsson:
These are my Monday morning changes to ARTS for cleanup week.  Mostly adding documentation.  Removed some files.


src/linefunctiondata.h
```diff
@@ -37,19 +37,40 @@
 #include <numeric>
 #include "constants.h"
 
-//! Select the derivative that will be used in Jacobian calculations --- also checks validity of var and coeff
+/** Return the deriveative type based on string input
```
olemke: deriveative -> derivative

PR #26, opened 2019-09-02
Review of headers in files for ppath and PSDs.

erikssonpatrick:


src/m_ppath.cc
```diff
@@ -59,7 +52,7 @@ extern const Numeric DEG2RAD;
   === The functions (in alphabetical order)
   ===========================================================================*/
 
-/* Workspace method: Doxygen documentation will be auto-generated */
+/** Workspace method: Doxygen documentation will be auto-generated */
```
olemke: These should all be normal comments for Workspace methods. The doxygen documentation is automatically generated in auto_md.h. If you use a '/**.../' here, it will completely hide the automatically generated documentation.

erikssonpatrick: Oh, shit. OK, I will fix that. I was about to make a new pull request. Seems easiest that I fix this and include it the new pull. Can you cancel this request?

PR #34, opened 2019-09-03
Header cleaning in ppath and rt files

erikssonpatrick:
Also removed some files and functions.


.gitignore
```diff
@@ -14,4 +14,4 @@
 heaptrack.*.gz
 3rdparty/wigner/fastwigxj/gen/
 3rdparty/wigner/fastwigxj/obj/
-
+cmake-build-debug
```
simonpf: Is this intended?

erikssonpatrick: Yes, that was on purpose. That folder showed up for me and it should not be committed.  Missed to comment on it.

PR #43, opened 2019-09-04
More doxygen updates, move Faddeeva to 3rdparty

riclarsson:


src/m_linemixing.cc
```diff
@@ -2964,7 +2964,7 @@ void abs_lines_per_bandSetLineMixingFromRelmat(
 }
 
 #include <Eigen/Eigenvalues>
-#include "Faddeeva.hh"
+#include "../3rdparty/Faddeeva/Faddeeva.hh"
 #include "physics_funcs.h"
```
olemke: Please move the include statement to the top of the file.

olemke:
Please use branches in the future. When you push new commits to your fork while the previous pull request hasn't been merged yet, they are added to the pull request.
Starting a new branch for your Faddeeva changes would have been much cleaner. Now I have to delay the documentation changes until we have sorted out the Faddeeva issues.

src/m_linemixing.cc
olemke: Also replace the relative include path by:

```cpp
include <Faddeeva/Faddeeva.hh>
```

This is better for treating Faddeeva as a 3rd party library.

riclarsson: Done.

Issue #41, opened 2019-09-04
Finish iyActiveSingleScat2


erikssonpatrick:
Richard: I have created the first version of a test targeting Jacobians by iyActiveSingleScat. Presently it just checks that things run. The test runs with iyActiveSingleScat but crashes with iyActiveSingleScat2. 

The test is called TestIyActive_wfuns.arts

riclarsson:
Note, pull request #46 is not using the naive reflection routine you had implemented for derivatives Patrick.  Instead, it is looking directly at the pure forward equation and defines the derivative from it.  This might be why there are differences.  At this point though, I have done my part of this problem and will unassign myself from further work on this.

erikssonpatrick:
New goal of issue: Check iyActiveSingleScat2 and when done make this iyActiveSingleScat

erikssonpatrick:
Replaced by issue #205

PR #46, opened 2019-09-05
Partial fix for issue #41 (Finish iyActiveSingleScat2)

riclarsson:
@erikssonpatrick This makes iyActiveSingleScat2 produce the same forward results as iyActiveSingleScat.  The Jacobian output is very different though.

I am unsure how to read the Jacobian.  I think there are issues in my implementation of the backscatter solver that makes it differ from your implementation.  An easy way to play with this is to implement a new solver in set_backscatter_radiation_vector.  I am using my BackscatterSolver::Full implementation in this code but if I switch to BackscatterSolver::Commutative_PureReflectionJacobian the differences are much smaller between the two codes.

Still, in either case, your implementation have relatively small trailing values after the larger parts.  I cannot reproduce these trailers.  It is relatively simple to implement a new solver if you can spot the problem in set_backscatter_radiation_vector.  Just add an enum to BackscatterSolver and implement it in set_backscatter_radiation_vector.

Two more things: 1) The particulate derivative implementation does not follow the same logic as the LBL derivatives.  If a derivative is zero, I set them to zero.  Your derivatives are instead empty() matrices.  2)  The auxiliary values are not stored.  This needs to be re-added before removing the other function but is not so important before we understand the diff between the Jacobian.

### DIFF
--- a/src/m_cloudradar.cc
+++ b/src/m_cloudradar.cc
@@ -1155,12 +1155,10 @@ void iyActiveSingleScat2(Workspace& ws,
       }
     }
   }
-
   // FIXME: Add the aux-variables back
-  // FIXME: Add the diy_dpath
 
   // Finalize analytical Jacobian
-  if (j_analytical_do and false)  // FIXME: remove false
+  if (j_analytical_do)
     rtmethods_jacobian_finalisation(ws,
                                     diy_dx,
                                     diy_dpath,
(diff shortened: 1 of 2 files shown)


erikssonpatrick:
Seems easiest to accept the pull request, and I look at it when I have free time. Issue not closed, but still a step forward.

PR #47, opened 2019-09-05
Fix to ARTSCAT-4 new reading by allowing to return 0-vmrs

riclarsson:
The old problem seems to be that ARTSCAT-4 allowe 0-vmrs.
The new code considered this bad and generated non-normal
values.  Now instead the 0-vector is accepted as a plausible
output.

Also move LineShape::Model::vmrs implementation to .cc file

Edit:  This should fix the problem reported by #31

### DIFF
--- a/src/linefunctiondata.cc
+++ b/src/linefunctiondata.cc
@@ -91,25 +91,20 @@ JacPropMatType select_derivativeLineShape(const String& var,
 std::istream& LineShape::from_artscat4(std::istream& is,
                                        Model& m,
                                        const QuantumIdentifier& qid) {
-  // Special case when self is part of this
-  auto self_in_list = LegacyPressureBroadeningData::self_listed(
-      qid, LegacyPressureBroadeningData::TypePB::PB_PLANETARY_BROADENING);
-  auto i = self_in_list ? 1 : 0;
-
   // Set or reset variables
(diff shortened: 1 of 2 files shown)


src/linefunctiondata.cc
```diff
@@ -91,55 +91,59 @@ JacPropMatType select_derivativeLineShape(const String& var,
...
+  m.mspecies = ArrayOfSpeciesTag(7);
 
   // Set species
-  m.mspecies[i] = SpeciesTag(String("N2"));
-  m.mspecies[i + 1] = SpeciesTag(String("O2"));
-  m.mspecies[i + 2] = SpeciesTag(String("H2O"));
-  m.mspecies[i + 3] = SpeciesTag(String("CO2"));
-  m.mspecies[i + 4] = SpeciesTag(String("H2"));
-  m.mspecies[i + 5] = SpeciesTag(String("He"));
+  m.mspecies[1] = SpeciesTag("N2");
```
olemke: Is it correct that element 0 is not set?

riclarsson: Yes.  Element "0" is "SELF" broadening in ARTSCAT-4.  When mself is true, then element "0" is not named in LineShape::Model.  The same for element "-1" when "AIR" broadening mbath is true.

PR #50, opened 2019-09-05
Rename linefunctiondata

riclarsson:
This renames line function data


riclarsson:
Why did this not work?  One is a bug fix in a branch, the other is a rename in a different branch.  I believe the branch system is bad to use...

olemke:
It did work as expected, mostly. ;-) Because you based your second branch of the first one, it also contained the artscat-4 fix. This is expected. And would have not caused any issue. But, because you pulled in new commits from upstream/master into the second branch, the rebase option we use by default, caused the commit id of the artscat-4 fix to change (because rebase reorders the commits). This caused the automatic merge to fail. My fix was to remove the offending commit from this branch and merge it with master manually to make github happy again.
Unfortunate that it happend to you, but an important discovery to be aware of.

PR #59, opened 2019-09-06
Added OEM chapter to userguide.

simonpf:


erikssonpatrick:
Simon: Is there a way for us to generate the PDF? Easier to check the equations then. Or can you send us the pages from your version?

erikssonpatrick:
One comment already now. Note that the AUG has an index. WSVs and WSMs are added to the index as: \wsvindex{y} and \wsmindex{OEM}.
I suggest that you use these commands the first time you mention a variable or method inside the chapter. 

You could argue that you can search in the pdf, but not easy to search where the WSV y is discussed. Will give many hits!

olemke:
> Simon: Is there a way for us to generate the PDF? Easier to check the equations then. Or can you send us the pages from your version?

@erikssonpatrick Use the @username notation when addressing someone directly to make sure the person gets a notification.

olemke:
You can switch to Simon's branch in your own working copy with the following steps:

1. Make sure you don't have any uncommitted changes in your current branch.

2. Add Simon's repository as a remote. Fetch the current state of his repo. Then checkout the branch from this pull request:

```
git remote add simon https://github.com/simonpf/arts.git
git fetch simon
git checkout -b simon-oem-userguide simon/oem_userguide
```
3. Once you're done, go back to your own master branch. If you were working in any other branch before, replace `master` with your working branch name instead:

```git checkout master```

erikssonpatrick:
@simonpf Followed Oliver's advice. FYI, AUG did not compile. Seems that you have missed adding the oem_flow figure.

olemke:
I plan to add a Travis build job soon that also covers the user guide to spot such problems in the PR immediately.

erikssonpatrick:
@simonpf Now compiles. Looks fine. Just one comment. Consider if the paragraph around eq 21.3/4, about n and m-form, is needed. Are these forms actually used inside OEM?
Feel free to add you as an author in the table on page 3. If you do it, please update our affiliation with the new name of the department. 
OK to this pull request from me.

simonpf:
Thanks for the feedback. Both standard and m-form can be chosen via the method string, so I felt it was good to explain what they mean.

stefanbuehler:
Hi @simonpf!

I very much like your OEM parts, and also the new text.I think the text is on the right level, giving a high-level overview without going into too much detail.

Some comments:

Don't forget to add your name to the contributers table on page 3.

Trivial text fix:
"After the retrieval definition block can be finalized" --> "After the retrieval definition block has been finalized"

Just for my curiosity: Why are there dedicated methods to set the x and y covariance matrices? If there is a simple explanation, consider adding a sentence in "Defining the state space" to explain. (I did look in the built-in documentation of covmat_sx and covmat_sy, but it's not explained there either.)

On Equation 21.2:
You write below the equation that H has the dimension of the measurement space, but it appears to have the dimension of the state space.

On Equatione 21.3 and 21.4:
These are the equations for completely linear retrieval (no xi on the right side), not for GN-itereration, as you claim. Also, I suspect the K in the last term of 21.3 is wrong and should be Sx, so that 21.3 actually is identical to 21.2 if you make it iterative.
To summarize, I think there should be only two equations in this section, 21.2, which is the n-form, and the corresponding iterative m-form equation. (Unless there is a mathematical reason why 21.4 can not brought to a proper iterative form resembling 21.2.)
Also, the text should be clear about which methods can actually be used in ARTS. Is it possible to choose between n form and m form? Or is this choice made automatically?

Cheers,

Stefan

stefanbuehler:
I must have gotten it wrong with the equations. But then I still don’t understand why you have three equations, shouldn’t there just be two, an n-form and an m-form?

simonpf:
No, in the book there are three forms. The what are called the n and m forms are written as a difference w.r.t. xa, whereas the standard form computes a difference from the current iteration step.  But I haven't enountered an application yet where it would make a critical difference which one of n or standard form is applied.

olemke:
@simonpf, @stefanbuehler Is this ready to be merged?

Issue #55, opened 2019-09-06
Wind Jacobian


erikssonpatrick:
This Jacobian is now zero for zero wind? Formerly correct? If not, find another way to do the calculations. If yes, find a practical workaround.

riclarsson:
The question is poorly formulated.  The wind Jacobian is not zero in ARTS for zero wind.  The component-wise wind Jacobian is zero when the wind field is zero.  The latter is not part of the physics of what is happening in the atmosphere but a projection.  That projection could cause problems.

By function "get_stepwise_f_partials" in rte.cc, which does the adoption, these values are zero when the function "dotprod_with_los" is zero.  I actually don't see why it is zero for all los.  There is something in "mirror_los" causing this?

erikssonpatrick:
@riclarsson I had a look at the Qpack demos involving wind retrievals and they work fine even with a zero wind as a priori. After some testing, I figured out that the problem in TestWfuns was that abs_interp_f was zero. When setting abs_interp_f=3 I got a good agreement with the perturbation calculations. That is, all seems OK. I added a check to jacobianAddWind to avoid this problem in the future. So you had fixed this some time anyhow!
Maybe you should have a look at the internal Jacobian code to remind you about what is going on. Maybe most critical is if abs_interp_f must > 0 for some other Jacobian calculations. If yes, the same test should be added to the associated Add method. I leave it to you to close this issue when you have considered this.

@stefanbuehler As you were active in the discussion.

olemke:
Needs further discussion based on PR #64 whether the constraint of abs_interp_f>0 can be removed.

riclarsson:
@erikssonpatrick Good.  It seems I was not remembering solving that issue then.  Or perhaps I messed it up with VMR behaving that way.  Anyways...

I had a quick look at your abs_f_interp_order variable.  The previous code was correct and you should revert your change.  The Jacobian should be 0.  If you use nearest-value interpolation, then the wind changes the frequency in step functions.

No sane user will ever run lookup table calculations with abs_f_interp_order=0 and expect a good wind Jacobian anyways.  At the same time, ARTS never promises to check user sanity.  In fact, returning zeroes here means that the user can discover their error.

I would mark this issue as wontfix, revert #64, and remove the addition of abs_f_interp_order before merging that pull back into ARTS.

stefanbuehler:
Hi @riclarsson,

this is not the intended behaviour of the lookup table. The relevant lines in abs_lookup.cc (685) are:

[...]

PR #64, opened 2019-09-08
Revision of Jacobian testing

erikssonpatrick:
This should conclude my present work on tests of the Jacobian calculations. 
Closed the previous pull request when I found another issue. Did not manage to reopen that one and now made a new pull request.
Details in the commit comments.


riclarsson:
Please explain why this requires an extra workspace variable and not just a GIN for the relevant function.  Which is the relevant function?

erikssonpatrick:
@riclarsson See my comment regarding the issue on Wind Jacobian

riclarsson:
Right, I think this pull should remove the addition of abs_f_interp, and that we should instead look into the cause of the problem.  This plaster seems excessive since it changes functions that are not relevant for the error itself.  (Also, at least sometimes, if you use the hashtag to write the number as #55 , github seems to create an automatic link.  Just an FYI for the future.)

erikssonpatrick:
@riclarsson If you can find a solution to this that removes the constraint of abs_interp_f>0, that is of course excellent. I can then rewind that part of this pull request. But until we know that is possible, I prefer that this pull request is not delayed as it involves a number of commits and would be nice to get them into the master.

olemke:
Merge this for now. Please continue discussion in #55

PR #69, opened 2019-09-11
Declare Matrix conversion operators as explicit.

olemke:
Affects: Sparse and CovarianceMatrix classes.

Conversion operators of intricate types are better defined 'explicit'
to avoid unnoticed, unwanted, potentially expensive implicit
conversions.

This also happens to fix the internal compiler error with GCC 9 in
C++17 mode.

Additionally, ensure const-correctness in static_casts.

### DIFF
--- a/3rdparty/invlib/src/invlib/interfaces/arts_wrapper.h
+++ b/3rdparty/invlib/src/invlib/interfaces/arts_wrapper.h
@@ -249,7 +249,7 @@ public:
         if (is_inverse_) {
             return covmat_.get_inverse();
         } else {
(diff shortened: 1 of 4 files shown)


riclarsson:
Note that this commit made several of the tests fail.  Errors include:

```
test_sparse.cc:429:47: error: no matching function for call to ‘get_maximum_error(Sparse&, Sparse&, bool)’

src/test_sparse.cc:487:9: error: no match for ‘operator=’ (operand types are ‘Matrix’ and ‘Sparse’)

src/test_sparse.cc:584:54: error: no matching function for call to ‘get_maximum_error(Sparse&, Matrix&, bool)’
```

PR #80, opened 2019-09-23
Replace Fortran Disort with C Disort

olemke:
This PR adds cdisort version 2.1.3 to the 3rdparty directory in ARTS.

The WSM DisortCalc now uses cdisort by default. For testing purposes, reverting to Fortran Disort is supported via a generic argument to DisortCalc.

The disort ARTS code has been cleaned up and all unused functions, as well as the experimental DisortCalcWithARTSSurface, have been removed.


olemke:
@erikssonpatrick Is it okay with you to remove DisortCalcWithARTSSurface? Are you aware of anyone using it currently? @m-brath proposed do a new implementation at some point using the BDRF functionality available in cdisort.

erikssonpatrick:
@olemke @m-brath A good question. The surface is an issue for most of our scattering solvers! My opinion is that the scattering solvers should consider surface_rtprop_agenda and that the surface can be placed at any altitude. I assume DisortCalcWithARTSSurface was an attempt to move in this direction. Despite this, I say let's remove DisortCalcWithARTSSurface. I think it is best to start from scratch.

Can you inform me about how the surface is treated right now in the cdisort version? Whatever you have done, it is OK for now. But feel free to make an issue on improving the integration of Disort and RT4 with respect to how ARTS defines the surface properties. That would be something for me if I just find the time ...

olemke:
Currently, the surface is treated identical in C and Fortran Disort. It uses a lambertian bottom boundary with surface_scalar_reflectivity as the albedo.

erikssonpatrick:
@olemke As I wrote, further work on the surface inside cdisort should be a new issue. But maybe this could be quickly done: If my memory is correct, cdisort can do both lambertian and specular surface. Correct? If yes, could you just add a gin to select between the two options? Then using surface_scalar_reflectivity for both options.
Specular is what is assumed normally for microwave simulations, so would be good to have that option at hand.

olemke:
CDisort supports lambertian and different types of BRDF. I think if I turn off lambertian and set BRDF_TYPE to None, it should result in a specular surface.

olemke:
Thanks @erikssonpatrick. I'm indeed not certain, mostly due to the lack of documentation. :-( Except for the paper, everything I could find documentation-wise, is included in the [source directory](https://github.com/olemke/arts/tree/cdisort/3rdparty/cdisort). [This example in disotest.c](https://github.com/olemke/arts/blob/07b54f4747aa9ff853bf73a0a9276bf220c22929/3rdparty/cdisort/disotest.c#L1147) seems to do non-lambertian without special BRDF, but is it really specular? Maybe you can extract some helpful information from [DISORT2.doc](https://github.com/olemke/arts/blob/cdisort/3rdparty/cdisort/DISORT2.doc).

erikssonpatrick:
@olemke My conclusion is that there are two options for the surface. The simplest option is ds.flag.lamber = TRUE; This option should be identical to what we have in the fortran version. So this is OK for the moment. 
Second option: Other surface models are found in c_bidir_reflectivity. I assume setting ds.flag.lamber = FALSE; activates this function, but I have not checked the details. That is, it should hopefully be possible to link cdisort to surface_rtprop_agenda by c_bidir_reflectivity. 
In summary, no simple fix right now. You have done your part. OK for me to accept the pull request.

olemke:
@erikssonpatrick, that fits with @m-brath and my conclusion.

@m-brath and I want to do one more test of cdisort tomorrow with a realistic scenario. If that is successful, I'll remove the Fortran version and merge the PR.

olemke:
The tests have shown that cdisort is in good agreement with the Fortran version. Thus, the Fortran code has been removed.

PR #86, opened 2019-09-27
Fix most of the iyActiveSingleScat2 problems

riclarsson:
The bit that is changed is only that the commutative approach to single back-scattering now also takes derivative of transmission to and from the scattering events into account.  The results did agree mostly with the old tests.  Still some odd issues such as my values often being much larger for the first entry to the list.  This should be a good enough improvement to warrant a look.

I could not figure out how to solve the full transmission derivatives.  There is something strange with the ground-to-scattering part of the derivative.  It is not acting as I thought it would.  The 2D case of the code is implemented but commented out and throws a runtime if activated by mistake.

Github is acting strange on this, pushing a lot of old code...

Edit:  I have no idea what is going with github but it says I cannot merge this but still push it, so I assume there is some config error on my side.  I cannot solve this so I will leave this here anyways.

### DIFF
--- a/src/m_cloudradar.cc
+++ b/src/m_cloudradar.cc
@@ -865,8 +865,10 @@ void iyActiveSingleScat2(Workspace& ws,
   ppvar_trans_cumulat.resize(np, nf, ns, ns);
 
   ArrayOfRadiationVector lvl_rad(np, RadiationVector(nf, ns));
-  ArrayOfArrayOfRadiationVector dlvl_rad(
-      np, ArrayOfRadiationVector(nq, RadiationVector(nf, ns)));
+  ArrayOfArrayOfArrayOfRadiationVector dlvl_rad(np,
+    ArrayOfArrayOfRadiationVector(np,
+    ArrayOfRadiationVector(nq,
+    RadiationVector(nf, ns))));
 
   ArrayOfTransmissionMatrix lyr_tra(np, TransmissionMatrix(nf, ns));
   ArrayOfArrayOfTransmissionMatrix dlyr_tra_above(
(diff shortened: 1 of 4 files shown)


olemke:
Run `git pull --rebase upstream master`. I don't know how you update from upstream, but your master branch is missing all merge commits since Sep 6 from upstream. I recommend to set rebase by default with `git config --global pull.rebase true`, then you don't have to specify the option everytime.

riclarsson:
I followed your commands.  git complains about something and adds a few lines of unparsable code to my files.  When I manually remove this the program complains about me being in the middle of a rebase but since I fixed the issue, I tell it to ignore the problems.  It does not allow me to continue the rebase so I instead tell it to skip the fantasy problem it perceives.  I already solved it anyways.  Then the above happens.

I simply think I will have to stop using the branch system.  It seems poorly implemented and it keeps generating random these random noise even though the code compiles.

erikssonpatrick:
Seems easiest to sort out the commit issues and get this into master. I can do some checks when the pull request is finished.
@riclarsson So more later. But ask already now. What do you mean with "something strange with the ground-to-scattering part of the derivative"? We don't consider the ground in iyActiveSingleScat.

olemke:
@riclarsson I would really like to get to the bottom of your merge issues. Could you please describe what your exact workflow is to merge new commits from upstream into your branches (the actual commands)? I noticed in your newlinerecord branch as well that there are no merge commits anywhere. If you were using `git pull --rebase upstream master`, the order of commits would be different (your own would be on top, commits from master below in their original order, including merge commits for PRs). That suggests that you were using 'git pull upstream master' without rebase. But in that case, there should be an additional merge commit added by git every time you pull in new commits from upstream. These merge commits are also not present in your branch. That suggests that you're doing something else after pulling from upstream.

I tried to reproduce a branch that exhibits the same features as yours (containing all commits from upstream, but no merge commits at all). The only way I could do that was by merging from upstream (git pull upstream master) and then additionally running a `git rebase` without specifying upstream/master. Running `git rebase` without any explicit branch will, by default, rebase against the current tracking branch (origin/master or origin/newlinerecord) which at this point is outdated. This will remove the merge commits and treat all new upstream commits as if they only exist in your local branch. This could create the strange conflicts you're seeing later on when rebasing with upstream master again. git is now lacking the additional information from the merge commit that would otherwise tell it that these commits already exist in upstream/master and the proper order in which they should be applied.

My advice would be to only use `git pull --rebase upstream master` to bring your branches up to date with upstream master.

riclarsson:
@olemke I will try to follow that.  It seems to work from time to time but also to generate strange errors.  I only update the branch I am in since I need to be able to run other code regardless of if it is the latest or not.  Lately and for the foreseeable future before Stefan approves of the line record format, that will be the branch I follow.  It touches quite a lot of code so I suspect it will generate a lot of bugs once merged...

@erikssonpatrick I imagine the system is a ground-transmitter when I chooses my words.  Anyways, the bits not working are the bits filling the Jacobian before the back-scattering.  I believe there is a severe indexing error.  Because the math is somewhat easy but your way of using iy as a Tensor3 and diy as a Tensor4 by playing around with the indices is quite frankly not very intuitive to me.

PR #91, opened 2019-10-03
Single moment mass gamma psds

stuartfox:
I want to use some additional PSDs that are not currently in ARTS (e.g. the Field 2019 hail PSD). The PSDs are gamma distributions, but the intercept parameter is a function of the slope parameter, i.e. n(D) = n_alpha lambda^n_b D^mu exp(-lambda D). This form is commonly used for single-moment schemes where only the hydrometeor mass is defined. Note that the existing Abel & Boutle and Wang rain PSDs are also of this form.

I have implemented the psd (in psd.cc) as psd_gd_smm_common, which can either take generic arguments for n_alpha, n_b and mu, or specific named psds such as A12, W16, F19. This is then called by the workspace methods for specific PSDs (psdA12, psdW16 etc.) as well as the generic psdGdMassSM (SM standing for single moment).

However, I'm not sure if this is the best way to do things. 
  *  Would it be better to follow the approach of psdMgdMass and allow the parameters to be defined either as generic inputs or through the pnd_agenda_input? 
   *  Alternatively, psdMgdMass could probably be modified with an additional input n_b to permit a lambda-dependence on N0 within the existing function?
   *  Should the PSD be generalised to allow modified gamma distributions (as is already permitted for the psdMgd functions)?

Note that the current implementation replaces the existing A12 and W16 calculations and leads to some changes in the TestWfunHybCloudy output causing this test to fail. I think this is probably just because I've converted the W16 parameters to SI units and have introduced some rounding differences. I've also left psd_rain_W16 in psd.cc as it is still called directly from the old pnd_fieldCalcFromscat_speciesFields

Thoughts/comments welcome...


erikssonpatrick:
First of all, great that you are doing this. And that you have managed to build upon the existing system for MGD PSDs. I was afraid that it was too complex. 
Also great that you have replaced A12 and W16. Should have been done before ...

"Would it be better to follow the approach of psdMgdMass and allow the parameters to be defined either as generic inputs or through the pnd_agenda_input?"
If not clear, if a parameter is set as a GIN, it will be constant throughout the atmosphere. If it is set by pnd_agenda_input, it can vary between points. I don't think the latter option is of concern here. So the answer seems to be no.

"Alternatively, psdMgdMass could probably be modified with an additional input n_b to permit a lambda-dependence on N0 within the existing function?" 
It is probably already quite hard to understand the options provided, and this extension should make it worse. So no again. I am happy with your solution.

"Should the PSD be generalised to allow modified gamma distributions (as is already permitted for the psdMgd functions)?"
Do you mean psdGdMassSM here? Not critical from my side, but all that takes is another GIN, or? So why not. I leave this to you ...

The PSD names are a bit cryptic and I will switch to somewhat longer names, such as psdRainAbel12. Replacing Mgd with ModifiedGamma? To follow a general trend in ARTS, moving towards longer method names. Any opinion or suggestion?
This brings me to psdGdMassSM. Should we go for psdGammaMassSingleMoment?

(There was at least one description line in methods.cc that missed \n) 

So just some details ...

stuartfox:
Thanks for the comments Patrick. I've extended psdGdMassSM to also allow modified gamma distributions and renamed it as psdModifiedGammaMassSingleMoment.

I've also taken the opportunity to rename A12 and W16 as you suggested, although perhaps you'd prefer to do this in a separate wholesale change?

What's the process for fixing the failing TestWfunHybCloudy? I'm guessing the expected output either needs to be updated or I need to try and better reproduce the exact values from the old Wang16 calculation?

erikssonpatrick:
Keep the renaming of methods that you have done. I can fix the rest in the branch I have started to make my part of removing the old PSDs.
This work includes replacing the tests making use of the old PSD. As TestWfunHybCloudy is one of those, I suggest simply to remove this test from controlfiles/CMakeLists.txt. (I would not worry about the deviation from existing references values. I am sure your results are more correct).

stuartfox:
OK. I wouldn't say my results for TestWfunHybCloudy are "more correct" though:

The parameter values in the Wang paper use cgs units (e.g. for n_alpha and nb). The old implementation calculated N0 and lambda in cgs, with a conversion to SI just prior to calculating the exponential distribution (and a slightly confusing definition of c1 in SI units that then cancelled with the SI units of rwc to give a dimensionless quantity used in the calculation of lambda). For the new implementation I used SI units throughout, but probably introduced some rounding differences in the value of n_alpha.

erikssonpatrick:
@olemke From my side, you can approve the pull request. Still maybe some unclarity for TestWfunHybCloudy inside this branch, but I have made a commit in another branch where the test is removed.

PR #101, opened 2019-10-22
DISORT, RT4 and iyInterpCloudboxField

erikssonpatrick:
Both DISORT and RT4 can now handle cases with the surface above p_grid[0]. Some code cleaning in these parts. 
Also cleaned up around iyInterpCloudboxField. Now only one method for this and fixed a bug.
Some improvements of TestScatSolvers.arts.


olemke:
Please try to have a [short first (subject) line (<=50 chars) in your commit messages](https://chris.beams.io/posts/git-commit/#seven-rules) in the future to keep the log tidy.

src/m_cloudbox.cc
```diff
@@ -578,6 +579,452 @@ void cloudboxSetManuallyAltitude(  // WS Output:
...
+        "(This method does not yet handle dispersion type calculations.)");
+  //---------------------------------------------------------------------------
+
+  // At each lat/lon, one value of the intensity field is defined at the
+  // surface. Simplest way to handle this is to insert z_surface into z_field
+  Tensor3 z_with_surface = z_field;
+  for (Index ilat=0; ilat < z_surface.nrows(); ilat++) {
+    for (Index ilon=0; ilon < z_surface.ncols(); ilon++) {
+      Index ip = 0;
+      for (; z_surface(ilat,ilon) >= z_field(ip+1,ilat,ilon); ip++) {}
```
olemke: This for loop with an empty body gives the impression that it was accidentally left empty.
A while loop would make your intention clearer:

```cpp
while (z_surface(ilat,ilon) >= z_field(ip+1,ilat,ilon)) {
  ip++;
}

PR #107, opened 2019-11-07
Access to empty array if jacobian_do = 0

simonpf:
This was reported by Rita Edit  Kajtar on the ARTS mailing list and I tracked it down to a missing check for jacobian_do in stepwise_source causing access to an array of size 0.

What I don't know, however, is why this didn't show up before.

Maybe one of those who are more familiar with this part of the code can check if this is a valid fix.


erikssonpatrick:
This looks correct to me.

PR #111, opened 2019-11-21
errors in the absorption_theory.tex

gongxunhlj:
correct some errors in the document absorption_theory.tex


olemke:
Thanks for your contribution!

PR #117, opened 2020-01-09
Add some Zeeman species to test and to find good format

riclarsson:
The main point with this commit is to see if we can find a good way to use the formats for these extra Zeeman species.  I add CO2 and CO and OCS here, all of which have an extremely weak Zeeman effect.

### DIFF
--- a/src/zeemandata.cc
+++ b/src/zeemandata.cc
@@ -91,10 +91,10 @@ Numeric case_b_g_coefficient_o2(Rational j,
            2 * GLE * cos(2 * phi) / (2 * J + 1) - GR;
 }
 
-Numeric closed_shell_trilinear(Rational k,
-                               Rational j,
-                               Numeric gperp,
-                               Numeric gpara)
+constexpr Numeric closed_shell_trilinear(Rational k,
+                                         Rational j,
+                                         Numeric gperp,
+                                         Numeric gpara)
 {
   if (k.isUndefined() or j.isUndefined() or j == 0)
     return 0;
--- a/src/zeemandata.cc
+++ b/src/zeemandata.cc
@@ -64,12 +64,11 @@ Numeric case_b_g_coefficient_o2(Rational j,
   else if (j == 0)
     return 0;
 
-  Numeric J = j.toNumeric();
+  auto J = j.toNumeric();
 
-  Numeric nom = (lB + lD * (J * J + J + 1) + lH * pow2(J * J + J + 1)) *
-                (2 * sqrt(J * J + J) / (2 * J + 1));
+  auto nom = (lB + lD * (J * J + J + 1) + lH * pow2(J * J + J + 1)) * (2 * sqrt(J * J + J) / (2 * J + 1));
 
-  Numeric denom =
+  auto denom =
       B * J * (J - 1) - D * pow2(J * (J - 1)) + H * pow3(J * (J - 1)) +
       (gB + gD * J * (J - 1) + gH * pow2(J * (J - 1))) * (J - 1) +
       (lB + lD * J * (J - 1) + lH * pow2(J * (J - 1))) *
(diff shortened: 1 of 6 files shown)


src/zeemandata.cc
```diff
@@ -147,6 +157,49 @@ Zeeman::Model Zeeman::GetAdvancedModel(const QuantumIdentifier& qid) noexcept {
...
+      Numeric gperp = -.0285 / Constant::mass_ratio_electrons_per_proton;  // Flygare and Benson 1971
+      Numeric gpara = -.061 / Constant::mass_ratio_electrons_per_proton;  // Flygare and Benson 1971
+      
+      auto JU = qid.UpperQuantumNumber(QuantumNumberType::J);
+      auto KU = qid.UpperQuantumNumber(QuantumNumberType::Ka);
+      auto JL = qid.LowerQuantumNumber(QuantumNumberType::J);
+      auto KL = qid.LowerQuantumNumber(QuantumNumberType::Ka);
+      
+      return Model({closed_shell_trilinear(KU, JU, gperp, gpara),
+                    closed_shell_trilinear(KL, JL, gperp, gpara)});
```
olemke: Don't see any way to handle that more cleanly for all species in general due to the very different parameters needed for each species. But we can try to reduce the redundant code per isotopologue. Something like this maybe:

```cpp
 } else if (qid.SpeciesName() == "OCS") {
    Numeric gperp, gpara;

    if (qid.Isotopologue() == SpeciesTag("OCS-622").Isotopologue()) {
      gperp = -.02889 / Constant::mass_ratio_electrons_per_proton;  // Flygare and Benson 1971
      gpara = 0 / Constant::mass_ratio_electrons_per_proton;  // Flygare and Benson 1971
    } else if (qid.Isotopologue() == SpeciesTag("OCS-624").Isotopologue()) {
      gperp = -.0285 / Constant::mass_ratio_electrons_per_proton;  // Flygare and Benson 1971
      gpara = -.061 / Constant::mass_ratio_electrons_per_proton;  // Flygare and Benson 1971
    } else {
      return Model();
    }
    const auto JU = qid.UpperQuantumNumber(QuantumNumberType::J);
    const auto KU = qid.UpperQuantumNumber(QuantumNumberType::Ka);
    const auto JL = qid.LowerQuantumNumber(QuantumNumberType::J);
    const auto KL = qid.LowerQuantumNumber(QuantumNumberType::Ka);

    return Model({closed_shell_trilinear(KU, JU, gperp, gpara),
                  closed_shell_trilinear(KL, JL, gperp, gpara)});
```

riclarsson: I like the clearer split; that there is either a zero-return or a good return in the end.  Your example saves some lines but I find it more cumbersome to deal with since it would have to happen anyways per species.

That said, I don't like that the only way I can figure out how to get to the 12C-16O isotopologue, for example, is by the use of SpeciesTag("CO-26").  There must be some constexpr-able way of having the Index of the isotopologue automatically found out so that an overarching switch-statement for all isotopologue can be used.  Especially as I am almost certain that "CO-26" is the same isotopologue in ARTS as "OCS-622".  We would be better off separating the real isotopologue and the full models with a redesign of species data...

olemke: I'm fine with leaving it as it is. With a redesign of species_data we should hold off until more of the data types actually support constexpr. Since there is no constexpr support for enums, strings or vectors yet, it feels like quite some compromises in the implementation would be needed to get it to work right now. And while it would be convenient to be able to use a switch statement instead of if-elseif..., the gain would be rather marginally to justify a complete redesign.

Should I go ahead with merging this PR?

riclarsson: Right, I agree we can wait until there are more constexpr things.  It's just that I am working with quite a few things that require species-specific data that is not in species_data.  It would be nice to have a cleaner way to define species_data so that these species data can be modified and added.  For instance, the numbers in for O2 Zeeman effect are what is needed to compute the energy levels for line mixing.  These are therefore defined twice, once per functionality.  Since I know the numbers are good for the Zeeman stuff, they should be the same for line mixing stuff.  Yet because the line mixing stuff is already so complicated and there is still a bit to do to make it work (and I have little time for it), I have had to redefine things to make the functions readable.

Anyways, it is OK to go ahead and merge this code.  I will see if I can update other things later.

PR #120, opened 2020-01-17
Update absorption_theory.tex

gongxunhlj:
Update absorption_theory.tex


olemke:
@gongxunhlj Thanks for the corrections.

PR #121, opened 2020-01-20
Fix reference comparisons in TestNLTE

olemke:
The test was storing the current result with the same filenames as the
reference files, reading them for comparison instead of the real reference
files. Thus, the test never failed.

Added 'REFERENCE' to the filenames and updated them with current values.
The test was failing with the old reference files, I assume they were
outdated.

@riclarsson: The sign has changed for most values in the new reference files compared to the old ones. Do you think this is okay?


riclarsson:
Hi Oliver,  I don't see the sign change?  What are you talking about exactly?

olemke:
Oops, sorry, I fooled myself by the diff output. So the values only changed slightly. I guess it's okay to merge this then.

PR #122, opened 2020-01-23
Add dark theme for docserver

lkluft:
Add a dark theme for the built-in documentation server. The dark theme
is used depending on the user's system settings.


lkluft:
In commit 91200c039836ee, I adjusted the link colors to make them distinguishable from normal text (find links in text blocks). The dark mode uses the same color coding as the light mode, only the saturation is adapted.

PR #127, opened 2020-02-03
Mpm in absorptionlines

riclarsson:


riclarsson:
This was meant as an addition after the other branch had been accepted.  It should be pushed ontop and not parallel to the other commit.  Somehow git is confused about it and includes the other commits...

Anyways, @erikssonpatrick , if you want MPM-style ARTS input via catalog, you need to use the @6822893 branch style of code.  If I have not made too many errors, which I may have (I will check and update), it should be enough.  Note that this code is not tested but describes the process.

riclarsson:
Here is how a catalog would look that can do MPM-like calculations inside the ARTSCAT-6 line catalog format for reference.

[O2-66.zip](https://github.com/atmtools/arts/files/4149025/O2-66.zip)

riclarsson:
This now works with VVW renormalization.  Note that this code should not really be merged before we consider #128

PR #133, opened 2020-02-18
Update the arts_thoery document

gongxunhlj:
fix some errors.

### DIFF
--- a/doc/uguide/absorption_theory.tex
+++ b/doc/uguide/absorption_theory.tex
@@ -2527,7 +2527,7 @@ four for the line width. The total line absorption coefficient
 coefficients:
 \begin{equation}
   \label{eq:abs_theory:mpm93:absline}
-  \alphampmnl = 0.1820 \cdot \nu \cdot \phzo \cdot 
+  \alphampmnl = 0.1820 \cdot \nu \cdot
   \sum_{k}{\inten \cdot \shape}\hspace*{10mm}\mbox{dB/km}
 \end{equation}
 where $\inten$ is the line intensity described by the parameterization
--- a/doc/uguide/absorption_theory.tex
+++ b/doc/uguide/absorption_theory.tex
@@ -2058,7 +2058,6 @@ following expression
     \alphac &=& \widehat{C} \cdot \frac{\nu^2}{(1.0+a\cdot \nu^{n_{\nu}})} 
                 \cdot \Theta^{3.5} \cdot \pda^2
                 \hspace*{16mm}:~\mbox{MPM93}\\
-          a &=&  
 \nonumber
 \end{eqnarray}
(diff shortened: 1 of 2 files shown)


gongxunhlj:
Update the arts_thoery document

doc/uguide/absorption_theory.tex
```diff
@@ -2527,7 +2527,7 @@ \subsubsection{MPM93 water vapor absorption model}
 coefficients:
 \begin{equation}
   \label{eq:abs_theory:mpm93:absline}
-  \alphampmnl = 0.1820 \cdot \nu \cdot \phzo \cdot 
+  \alphampmnl = 0.1820 \cdot \nu \cdot
```
olemke: First of all thanks a lot for your continued work on corrections for the theory guide. We really appreciate that (and are surprised how many typos there actually are ;-) ).
I have one question about the change in this line: Why did you remove the PH2O in this place? It seems to me it should be present here just like in all other MPM models?

gongxunhlj:
This software is excellent, and I try to learn it.
When I read the arts theory, I code a python program. But the result was different from that in ref. [1]. So I compare MPM93 in arts theory with ref. [1]:
in ref [1], the pressures are only present in line intensity equations, not in alpha, so is MPM93 oxygen model in arts theory. But in MPM93 water vapor model of arts theory, the PH2O is present in both alpha and line intensity. I think one of them should be removed.
I didn't check the MPM89 and MPM87 water vapor models. I think maybe there are the same problems in them.
Please check them (include MPM93). Maybe I'm wrong.


Reference:
[1] Liebe, H. J., G. A. Hufford, and M. G. Cotton, Propagation modeling of moist air and suspended water/ice particles at frequencies below 1000 GHz., in AGARD 52nd Specialists Meeting of the Electromagnetic Wave Propagation Panel, Palma de Mallorca, Spain, 1993


 
From: Oliver Lemke
Date: 2020-02-25 14:45
To: atmtools/arts
CC: gongxunhlj; Author
Subject: Re: [atmtools/arts] Update the arts_thoery document (#133)
@olemke commented on this pull request.


In doc/uguide/absorption_theory.tex:
> @@ -2527,7 +2527,7 @@ \subsubsection{MPM93 water vapor absorption model}
 coefficients:
 \begin{equation}
   \label{eq:abs_theory:mpm93:absline}
-  \alphampmnl = 0.1820 \cdot \nu \cdot \phzo \cdot 
+  \alphampmnl = 0.1820 \cdot \nu \cdot

First of all thanks a lot for your continued work on corrections for the theory guide. We really appreciate that (and are surprised how many typos there actually are ;-) ).
I have one question about the change in this line: Why did you remove the PH2O in this place? It seems to me it should be present here just like in all other MPM models?

PR #138, opened 2020-03-11
Python integration

simonpf:
- Add `arts` Python pacakge to ARTS
- Adds automatically generated conversion of `*.arts` files in the controlfiles directory
- Test now automatically also run Python version
- `ctest -R fast` passes ARTS and Python versions of tests.

PR #140, opened 2020-03-16
ARTS classes by C-pointer in python

riclarsson:
Since we are soon able to rely on python being there and working, I took some time to put together a class representation of the ARTS classes to be accessed via python.  A lot of static_cast(s) later and it seems to work.  I am not entirerly sure how to get this to run with the new python-interface (it does not build for me when I download the file) but it works with the equivalent typhon interface.  I attach a file for testing.  Ideally, there's a way to send the pointers to ARTS variables somehow but I have yet to figure that out how...  

The attached file shows the python interface that can access and manipulate all of these classes.  I for now just implemented everything required for AbsorptionLines, since it had no python interface anyways...

[testing.zip](https://github.com/atmtools/arts/files/4339040/testing.zip)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -90,7 +90,7 @@ set_source_files_properties (m_oem.cc PROPERTIES
 ########### next target ###############
 
 if (C_API)
-add_library (arts_api SHARED arts_api.cc interactive_workspace.cc)
+add_library (arts_api SHARED arts_api.cc interactive_workspace.cc arts_api_classes.cc)
 add_dependencies (arts arts_api)
 set_target_properties(arts_api PROPERTIES SUFFIX .so)
 target_link_libraries (arts_api ${ALL_ARTS_LIBRARIES})
--- a/src/abs_species_tags.h
+++ b/src/abs_species_tags.h
@@ -62,6 +62,9 @@ class SpeciesTag {
 
   /** Molecular species index. */
   Index Species() const { return mspecies; }
+  
+  /** Set molecular species index. */
+  void Species(Index x) { mspecies = x; }
 
   /** Molecular species index. */
   Index BathSpecies() const { return mcia_second; }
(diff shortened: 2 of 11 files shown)


olemke:
I'll merge this for now since it is well contained in separate files and will be good for testing. The amount of code that has to be written manually for each class to define the interface (and adapted in case of changes) is a bit concerning though.
@riclarsson If the branch from PR #138 doesn't build for you, it would be good to post the error you get. Otherwise we can't fix it and I'm not entirely sure what you mean by "(it does not build for me when I download the file)".

riclarsson:
@olemke Thanks!  I I am sure plenty of this can be automated into the same routines that ensures we have xml-IO for all workspace variables.  The ones that can clearly go there are the codes for create*, print*, and delete*.  I am sure there's some smart way to also encode xmlload* and xmlsave* routines there.

The other interfaces I am not so sure about.  For instance, how do you actually ensure that a species is valid?  Or that a key for a line shape type is correct?  This requires special functions because it is hidden in the common initialization, and the variables there are not always one-to-one with what we actually store inside the code (which is why I added several ": Index" to our enums in this commit).

Besides, explicit code means that there will be explicit errors.  Which will be helpful for debugging..

simonpf:
My design idea for the Python interface was to have explicit interfaces only for ARTS groups for which in-place modification would be of interest. Is that the case for the absorption coefficient data? 

Otherwise I don't see why we shouldn't just use the XML serialization for the ARTS - Python bridge. In that case you only need to add a suitable class on the Python side. If that turns out to be a bottleneck, this can still be modified to avoid storing to disk (It's easy on linux, I think the only reason we don't have it is mac support.)

riclarsson:
@simonpf 

Absorption::Lines  class does not have a python equivalent yet.  It would be significantly more work to try and create a python version of that class than it was to just use the existing data structure.  I will make the IO work as for the other classes so that passing by XML is still possible.

It seems a bit wasteful since the class can work with just the pointer, but if I don't understand the other parts of the C-API so I cannot judge how much work it would be to be able pass the pointers.

I will make sure that the other classes that I had to implement to get to Absorption::Lines replaces their old implementations and that the ones that are Workspace variables also have XML-IO.

@olemke It builds.  I had to install some packages but I do not remember which.  The problem was I forgot to run "conda-develop $ARTS/build/python/"

PR #141, opened 2020-03-25
Pr138 python classes

riclarsson:
I am sure there is something weird here because this pull contains all of Simon's and Oliver's stuff as well.  This can probably be fixed after the #138 is merged.

Here is anyways anyways full IO support for all Workspace variables via mapping to objects generated by ARTS itself.  This means you can use ARTS xml-IO directly, and pass just a pointer to the variable in the interface to use a variable in python.  I have not touched the other python code.  There's plenty of things missing, like determining a Workspace type from just the file, but I am sure that can be fixed.

Some Workspace variables have limited interface at this stage and can only do IO.  These are:  Agenda, ArrayOfAgenda, CovarianceMatrix, GasAbsLookup, MCAntenna, ArrayOfRetrievalQuantity, Sparse, ArrayOfSparse, TelsemAtlas, ArrayOfTelsemAtlas, Timer, and ArrayOfXsecRecord. It is not too difficult to have them expose some interface but I am unsure how to make it without a lot of boilerplate.  Also, Agenda, CovarianceMatrix, and Sparse are implemented in current pointer transfer, so that code might be better to use.

There were some but not too many changes required in ARTS internal code to make this interface.  Some probably bad designs are in the enum-handling, where perhaps a throw would be better and to only allow the text-setting in the public interface.  I am sure there are other interface details that could be better as well.  A big thing that I have changed when I did not think it would change the binary output was to make all enums Index.  Notably, SingleScatteringData has an enum that that is not Index. Is it possible to set this top Index?  It anyways aligns the data to 8 byte because Numeric is 8 bytes, so nothing is saved internally.

As an extra, there seems to be a few debug unfriendly classes in ARTS that lack even basic stream output.  These are:  Ppath, TessemNN, Timer, and Any.  The last one is not so important (providing a "Don't do this" string as output would at most be nice), but the other three should probably have basic screen output.


olemke:
> I am sure there is something weird here because this pull contains all of Simon's and Oliver's stuff as well. This can probably be fixed after the #138 is merged.

Nothing weird here (related to git). Since the PR applies to master, it shows all changes that are not in master yet.

riclarsson:
@olemke OK, so this can be fixed when the other branch is merged.

I might commit some more code to try and interface to some of the remaining classes better in the meanwhile.

I am sure there are also some design choices that are better to establish before merging for how to interact with the classes.  Before the other merge, this commit log is a mess so it is better to wait with reading it before then.

riclarsson:
Thanks for the merge+push Oliver.  I musty have gotten the python version from Simon too early but after installing lark with pip it works.

Question: is there no way to have a "python-fix" command execute to "pip install" the required python packages?  If pip is in the system, which it often is when you have python, it should be fairly trivial to execute that command (even if we give the warning that it requires 3rdparty-downloads).

olemke:
No, you cannot make assumptions about the user's preferences how he/she wants to install additional packages in their system. Some might prefer to use conda, even though there is a pip command available. Some user might want to install directly with pip into their python path, other's might prefer a --user installation in their home directory.
The same way that users need to install build dependencies such as a compiler, latex, lapack or fftw, they should install the couple of Python dependencies. There are just too many variables.

riclarsson:
OK, I can agree with that.  However then the error messages must be better.  You say that you want lark but you want lark-parser.  These seems to be completely different packages that import to the same namespace.

The error message gave information about the wrong package.

Is there a way to check on start of compilation that lark-parser and not just lark is installed?  The latter does not provide the lark.Lark name.

olemke:
That's exactly why I insisted that we call our package pyarts on installation and not just arts. To match the name you use between the import and the package on PyPi. This is a complaint for the lark-parser developers who seems to have decided that they call there PyPi package lark-parser (probalby because lark was already taken), but they kept the name lark for their module instead of calling it lark-parser too. Really stupid. That's why I added the extra sentence about installing lark-parser, because the error lists the module names. I'll try to improve that message. It is not very nicely formatted anyway at the moment.

riclarsson:
This branch still requires a bit of work before merge.  The enum-handling is somewhat inconsistent.  There are no tests of the features anywhere.  Some classes have no interface to be manipulated internally.  Very few classes have an OK attribute for checking the sanity internally.  But most important, there is no way to easily translate from ARTS workspace to these python variables.  I will list these as more specific questions below but I don't expect this to be done too quickly.  Please also let me know if there is something else that should be fixed about this branch?

The enum-handling I have tried several options.  What is the best way to handle this?  A setter-getter of just index or should a "char *" setter be available for all variables with a a try-catch return of EXIT_FAILURE be the default?

Where and how do I populate the tests?  I would want some simple IO, IO+manipulation, and creation+testing of all properties.  So I need to store some data somewhere for each class.

The classes that lack interface are named in the classes __init__ file.  Some are available in the interface today (Sparse, CovarianceMatrix, Agenda) so I don't want to mess around there because I would just end up doing double work...

I am unsure how to add OK attributes to many classes because I have not been involved in their design.  I will try to add some to the ones I have touched on.  Is it even necessary to have one or should I just assume that ARTS methods will check that the classes are valid?

All these classes can get a simple void c-pointer to themselves and they will work.  I tried to use the pointer from the WorkspaceVariable itnerface but I think it is const?  Anyways, I get segfaults.  How do I get a modifiable pointer from an ARTS variable?

riclarsson:
So I can modify the pointer from the get_variable_value()-method it turns out.

Question:  Are the matpack types void pointing at their C-arrays rather than at the class in the workspace structure?  If so, is there any way to also get the class-pointer somehow from the Workspace?  Because I do not know how to modify the size of the underlying data in a workspace variable if I cannot access the pure pointer.

riclarsson:
The latest commit will/have/should fail

@simonpf  How do I access the cmake variable "ARTS_XML_DATA_PATH" in pyarts?

olemke:
In PR #151 I've added an environment variable to be exposed to the python scripts in controlfiles-python. Scripts can then access the path with `os.getenv("ARTS_XML_DATA_DIR")`. Note that this will only work when scripts are called via the testing framework by `make check-*` or `ctest -V -R TestExample.py`.

olemke:
Please define a function in each of the python scripts you import into TestClasses.py instead of writing all code on the toplevel. That would also remove the need of using `del` manually on all the variables you create for each test.

riclarsson:
I am not sure how to make the test succeed anymore.  Where and how have I declared that setuptools are not to be used?

olemke:
I don't think it's something you did. Maybe something changed in the linux distro no the build server. Give me a sec to look at it.

olemke:
Problem should go away as soon as I merge PR #152 . Not sure why but before setuptools got pulled in by dependencies automatically and now it needs to be installed explicitly. Upstream Python dependencies changed I guess.

riclarsson:
It works again!  Still my own copy fails because of "Release" not existing?!?!

Anyways, I will try to make some simple IO tests tomorrow for the non-xml files but after that I think this is ready for merger.  Or have I missed something?

The interface to workspace is not very clean yet but I don't know how to make it much cleaner without adding a property to the workspace-variables class.  But I do not know what to call this property...

olemke:
> It works again! Still my own copy fails because of "Release" not existing?!?!

Need more info. :-) You mean cmake's "Release" build type is broken?

> Anyways, I will try to make some simple IO tests tomorrow for the non-xml files but after that I think this is ready for merger. Or have I missed something?

I'll have a look through the code today and see if I find any show stoppers.
 
> The interface to workspace is not very clean yet but I don't know how to make it much cleaner without adding a property to the workspace-variables class. But I do not know what to call this property...

Not sure what to reply here, not much info to go on. :-)

BTW, as we discussed earlier, I had a look at the enum-string conversion. There seems to be no proper way to automate this conversion. Only thing I found was a janky hack that requires C++17, not really good either.

As using SWIG for the interface, it looks promising because it would only require one interface file to be written from which the C wrapper and Python classes are generated automatically. But one would need to dive deep into its inner workings to map some of our classes properly. It also seems to have some problems with protected default constructors (e.g. in our const views). So for now, I won't follow that road anymore.

riclarsson:
> > It works again! Still my own copy fails because of "Release" not existing?!?!
> 
> Need more info. :-) You mean cmake's "Release" build type is broken?

The warning is gone now.  On my github copy of ARTS, I used to get the warning on every commit.  the release thing was somehow related to Debian not having a release target.  It was very odd.  It worked on the checks here though.

> I'll have a look through the code today and see if I find any show stoppers.

Thanks!  I have added every class to have an interface now.  Sparse, CovarianceMatrix, and TelsemAtlas still lacks tests because I do not know how to easily generate them.

> Not sure what to reply here, not much info to go on. :-)

Presently, pyarts.classes.from_workspace(x) takes a workspace variable and turns it into a python class with full interface.  It would be nice if this was part of the normal WorkspaceVariable workflow, like how "ws.f_grid.value" works.  But I do not know what propery to occupy in WorkspaceVariable.

> BTW, as we discussed earlier, I had a look at the enum-string conversion. There seems to be no proper way to automate this conversion. Only thing I found was a janky hack that requires C++17, not really good either.

OK, I still think we need a method that we are OK with here.  Perhaps a simple but ugly `#define ENUMARTS(ENUMTYPE, ENUMLIST, ENUMSTR)`? Something like:

```
#define ENUMARTS(ENUMTYPE, ...) \
enum class ENUMTYPE : Index {__VA_ARGS__} ; \
constexpr bool good_ ## ENUMTYPE(Index x) noexcept {return x < Index(ENUMTYPE::FINAL) && x >= 0;} \
std::array<String, Index(ENUMTYPE::FINAL)> ENUMTYPE ## stringarray() noexcept {std::array<String, Index(ENUMTYPE::FINAL)> out; std::istringstream x(#__VA_ARGS__); for(Index i=0; i<Index(ENUMTYPE::FINAL); i++) std::getline(x, out[i], ','); return out;} \
String ENUMTYPE ## 2string(ENUMTYPE x) noexcept {static const auto strs=ENUMTYPE ## stringarray(); if (good_ ## ENUMTYPE(Index(x))) return strs[Index(x)]; else return "BAD " # ENUMTYPE; } \
ENUMTYPE  string2 ## ENUMTYPE(const String& x) noexcept {static const auto strs=ENUMTYPE ## stringarray(); for(Index i=0; i<Index(ENUMTYPE::FINAL); i++) if (strs[i] == x) return ENUMTYPE(i); return ENUMTYPE(-1); }
```

Shouldn't the `static const` make this a single execution?

> As using SWIG for the interface, it looks promising because it would only require one interface file to be written from which the C wrapper and Python classes are generated automatically. But one would need to dive deep into its inner workings to map some of our classes properly. It also seems to have some problems with protected default constructors (e.g. in our const views). So for now, I won't follow that road anymore.

Ok.  Anyways, the code should now have interface to all normal classes.  Though some of these classes have no data in arts-xml-data nor in controlfiles, so I do not know how to generate tests fro them.

olemke:
> OK, I still think we need a method that we are OK with here. Perhaps a simple but ugly `#define ENUMARTS(ENUMTYPE, ENUMLIST, ENUMSTR)`? Something like:
> 
> [snip]

While it is a viable solution, it won't work exactly like this. You cannot execute the macro as it is here in header files. While the enum usually needs to be defined in a header file to make it visible in other parts of the code, the macro also contains function implementations. These cannot be in header files because it will lead to linker errors because of duplicate definition of symbols.

Therefore it would need to be splitted into two macros, one for the definition of the enum and declaration of the mapping functions and the other one for the implementation of the mapping functions:

```cpp
#define ENUMARTSDECL(ENUMTYPE, ...)                                            \
  enum class ENUMTYPE : Index { __VA_ARGS__ };                                 \
  constexpr bool good_##ENUMTYPE(Index x) noexcept;                            \
  std::array<String, Index(ENUMTYPE::FINAL)> ENUMTYPE##stringarray() noexcept; \
  String ENUMTYPE##2string(ENUMTYPE x) noexcept;                               \
  ENUMTYPE string2##ENUMTYPE(const String& x) noexcept;

#define ENUMARTSIMPL(ENUMTYPE, ...)                                \
  constexpr bool good_##ENUMTYPE(Index x) noexcept {               \
    return x < Index(ENUMTYPE::FINAL) && x >= 0;                   \
  }                                                                \
  std::array<String, Index(ENUMTYPE::FINAL)>                       \
      ENUMTYPE##stringarray() noexcept {                           \
    std::array<String, Index(ENUMTYPE::FINAL)> out;                \
    std::istringstream x(#__VA_ARGS__);                            \
    for (Index i = 0; i < Index(ENUMTYPE::FINAL); i++) {           \
      std::getline(x, out[i], ',');                                \
      out[i].erase(std::remove(out[i].begin(), out[i].end(), ' '), \
                   out[i].end());                                  \
    }                                                              \
    return out;                                                    \
  }                                                                \
  String ENUMTYPE##2string(ENUMTYPE x) noexcept {                  \
    static const auto strs = ENUMTYPE##stringarray();              \
    if (good_##ENUMTYPE(Index(x)))                                 \
      return strs[Index(x)];                                       \
    else                                                           \
      return "BAD " #ENUMTYPE;                                     \
  }                                                                \
  ENUMTYPE string2##ENUMTYPE(const String& x) noexcept {           \
    static const auto strs = ENUMTYPE##stringarray();              \
    for (Index i = 0; i < Index(ENUMTYPE::FINAL); i++)             \
      if (strs[i] == x) return ENUMTYPE(i);                        \
    return ENUMTYPE(-1);                                           \
  }
```

The first one can be called in the header while the second one needs to be called in a .cc file.
To avoid duplicating the enum values in both macro calls, it could be done like this:

.h file:
```cpp
#define BACKSCATTERSOLVER_ENUM_VALUES \
  Commutative_PureReflectionJacobian, Full, FINAL
ENUMARTSDECL(BackscatterSolver, BACKSCATTERSOLVER_ENUM_VALUES);
```

.cc file:
```cpp
ENUMARTSIMPL(BackscatterSolver, BACKSCATTERSOLVER_ENUM_VALUES);
```

> Shouldn't the `static const` make this a single execution?

It creates one static instance of the strs variable in _each_ function.

riclarsson:
Can't we inline it?  I don't like the multiple definitions.

Here's a slightly more readable version of before:

```
template<typename T> constexpr bool good_enum(T x)
{
  return Index(x) < Index(T::FINAL) and Index(x) >= 0;
}

template<typename T> auto enum_strarray(const String& strchars) -> 
std::array<String, Index(T::FINAL)>
{
  std::array<String, Index(T::FINAL)> out;
  std::istringstream x(strchars);
  for (Index i=0; i<Index(T::FINAL); i++) {
    std::getline(x, out[i], ',');
    out[i].erase(std::remove(out[i].begin(), out[i].end(), ' '), out[i].end());
  }
  return out;
}

#define ENUMARTS(ENUMTYPE, ...)                                               \
  enum class ENUMTYPE : Index {__VA_ARGS__, FINAL} ;                          \
                                                                              \
  static_assert(not good_enum(ENUMTYPE::FINAL), "FINAL " #ENUMTYPE " good");  \
  static_assert(good_enum(ENUMTYPE(0)), "first enum FINAL for " #ENUMTYPE);   \
                                                                              \
  namespace enumstrs {                                                        \
  static const auto ENUMTYPE = enum_strarray<::ENUMTYPE>(#__VA_ARGS__);       \
  };                                                                          \
                                                                              \
  inline String to_string(ENUMTYPE x) noexcept {                              \
    if (good_enum(x))                                                         \
      return enumstrs::ENUMTYPE[Index(x)];                                    \
    else                                                                      \
      return "BAD " # ENUMTYPE;                                               \
  }                                                                           \
                                                                              \
  inline ENUMTYPE  to_ ## ENUMTYPE(const String& x) noexcept                  \
  {                                                                           \
    for (Index i=0; i<Index(ENUMTYPE::FINAL); i++)                            \
      if (enumstrs::ENUMTYPE[i] == x)                                         \
        return ENUMTYPE(i);                                                   \
    return ENUMTYPE::FINAL;                                                   \
  }
```

Which also only keeps a single copy of the string-array outside of the global namespace.  I put this in rte.h and it compiles as intended.  I think inline-ing is easier than using multiple definitions, even if compile time is damaged a bit.

olemke:
> Can't we inline it? I don't like the multiple definitions.

Since the function are rather small anyway, that is even better.

olemke:
> Presently, pyarts.classes.from_workspace(x) takes a workspace variable and turns it into a python class with full interface. It would be nice if this was part of the normal WorkspaceVariable workflow, like how "ws.f_grid.value" works. But I do not know what propery to occupy in WorkspaceVariable.

That's probably better answered by @simonpf, but what I like about your current approach is that the whole classes implementation is confined to its own submodule. Maybe we should keep it like that for now and see how it works before interweaving it with other parts.

riclarsson:
Lets keep things as they are then.  The enum-issue should perhaps also be delegated to another commit, where the enum-interface to python is also fixed.

PR #148, opened 2020-03-27
Improve error message for missing Python packages

olemke:
Error message for missing packages now looks like this:
```
cmake .. 
-- The C compiler identification is GNU 9.2.1
-- The CXX compiler identification is GNU 9.2.1
[SNIP]
-- Found PythonInterp: /home/olemke/anaconda3/bin/python3 (found version "3.7.6") 
-- ERROR: Required Python package not found: docutils

-- ERROR: Required Python package not found: lark-parser

-- ERROR: Required Python package not found: matplotlib

-- ERROR: Required Python package not found: numpy

CMake Error at cmake/modules/CheckPythonModule.cmake:47 (message):
  Please install the missing Python package(s)
Call Stack (most recent call first):
  CMakeLists.txt:58 (check_python_modules)


-- Configuring incomplete, errors occurred!
See also "/home/olemke/Hacking/uhh/arts/build/CMakeFiles/CMakeOutput.log".

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,17 +55,7 @@ set(Python_ADDITIONAL_VERSIONS 3)
 find_package (PythonInterp REQUIRED)
 
 include (CheckPythonModule)
-set(REQUIRED_MODULES
-  docutils
-  lark
-  matplotlib
-  netCDF4
-  numpy
-  pytest
-  scipy
-  setuptools
-  )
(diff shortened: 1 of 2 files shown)


riclarsson:
I like the new error message.  What happens to people that got lark installed by mistake?

olemke:
Oh, right, the PyPi lark package also provides a lark module. Wow, what a mess. I'll add a check for that as well.

PR #160, opened 2020-04-20
Templates to help methods.cc

riclarsson:
This adds two templates to make dealing with methods.cc additions easier.  I really do not follow the old approaches errors.  This new way should make the function fail at compile time if it is not good (although several things are missing, like checking of "\n" at the end of the description.)

### DIFF
--- a/src/array.h
+++ b/src/array.h
@@ -120,6 +120,9 @@ class Array : public std::vector<base> {
   Array(std::initializer_list<base> init)
       : std::vector<base>(init) { /* Nothing to do here. */
   }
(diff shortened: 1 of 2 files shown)


olemke:
I like this version including the static_asserts. Getting compile-time errors if something is wrong in the WSM definitions is definitely better than before. Since it touches every WSM in methods.cc, I'll check if @erikssonpatrick or @simonpf are currently working on any WSMs. I don't want to create conflicts on their end. We have a video meeting soon and I will merge this afterwards if it's ok with them.

Issue #157, opened 2020-04-13
Raw data in ARTS


riclarsson:
With the OEM module in place, I think it would be a good idea to have the ability to process raw-data in ARTS to generate lists of `y` and `covmat_se`/`covmat_seps` from some raw-time series.  (The latter does not exist today, since `covmat_se` is defined to contain it.)

Is this an idea accepted by others?  Or do you want to keep rawdata processing out of ARTS still?  I think it fits to have it given the addition of OEM.

My design of this would be very simply in the beginning but could of course evolve if it helps other observation schemes.  An agenda would be created, called `rawdata_agenda`.  This agenda outputs a `ybatch`, an `ArrayOfCovarianceMatrix` `covmat_sepsbatch`, and an `ArrayOfIndex` `ybatch_count`.  It takes only a `Matrix` `rawdata` as input.  The methods inside this agenda could then require more information.

In the original implementation I intend to use this for I have a simple repeating cold-atm-hot-atm measurement cycle.  The method for generating `ybatch` from this would be called `ybatchFromRawdataCAHA` and require an additional `hot` and `cold` temperature `Vector` input.  (Just to keep the start of this simple.)  Additionally, a `TimeAverageBatchCalc` method would be created that takes a `time` `Vector` and a `time_target` `Numeric` to reduce the size of `ybatch` and generate `ybatch_count` and `covmat_sepsbatch` at the correct sizes.

erikssonpatrick:
I have probably some knowledge of the calibration routine of 6-8 radiometers, and they all differ. I have never seen anyone just using single hot and cold measurements to produce spectra. There is always some averaging of hot and cold, at least averaging of the closest two of each. There could also be observation cycles, observing one load more frequently than the other. There could be one or several atmospheric views for each calibration view. For ground-based, a tipping curve procedure could be involved. For satellites, non-linear correction terms could be applied.

So despite it would make sense if ARTS could handle the complete chain from L0 to L2, I just don't think this can be standardised sufficiently to make it fit ARTS without causing confusion in the end.

Is this not something for Python? To make calibration routines for different instruments. Maybe this can be compared to reformat the output of OEM (the x vector) to the L2 format around different sensors. This is again nothing can be standardised and fits better for keeping in Python, Matlab ...

So I take the liberty to close this one.

riclarsson:
I will go ahead and open this issue again because you answer seems to miss somethings Patrick.

Can we agree that ARTS should be able to handle everything from L1 to L2?

[...]

PR #164, opened 2020-04-27
Arts simple raw

riclarsson:
Adds a very simple set of scripts to go from raw measurements of cold-atm-hot to calibrated brightness temperatures.  Also adds a port of python code for time averaging and computations of time averaged spectra.  Also adds a port of the most simplistic type of tropospheric corrections.  This addresses: #157 .  The new time variable is in the python classes with some tests and simple arithmetic.  By using datetime as a basic interface, matplotlib integration works quite good at this point.

@erikssonpatrick: Are you OK with this very basic start?


olemke:
@erikssonpatrick Should I go forward and merge this?

erikssonpatrick:
Simple no free time to look into the details. Richard and I discussed this, and a general OK for introducing this. My main worry is documentation. Please make sure that the built-i doc is fairly easy to understand. And that there is a test/demo case.

riclarsson:
The new tests in python-controlfiles/raw should make this enough to merge now.  They should make clear how to use the code and what it does, together with the updates to methods.cc

PR #179, opened 2020-05-27
Fix undefined return value

### LINKED ISSUE #178: OEM errors are not distinct (and can cause seg faults in lm mode)
**Describe the bug**
The reason OEM does not work is sometimes very unclear.  In the attached file, a simple test case runs through some code trying to retrieve water vapor in a simplified setup from some data.  The code fails.  I am sure it fails because the inversion of the measurement error covariance matrix is not very good for this particular case,  This is reported simply as "Error during OEM computation.", which is somewhat vague.  Should we perhaps introduce a "oem_checkedCalc" function, and let this deal with error messages?

Anyways, there is a larger problem here as well.  If I run OEM again after seeing the std::cout-error, I get a segmentation fault.

**To Reproduce**
Unzip this file:
[tmp.zip](https://github.com/atmtools/arts/files/4634909/tmp.zip)

Change the XMLDIR variable at the top of the python file to a string pointing at your own arts-xml-dir.  Run to see the segmentation fault.  If you run while changing "lm" to "li", again in the settings the oem_model variable, then the error persists but the segmentation fault is not there, so it looks like a "lm" problem.

If applicable, error output:

```bash
                                MAP Computation                                
Formulation: Standard
Method:      Levenberg-Marquardt

 Step     Total Cost         x-Cost         y-Cost    Conv. Crit.   Gamma Factor
--------------------------------------------------------------------------------
    0   -6.30048e+17              0   -6.30048e+17                    100.000000
--------------------------------------------------------------------------------

Error during OEM computation.

                                      ----                                      


                                MAP Computation                                
Formulation: Standard
Method:      Levenberg-Marquardt

 Step     Total Cost         x-Cost         y-Cost    Conv. Crit.   Gamma Factor
--------------------------------------------------------------------------------
    0            nan            nan   -6.30048e+17                    100.000000
Segmentation fault (core dumped)

```

**Expected behavior**
Either tell me the error in clear text in a throw/raise call, or ensure it is dealt with quietly

**System info (please complete the following information):**
 - OS: Ubuntu 20.04

olemke:
Fixes #178

### DIFF
--- a/3rdparty/invlib/src/invlib/map.cpp
+++ b/3rdparty/invlib/src/invlib/map.cpp
@@ -63,6 +63,7 @@ auto MAPBase<ForwardModel, MatrixType, SaType, SeType, VectorType>
         if (robust) {
             return std::numeric_limits<RealType>::max();
         }
+        throw;
     }
 }
 


3rdparty/invlib/src/invlib/map.cpp
```diff
@@ -63,6 +63,11 @@ ::cost_function(const VectorType &x,
         if (robust) {
             return std::numeric_limits<RealType>::max();
         }
+        std::throw_with_nested(
```
simonpf, line 66: IMO nesting the error once more is not necessary because that's already done in the `evaluate(...)` call.

simonpf:
Good job! Thanks for catching this!

3rdparty/invlib/src/invlib/map.cpp
olemke: Good point. I changed it to just rethrow the exception.

PR #182, opened 2020-06-05
Jacobian changes

riclarsson:
This commit begins a change after a discussion we had a few years ago to try and remove the `String`s from retrieval quantity variables, and to remove the idea that some `RetrievalQuantity` requires "`Integration`" and "`Analytical`" to work just because of old boilerplate.  The former is removed and the latter is not by this commit.

The replacement code basically lists a bunch of possible partial derivatives in different namespaces (that have a few auto-generated names with them).  There are four of these namespaces "`Special`", "`Sensor`", "`Line`", and "`Atm`".  These are all further encapsulated by the "`Jacobian`" namespace.  To select namespace a further "`Jacobian::Type`" is also checked to see what the type represents.  To make it easy to work with, the sub-types are unionized.

The first subtype (`Special`) is just for Tag-based VMR, since it introduces a very special exception to how calculations are done compared to the other types.  The second (`Sensor`) is basically empty at this point because I do not know how the `Sensor` Jacobians work.  `Atm` and `Line` contains everything that is done in the absorption cross-section calculations analytically (except, for the fact that our interpolation routines still do not have local derivative calculations...).  `Line` contains the other method for computing "`VMR`", since it is based on species rather than tags and are thus identified across different tags.

The goal, to finalize, is to be able to remove these five variables from `RetrievalQuantity`:
```
  String mmaintag;
  String msubtag;
  String msubsubtag;
  String mmode;
  Index manalytical;
```
the calculations for these should be understood by the enum-tags introduced here (once `Sensor` has been populated by names people can understand) and perhaps an enum-modification variable in `RetrievalQuantity` for things like hydrostatic balance.  To achieve this goal, someone familiar with `Sensor`-style Jacobian has to have a look at this.

@erikssonpatrick Can you find time to look at this?  The goal we talked about before basically implies we want to be able to incorporate the logic of `rtmethods_jacobian_init` into `RetrievalQuantity`.  The function should be removed, and both the `FOR_ANALYTICAL_JACOBIANS_DO` and `FOR_ANALYTICAL_JACOBIANS_DO2` variables should be removed.  (Perhaps adding a surface-type enum class is required here as well, this is another one of those things in ARTS I do not use so I don't know how many types of surface variables are required.)


python/pyarts/classes/JacobianTarget.py
```diff
@@ -0,0 +1,178 @@
...
+            self.__delete__ = True
+            self.__data__ = c.c_void_p(lib.createJacobianTarget())
+            if data is not None:
+                raise RuntimeError("Only supports void initialization")
+
+    @property
+    def type(self):
+        """ Enumeration type of Jacobian (String) """
+        tmp = String(c.c_void_p(lib.enumgetTargetTypeJacobianTarget(self.__data__)))
+        tmp.__delete__ = True  # Manual delete not on by default for pointers!!!
```
riclarsson, line 44: Is there a more pythonic way to deal with this @olemke ?  Like, copy-construction, but I need the __delete__ variable to no interfere with proper references...

olemke: To avoid interfering with the internal `__delete__` attribute from outside the `String` class, it is cleaner to add an optional argument to the constructor of the `String` class. This gives the caller the option to give a hint to the `String` class how the pointer memory should be treated:
```python
class String:
    def __init__(self, value, free_memory=False):
        if isinstance(value, c.c_void_p):
            self.__delete__ = free_memory
            self.__data__ = value
        else:
            self.__delete__ = True
            self.__data__ = c.c_void_p(lib.createString())
            self.val = value
```
Adding a copy method inside the `String` class also helps with readability:
```python
    def copy(self):
        return String(str(self));
```
Then, the `type` getter in `JacobianTarget` would look like this:
```python
    @property
    def type(self):
        """ Enumeration type of Jacobian (String) """
        return String(
            c.c_void_p(lib.enumgetTargetTypeJacobianTarget(self.__data__)),
            free_memory=True,
        ).copy()
```

erikssonpatrick:
Richard,

It's great if you have time and energy to attack this. Clearly would be good for the future. Time? I have had a quick look now and could hopefully provide advice, but no time for actual coding before the summer vacations start. Frustrating, it just gets harder to find time for ARTS coding.

Comments suggestions:
* "Special" is very vague. Call it "AbsSpecies" or similar.
* Yes, "Surface" is needed. The information that has to be passed on is a free format string.
* For sensor variables, the size of the perturbation must be passed on. The rest should be encoded by the method added to jacobian_agenda. But I could be missing something and some more information could be needed.
* As the abs and scat species work quite differently, maybe a good idea to introduce "ScatSpecies" from start. For the moment there are two free text strings that should be passed on. But Simon and I could introduce more versions.

 I hope this provided some help ...

/P

riclarsson:
@erikssonpatrick 

The point we made before was to remove all strings because we thought they cause some minor slowdowns but mostly because having a stricter enum-system means you can more easily switch over the different types.  This, and removing the "Analytical" Index-flags, will probably make the code easier to read and follow.

- I will leave it as part of `Special`, it might be joined by other properties that must be treated in a weird way.  I will switch "TagVMR" to "ArrayOfSpeciesTagVMR" to make it clearer what it is doing that is considered special.  I will also move all the Wind components and the Magnetic magnitude component into `Special`, since they also are computed in a weird way compared to the rest of the variables.
- If you truly need to parse a free-format string then something is wrong.  Are you truly parsing strings in your calculations?  Or just using them as names?  Because the latter means you don't need them.  Instead, the surface parameter (be it wind speed, albedo, or whatever the models need) , should be placed in an enum so that `Surface::WindSpeed` is part of the output derivatives of that method.
- You need perturbations to be numerical or index or both for the sensors?  The former is in there already because we don't have in position derivatives from the interpolation routines yet.
- How are they different?  I can imagine if you want to use "ArrayOfSpeciesTagVMR" that you might want to have instead "LinearArrayOfSpeciesTagVMR", "QuadraticArrayOfSpeciesTagVMR", and so on for scatterers.  Is that it?  Would it be enough to just have an index to say what order the VMR influences the absorption at instead?

erikssonpatrick:
Richard,

What you have to understand for particles and surface is that there is no standard way to describe their properties. It is completely different from gases. Expressed simply, if you specify O3-666 you know what you have to handle. But for particles, there is not even agreement on how to describe their size, or what you mean with IWC. In any way, there is no fixed set of parameters to use an trying to squeeze this into an enum system would just cause extra work for developers and users.

This also means that a scattering species has no fixed meaning. It could be single particle, a mix of particles having the same basic shape but varying size, a totally general mix of particles or bulk properties.

Efficiency is no issue on this side as the strings are parsed very few times. 

I discussed this with Simon and he agrees. 

Bye,

Patrick

riclarsson:
The efficiency of checking equality between two `String` is definitely not important.  My main goal is to be able to stop using `jacobian_quantites` in xsec and propmat-agendas since it cannot be saved and read reliably to and from files.  The new Target class makes that easy and allows you to continue experimenting with `RetrievalQuantity`.

This will never cause extra work for the user, there will always be simple "AddJacobianType"-style functions as today, just that the strings input maps to internal calculations type.  The internals simply just uses an enum for a set of predefined styles of calculations instead of the `String`.

Anyways, since you can only handle a single type of Jacobian (i.e., the string-parsing type), this makes things easy.  Both surface and scattering are simply `Special`.   If anyone ever adds a method that has a predetermined set of calculations, they are free to extend the general `Type` to include this at that point without touching the `Special::ScatteringString` and `Special::SurfaceString` interfaces.  (The `String`-part of the name must be there to be clear that a `String` of some value(s) has to be parsed in the calculations to understand the derivative.)

olemke:
Merge conflicts are resolve. @erikssonpatrick Should it go into master now?

erikssonpatrick:
If checks are running I don't see any problem. So OK from me.

PR #183, opened 2020-06-11
Linemixing HITRAN

riclarsson:
This is a port of HITRAN linemixing code to ARTS.  It is absolutely horrible C++ code but I think it emulates their intent well.  To run this code, you have to be in the same folder as the `data_new` directory from the officially [released `Supplemental_LM` package](https://hitran.org/supplementary/).  The code is a port of their `LM_calc_CO2_2017.for`.  The current implementation is very basic and only allows for playing around.

@stefanbuehler, @m-brath or @olemke , can one of you look over the code to see if the port misses something obvious?  I think I am following the Fortran implementation faithfully but I am going blind looking at the Fortran code at this point so I do not know...

I need some verification that it is probably "OK" before I go ahead and integrate it with common ARTS features.  I am especially worried about the way I ignore and work around the Lapack linear algebra.  I don't know how Lapack works here, and it might be that their methods produce different results...

Anyways, currently the VP/SDVP output looks OK, the Full output is also mostly OK bar a few exceptions, but the Rosenkranz VP/SDVP output is a complete mess...


olemke:
Since the HITRAN LM code comes with an example program, I wonder, have you tried running the original code and the ported code with the same parameters? That would be a good first step to verify the correctness of the new code.

P.S.: About the log message in your commit bea7aee : Note that Fortran uses column-major storage whereas C/C++ use row-major. Just wanted to mention this because the message suggests the opposite.

riclarsson:
The LM HITRAN code does not compile for me.  It would be very useful if it did.  They are not following any Fortran standard as far as I can tell, and basic tools that translates Fortran to C (i.e., f2c and fable.cout) were unable to produce code I could understand.

About rowmajor/columnmajor storage.  I see, OK, I should switch to just use default Eigen then.  Anyways, I need to know whether the Fortran LAPACK routines produce the same style matrix as the much simpler-to-use Eigen stuff...

olemke:
Try the legacy option: `gfortran -std=legacy -o LMcalc LM_calc_CO2_2017.for`

riclarsson:
@olemke  Thanks for the help.  This works very well now for SDVP and Rosenkranz parameters.  I still need to investigate the eigenvalue decomposition.  The eigenvalues produced seem the same but the order of the values are output differently.  This of course is supposed to be OK from a mathematical POV but the methods seem to want some strict ordering by what is written from lines 938-944.

If there is any experience from sortings of eigenvalues and vectors here it would be very useful, otherwise I will have to try and remember my linear algebra in more details...

olemke:
Answering some questions/remarks hidden in the commit messages:

1. Is it a good idea to switch to use one of the versions of Eigen on GitLab?

Which version are you referring to? 3.3.7, the one you added now, is also the latest release available on their gitlab repo. They haven't released a stable version for quite some time, but I couldn't find any information on their site that  recommends to use the 3.3 branch instead.

2. [...] there is an error that is probably caused by some division
near zero.  I do not know why the Fortran code does not experience
the same issues.

On macOS, the Fortran code completes but also gives me an underflow warning:
```
./LMcalc
[snip]
Note: The following floating-point exceptions are signalling: IEEE_UNDERFLOW_FLAG IEEE_DENORMAL
```
So this issue seems to also be present in the original code as well.

riclarsson:
@olemke 

The printed errors are the same for me for the Fortran code.  There is apparently a Fortran standard from 2008 or something that says all IEEE errors must be printed to screen when a "stop" statement is reached.  I guess the gfortran people simply lets it be on for all the "-std".  This is not the main problem.  The Fortran code produces reasonable output for the absorption by somehow handling this.

The C++ port produces unreasonable output sometimes.  Values should be in the range (0, 10) but can have negatives and (almost) positive infinity  I am not sure why.  So I removed the offending code for now.  It only works at high pressures now.

olemke:
Just a heads up to not compile ARTS with "-DENABLE_FORTRAN=1" if you want to save some debugging time. I got segmentation faults in `test_propagationmatrix` in one configuration, but it ran fine in another. As it turns out, if FORTRAN is enabled, `lapack::zgetrf` will not call into the system Lapack but instead call `zgetrf` from our own tmatrix FORTRAN module. I'll work out a fix for that and other possible Lapack name clashes.

riclarsson:
@stefanbuehler @m-brath @olemke 

This essentially works now.  The SI version, the CGS version, and the HITRAN version produce essentially the same results.

You can test it out with the test_propagationmatrix script from the same folder you would run the Fortran code.  There are a few missing bits still, mainly setting up the checks for the isotopologue's better.  But these are minor things and since the code runs and works I am not sure how to do the tests here correctly.

Anyways, I still need to design the interface for "methods.cc" so this can be used normally.  I am not entirely sure how to do that.  I have currently encapsulated everything that I need into a single struct that is populated from file.  It contains everything, including the line catalog.  I am inclined to just use it as it is as the workspace variable.  It is a bit messy though, since it also encapsulates the line catalog.  This means it will not be easy to kick-out included lines from the normal calculation flow.  Are there any other suggestions?

The struct is at lines 32+ in "linemixing_hitran.h".  It is populated by the "read" function in the same file and used for computations by the "compute" function that takes it as an input.  (The other "compute" function just does the CGS calculations).

stefanbuehler:
Dear Richard,

should not this work well with your new catalogue features? Especially 
that you can hold lines that belong to one band together. Why do you 
have to hardwire the catalogue?

Best wishes,

Stefan

riclarsson:
Dear Stefan,

It can be done that way but it will require some more changes.  Most importantly, it will demand that the user knows how many Wigner symbols they want to compute.  The Wigner library we have got is good and fast enough to do this, but the UX is really bad.  The original library has a lot of "exit(1)" statements in it.  I do not wish to change these, but we must perhaps wrap a check in more user-friendly terms.

With hope,
//Richard

riclarsson:
This can now be merged.  It produces the results we want it to.

test

Here's the absorption of the new method and the old with the same inputs.  "LM test 1" is the full relaxation method in test_propagationmatrix, "LM user method" is the same input to the user method, "Old VP" is loading all of HITRAN CO2 in xml-data/cat/ and running the line-by-line code that way, and "LM test 2" is the Voigt implementation without line mixing in the new code.

You activate this by running
```

ws.abs_hitran_relmat_dataReadHitranRelmatDataAndLines(basedir="LM/data_new/")
ws.propmat_clearskyAddHitranLineMixingLines()

```
where the basedir is wherever your copy of the line mixing data resides.

olemke:
Since all code references to the relmat interface have been removed, I would like to also delete the empty `src/linemixing.{cc,h}` files and the unused `3rdparty/relmat/` module. Can I go ahead with that @riclarsson?

riclarsson:
@olemke Sure,  I think I need to make one more commit to this though, because the absorption should be negative in parts...

olemke:
I have also some other commit with minor changes to the WSM docs. I'll commit those when you're ready.

Issue #185, opened 2020-06-23
How to get the species name from the AbsorptionLines class in pyarts?


olemke:
Hi @riclarsson,

I'm trying to read an ARTS catalog with the new pyarts classes interface and would like to output the species and some parameters of each line. Accessing f0 and i0 is no problem and I'm sure I'm just completely blind here, but how do I access the species name?

```python
import pyarts

ws = pyarts.workspace.Workspace()

ws.ReadHITRAN(filename="HITRAN2016.par", fmin=1e9, fmax=1.1e9)
ws.WriteXML("ascii", ws.abs_lines, "hitran.xml")

aalines = pyarts.classes.ArrayOfAbsorptionLines()
aalines.readxml("hitran.xml")

all_lines = [
    {"f0": line.f0, "i0": line.i0} for alines in aalines for line in alines.lines
]

all_lines.sort(key=lambda x: x["f0"])
for line in all_lines:
    print(line)
```

The species name is in the xml file as an attribute, but I couldn't find a way to access it through the `AbsorptionLines` python class.

riclarsson:
There is no way to do this.  We do not store the species name anywhere in the class and therefore I did not make an interface for this.  It is somewhat straightforward to implement the interface but I will not do this before I get the line-mixing stuff running with a user-interface. There are two options:

1. There is a function `SpeciesName()` in `arts::AbsorptionLines` that does the conversion.  The way to port this would be to just make an interface to that function, returning a `new` `String` as a `(void *)` and then ensure that a `python.classes.String `that sets `__delete__` returns it.  This is what I do for the enum-types in the open Jacobian commit, but it needs a better interface for the `String`.
2.  A possibly more difficult but definitely more useful alternative is to create an interface to the ARTS array of `SpeciesRecord`.  Then you could create a "SpeciesName" in `pyarts.classes.AbsorptionLines` that simply returns the correct string based on the species and isotopologue..

olemke:
I went for option 1 right now. Was a good opportunity to get a bit more familiar with your classes interface. We should revisit option 2 in the future.

riclarsson:
Excellent.  After a bit of reading, you actually even took care to ensure that the class is OK, which I ignored.  So yours is better.  Ignore my commit.  I think it is perhaps important to point out that using "delete" manually might cause segfaults, and this is something you have to be very careful about.

olemke:
Sorry that you wasted some time on implementing this as well. After your first comment I didn't expect you to work in this that quickly.

PR #187, opened 2020-06-23
Add species name output and unsafe_copy String handling

riclarsson:
This should be enough for #185


riclarsson:
I will close this since @olemke just also committed the same code

PR #188, opened 2020-06-24
Add prefix to all tmatrix LaPack functions

olemke:
This resolves a naming clash that can lead to unexpected behaviour and
crashes. If FORTRAN support is enabled, the LaPack functions included in the
tmatrix module take precedence (even in the C++ code) over the system LaPack
library.

This was noticed in Richard's new linemixing implementation.
Segmentation faults only occured when FORTRAN support was enabled.

Since modifying tmatrix to use the system LaPack library might also have
possible sideeffects on the tmatrix calculations, the tmatrix implementation
is kept, but all its LaPack functions were renamed to carry the prefix TM in
their name. tmatrix also comes with a quadmath implementation of its
LaPack functions which we want to keep for backwards compatibility.


riclarsson:
A short question: is it not better to just use lapack by binding it to the tmatirx package directly using the C-bindings? It is a bit silly to repeat the code again instead of relying on available packages

olemke:
@richard Please re-read the last paragraph in my [commit message](https://github.com/atmtools/arts/pull/188#issue-439225536). It explains why I did what I did.

riclarsson:
Right, I didn't see the last paragraph for some reason yesterday.  It makes sense then.

PR #191, opened 2020-07-08
Add missing no-docserver guard

riclarsson:
I tried to get rid of the GhostScript error mess so that I can read true compile errors instead of nonsense errors but disabling docserver did not help...

Found this bug though!

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -808,7 +808,7 @@ int main(int argc, char** argv) {
 
   // Now we are set to deal with the more interesting command line
   // switches.
-
+#ifdef ENABLE_DOCSERVER
   if (parameters.check_docs) {
     // Check built-in docs and then exit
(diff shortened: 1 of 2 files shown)


olemke:
Good catch. Although I'm not following your logic why disabling the docserver should get rid of ghostscript errors. The docserver doesn't use ghostscript. Ghostscript messages can only come from LaTeX in doc/uguide. If you want to concentrate on compile messages and save building time, don't build everything, just run `make arts` instead.

riclarsson:
I see, I will try to disable latex from the builds then.

I didn't know what ghostscript was doing.  Anyways, I tend to put my stuff in "test_*" files while implementing things so I want to just use a default build-argument so that I don't have to change it depending on what I touch.

olemke:
Then you can use `make -C src` to build everything in the src directory.

Issue #198, opened 2020-08-27
abs_linesReadSpeciesSplitCatalog should trow error if file does not exist


stefanbuehler:
Right now abs_linesReadSpeciesSplitCatalog just returns empty lines if no catalogue files are found (for example if the path is mistyped).

I think it should trow a runtime error if no file exists for a species tag that is a line-by-line species. (This type of silent failure leads to bugs that are hard to find.)

Optionally, in case anyone needs the behaviour of silent failure for a script, there could be an input argument "robust" with default 0, that triggers the silent behaviour if set to 1. This would be consistent with DOBatchCalc.

PR #199, opened 2020-08-27
Add robust flag and throws by defauly in case there are no bands

riclarsson:
Fix for #198


stefanbuehler:
Hi Richard,

great! Could you please do the same for abs_lines_per_speciesReadSpeciesSplitCatalog?

Sorry, I discovered only afterwards that that is the even better reading function for my case. 

Cheers

Stefan

stefanbuehler:
Thanks!!

PR #216, opened 2020-09-19
Added missing BRANCH_NAME

erikssonpatrick:
Added to: git -pull rebase origin

(Just wanted to fix a minor thing CONTRIBUTING.md. Did not expect that this should end up with a pull request!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,7 +86,7 @@ If you are using different computers and pushed changes to your fork on one, you
 can update your local branch on the other computer with:
 
 ```
-$ git pull --rebase origin
+$ git pull --rebase origin BRANCH_NAME
 ```
 
 ## Pull request


olemke:
Technically, it is not necessary to specify the branch name for the pull. When checking out a branch locally, the corresponding upstream branch is set as the default tracking branch automatically, e.g.:

```
$ git checkout -b dummy origin/shtns-support
Branch 'dummy' set up to track remote branch 'shtns-support' from 'origin'.
Switched to a new branch 'dummy'
```
But since being explicit here might can problems depending on the user's configuration, it's probably best to have the full command in the docs.

> (Just wanted to fix a minor thing CONTRIBUTING.md. Did not expect that this should end up with a pull request!

I can add you as a maintainer if you want (and are fully aware of the potential hazards ;-) ). That gives you direct commit access to the master branch.

PR #218, opened 2020-09-22
Make ARTS linkable

riclarsson:
This is purely a infrastructure change so that it is possible to simply
put ARTS in another projects 3rdparty/ folder and in that project do
target_link_libraries(arts_interface PRIVATE artscore methods matpack cdisort)

I have yet to actually use ARTS in a 3rdparty project this way, and I am sure I
will have to do many things to make it work, but at least I can trick CMake
into linking and building everything from 3rdparty projects this way.

The main changes have been to simply switch away from using CMAKE_SOURCE_DIR
and CMAKE_BINARY_DIR to ARTS_*_DIR, since "project" defines these two.

NOTE:  There is a potential bug with pyarts in the current setup.  The
python CMakeLists.txt was renaming ARTS_BINARY_DIR away from the defauly
CMake name.  I think that this is very poor design, but I am not sure how
to fix it and what problem removing this renaming has caused... so
I submit the code but this must be fixed before it is merged


python/CMakeLists.txt
```diff
@@ -7,14 +7,13 @@ add_subdirectory(doc)
 file(GLOB_RECURSE PYTHON_SRC_FILES
   RELATIVE ${CMAKE_CURRENT_SOURCE_DIR}
   ${CMAKE_CURRENT_SOURCE_DIR}/*.py)
-set(ARTS_BINARY_DIR ${CMAKE_BINARY_DIR/src})
```
riclarsson, line 10: Here is the potential bug!  @olemke  or @simonpf , where do you use this?  I think it is a bad idea to overwrite standard CMake variables, but I suspect you need this to link the .so file?

simonpf:
Hi Richard,

I think it will be easier and more usable to use CMakes FindPackage functionality to achieve the aim that you
describe. With this you will be able to include ARTS in other projects without duplicating the code. Have a
look how Eigen handles this.

Moreover, if you intend to include the ARTS code within another project (basically the usecase you describe here),
I think the changes  to the build system should be limited to the shipped version and not affect the upstream version of ARTS.

riclarsson:
Hi Simon,

I mean, this is only some minor changes that will make it easier to use in 3rdparty/ code.  It now works to link to the artscore functions, which is what I wanted to achieve.

The overwrite of ARTS_BINARY_DIR must be fixed regardless, so I will ignore that bit.

Is there any good arguments to make it difficult to link otherwise?

olemke:
While I agree with the use of the FindPackage functionality being more appropriate (but also currently not available for ARTS), I do think Richard's changes make sense. Using the `ARTS_*_DIR` variables instead of the `CMAKE_*_DIR` doesn't hurt and increases compliance with CMake conventions.
Not overwriting standard CMake variables such as `ARTS_BINARY_DIR` seems also sensible. And the changes don't break anything a far as I can see. I think mainly because the default value for `ARTS_BINARY_DIR` is identical to what it was explicitly set to anyway.
The failing test in the `ubuntu-clang-9-nofortran` config doesn't see related to these changes.

olemke:
I retriggered the build jobs to see if `ubuntu-clang-9-nofortran` consistently fails.

simonpf:
Excuse my ignorance, but which conventions would using ARTS_*_DIR instead of CMAKE_*_DIR comply more with? To me the use of CMAKE_*_DIR seems much more intuitive, but again this may be because I have missed something.

I still do think that to link against ARTS, FindPackage is the way to go. With this you can have a single version of ARTS on your system and link against that one as opposed to having to copy and build ARTS for each project you are using. This is exactly why the concept of software libraries has been invented.

riclarsson:
I don't think that ARTS should be in the same category as BLAS here.  I
need the version I use to be stable so I can work on multiple machines, and
the best way to do that is to just get a copy and stick with it.

The convention we talk about is that Cmake sets some variables on a project
basis. Literally on calling project. Overwriting those is just bad. Using
them helps me, so that is why I committed the change.

Now you are probably correct that FindPackage is better. It is definitely
not as easy as just copying the entire folder tree though. I would rather
have easy-to-use over a generic better, unless there's some great
performance hindrance, of course.

olemke:
> Excuse my ignorance, but which conventions would using ARTS___DIR instead of CMAKE___DIR comply more with? To me the use of CMAKE_*_DIR seems much more intuitive, but again this may be because I have missed something.

"Compliance" was not a good choice of words of mine. "Compatibility" would have fit better. As mentioned in the [CMake documentation](https://cmake.org/cmake/help/latest/variable/PROJECT-NAME_BINARY_DIR.html) using `<PROJECT>_*_DIR` is useful when connecting different projects. Since we are already using the `project` command in our top-level `CMakeLists.txt` anyway, I don't see any reason why we shouldn't try to support features it provides.

> I still do think that to link against ARTS, FindPackage is the way to go. With this you can have a single version of ARTS on your system and link against that one as opposed to having to copy and build ARTS for each project you are using. This is exactly why the concept of software libraries has been invented.

This is definitely true for commonly used libraries, but I'm not sure ARTS falls in this category. As it is scientific research code, most people that use it might not be able to install it into the system (since they generally don't have root access on compute clusters). On the other hand, installing a library into a user-defined prefix and dealing with the path setup to link other programs against it (from my experience with users) is mostly perceived as cumbersome and sometimes an insurmountable hurdle to overcome.
So why not support both ways to provide convenience for both use cases. Especially since Richard already provided a working patch for one of them.

simonpf:
Yes, sorry, I wasn't aware of the <PROJECT>_*_DIR variables. So I take back all my objection. Please go ahead with this.

PR #224, opened 2020-09-30
Add pyarts dev version support

olemke:
Add `.dev0` postfix to the pyarts version number if the current minor ARTS
version number is odd. This enables support for having a stable version and
a development version for pyarts at the same time.

`pip install pyarts`

Will install the latest stable version.

`pip install --pre pyarts`

Will install the latest development version.


simonpf (approved):
Looks good as far as my undertanding of PEP440 goes.

PR #227, opened 2020-09-30
Fix pyarts doc

olemke:
This is a first step to get the PyARTS documentation into a working state.


simonpf (approved):
Looks good to me. Thank you for fixing this!

PR #230, opened 2020-10-01
Automatic ARTS interface generation to use ARTS in another C++ program

riclarsson:
This is the interface I hinted at before.  If interest exist to keep this in ARTS, I would be happier to have it here than in external code.

### Definiitions

(EDIT:) To use you need:

#### CMakeLists.txt

```cmake
include (FindOpenMP)
if (NOT ${OPENMP_FOUND})
  message(FATAL_ERROR "No OPEN MP found, ARTS requires it")
endif()

########################################################################################
# ARTS Library: For ARTS usage
add_library (arts_interface STATIC RELEVANT_FILES)
target_link_libraries(arts_interface PUBLIC public_arts_interface OpenMP::OpenMP_CXX)
########################################################################################
```

#### autoarts.h (auto-generated)
```cpp
namespace ARTS {
Workspace init(...);  // Initializes the workspace
namespace Var {...}  // Contains direct access to most Workspace variables
namespace AgendaVar {...}  // Contains access to most Workspace variables and can add Workspace variables
namespace Method {...}  // Contains direct access to most Workspace methods
namespace AgendaMethod {...}  // Returns MRecord of the agenda
namespace AgendaExecute {...}  // Directly executes single Agenda
namespace AgendaDefine {...}  // Defines executes single Agenda
}
```

#### PotentialUse.cpp
```cpp
#include <autoarts.h>
int main() {
  auto ws = ARTS::init(1, 1, 1);

  //  Use a method to print something
  ARTS::Method::Print(ws, Index(1));

  // Set a workspace variable and print it
  ARTS::Var::atmosphere_dim(ws) = 3;
  ARTS::Method::Print(ws, String{"Set atmosphere_dim to:"});
  ARTS::Method::Print(ws, ARTS::Var::atmosphere_dim(ws));

  // Create a new workspace variable
  auto test_num = ARTS::AgendaVar::NumericCreate(ws, 3.14, "TestNumber");

  // Define g0_agenda
  ARTS::AgendaDefine::g0_agenda(
      ws, ARTS::AgendaMethod::NumericSet(ws, ARTS::AgendaVar::g0(ws), test_num),
      ARTS::AgendaMethod::Ignore(ws, ARTS::AgendaVar::lat(ws)),
      ARTS::AgendaMethod::Ignore(ws, ARTS::AgendaVar::lon(ws)),
      ARTS::AgendaMethod::Print(
          ws, ARTS::AgendaVar::StringCreate(ws, "Print inside of Agenda",
                                            "TestString")),
      ARTS::AgendaMethod::Print(ws, test_num));

  // Execute g0_agenda
  ARTS::AgendaExecute::g0_agenda(ws);

  // Print the g0 that has now been set
  ARTS::Method::Print(ws, String{"Print outsjde of Agenda"});
  ARTS::Method::Print(ws, ARTS::Var::g0(ws));
}
```

#### Notes
The entire default namespace of ARTS is present at base.  So things like String and Numeric and Index are defined.  Also, all Workspace methods can be called directly via auto_md.h, but that sort of defeats the purpose of keeping track of Workspace, which is what ARTS does best.  It is up to the user to not have name collisions with ARTS main, but I keep things in namespaces because otherwise it gets silly.

I do not know why the inclusion of OpenMP::OpenMP_CXX is required.  I thought we already linked against it in ARTS but for some reason this linking is not persistent when reaching the interface file...

I require a change in Workspace.  It must be possible to simply push back against the vector of stacks that it is controlling.

It is not too difficult to have things like general_agendas.cc define functions like the Agendas more naturally, like this for the standard emission Agenda (note how it is now the same syntax pretty much as the python interface and normal controlfiles)
```cpp
Workspace& iy_main_agenda_emission(Workspace& ws) {
  using namespace AgendaMethod;
  using namespace AgendaDefine;
  iy_main_agenda(ws,
                 ppathCalc(ws),
                 iyEmissionStandard(ws));
  return ws;
}
```
I did not want to make this the default because I am not sure if we enforce different names between agendas and methods and workspace variables, and I don't want this to become a future issue.

Finally, the autoarts.h file is enormous because it generates all the default documentation in doxygen format for all methods (and, significantly, also for their valid templates).  I think it is worth it to have this documentation in autoarts.h because it makes the editting and writing of new cpp files much easier when a documentation parser is available inline.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -569,5 +569,5 @@ add_custom_target(make_autoarts_h_target SOURCES ${ARTS_BINARY_DIR}/src/autoarts
 # Generate autoarts.h for the ARTS interface
 add_executable(test_cpp_api test_cpp_api.cc)
 target_link_libraries(test_cpp_api public_arts_interface)
-add_test(NAME "arts.cpp_api.fast.silly_groundbased_water_test" COMMAND test_cpp_api > /dev/null)
+add_test(NAME "arts.cpp_api.fast.silly_groundbased_water_test" COMMAND test_cpp_api)
 ########################################################################################
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -564,3 +564,10 @@ add_custom_command(OUTPUT autoarts.h
                    DEPENDS make_autoarts_h)
 add_custom_target(make_autoarts_h_target SOURCES ${ARTS_BINARY_DIR}/src/autoarts.h)
 ########################################################################################
+
+########################################################################################
+# Generate autoarts.h for the ARTS interface
+add_executable(test_cpp_api test_cpp_api.cc)
+target_link_libraries(test_cpp_api public_arts_interface)
+add_test(NAME "arts.test_cpp_api" COMMAND test_cpp_api > /dev/null)
+########################################################################################
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -538,28 +538,29 @@ arts_test_cmdline("workspacevariables" -w all)
 arts_test_cmdline("check-docs" -C)
 
 ########### ARTS Interface ###############
-                                                   
+
 ########################################################################################
 # ARTS Library for including in C++ (must link to autoarts.h to get file)
-add_library (public_arts_interface STATIC autoarts.h)
-target_link_libraries(public_arts_interface PUBLIC ${ALL_ARTS_LIBRARIES})
-target_include_directories(public_arts_interface PUBLIC ${ARTS_BINARY_DIR}
-                                                        ${ARTS_BINARY_DIR}/src/
-                                                        ${ARTS_SOURCE_DIR}/src/
-                                                        ${ARTS_SOURCE_DIR}/3rdparty/
-                                                        ${ARTS_SOURCE_DIR}/3rdparty/invlib/src)
+add_library (public_arts_interface INTERFACE)
+target_link_libraries(public_arts_interface INTERFACE ${ALL_ARTS_LIBRARIES})
+target_include_directories(public_arts_interface INTERFACE ${ARTS_BINARY_DIR}
+                                                           ${ARTS_BINARY_DIR}/src/
+                                                           ${ARTS_SOURCE_DIR}/src/
+                                                           ${ARTS_SOURCE_DIR}/3rdparty/
+                                                           ${ARTS_SOURCE_DIR}/3rdparty/invlib/src)
+add_dependencies(public_arts_interface make_autoarts_h_target)
 ########################################################################################
 
 ########################################################################################
 # Generate the internal structure for the ARTS C++ interface
-add_executable(genautoarts_h genautoarts.cc)
-target_link_libraries(genautoarts_h PUBLIC ${ALL_ARTS_LIBRARIES})
+add_executable(make_autoarts_h make_autoarts_h.cc)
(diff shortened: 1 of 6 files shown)


olemke:
I'll have a look at the macOS build error and see if I can fix it.

riclarsson:
Thank you Oliver,  I suspect it is my piping of autoarts.h, since it is using Linux styles.  But could be anything.  There's still a few things that should be fixed here before a possible merger anyways.

1.  The std::pair idea is not very good since it does not actually set the variable but the pair.  This makes reading files a pain.

2)  My only translated controlfile has a segfault deep in path_stepGeometric...

olemke:
The size of your `libpublic_arts_interface.a` library is 0 (well 8 bytes ;-) ):
```
add_library (public_arts_interface STATIC autoarts.h)
```
There is no point in listing header files in the `add_library` command since they don't generate any symbols. It should be defined as an interface library instead which doesn't create a `.a` file:
```
add_library (public_arts_interface INTERFACE)
target_link_libraries(public_arts_interface INTERFACE ${ALL_ARTS_LIBRARIES})
target_include_directories(public_arts_interface INTERFACE ${ARTS_BINARY_DIR}
                                                        ${ARTS_BINARY_DIR}/src/
                                                        ${ARTS_SOURCE_DIR}/src/
                                                        ${ARTS_SOURCE_DIR}/3rdparty/
                                                        ${ARTS_SOURCE_DIR}/3rdparty/invlib/src)
```

riclarsson:
Without autoarts.h in the interface, there is no autoarts.h generated.  This file has to be generated since it is central to the making the interface controlfiles-like.

olemke:
Creating a custom target that triggers the generation of `autoarts.h` and add it as a dependency to the `public_arts_interface` library should help:
```
add_library (public_arts_interface INTERFACE)
add_dependencies(public_arts_interface makeautoartsh)
....
add_custom_target(makeautoartsh SOURCES ${ARTS_BINARY_DIR}/src/autoarts.h)
```

riclarsson:
Right, this is done Oliver.  I forgot to state that I added "set_main_agenda" as well to copy the C API somewhat.  This did not make a difference, but it is there.

Still, I cannot run my translated controlfile...  In RelWithDebug build mode I get the runtime error "basic_string::_M_create" and in Release build mode I simply get a segmentation fault deep in ppath_stepGeometric.  So there is some weird thing going on.

riclarsson:
It seems that I am not getting the right copies of things from calling deeper agendas.

A standard simple setup that has this chain of Agenda calls:
```arts
iy_main_agenda
  ppath_agenda
    ppath_step_agenda
```

Will yield correct inputs even for ppath_agenda but not for ppath_step_agenda.  In the latter, the first input "atmosphere_dim" is correct, but accessing the vectors that follows give segmentation faults or "basic_string::_M_create" and the Numeric is just a random value...

olemke:
riclarsson/arts#7 should fix the problem with Agenda calls you're seeing.

riclarsson:
Thank you very much Oliver!  It works as well as expected now.  I am not sure how to the test to the standard "make check" (it is fast) but I think it can be merged safely and that it is adding functionality.

I made some minor interface changes, such as demanding a lot of things be done properly in the interface but still allowing somewhat intuitive interface.

olemke:
Test looks fine, it is included in `make check` automatically. You can change the test name from `arts.testcpp` to `arts.fast.testcpp` if you like. One note: Test output should not be redirected to `/dev/zero`. `ctest` suppresses stdout anyway if no error occurs.
I'll merge this after #233 since you're using C++17 features (e.g. nested namespaces).

PR #231, opened 2020-10-01
Add Jekyll config file

olemke:
Prevent Jekyll from hiding directories that start with an _.
Prevents breakage when publishing docs to Github pages.


simonpf:
Isn't a  ```.nojekyll``` file sufficient here?

olemke:
Hmm, didn't know that's a possibility. I'll check. :-)

PR #233, opened 2020-10-06
Use C++17 by default

### LINKED ISSUE #212: Target C++17
If no isssues arise on thunder or mistral, we switch to C++17 as default.

olemke:
Up C++ standard used for compilation from 14 to 17.

Minimum CMake version is now 3.8, required for C++17 support.

Provide cmake option ENABLE_CXX20 for experimental testing.

Closes #212


riclarsson:
The Intel compiler still sets C++14

PR #236, opened 2020-10-06
Added CPP API documentation

riclarsson:
You were a bit too quick with the merge Oliver, I thought I would have time to add some documentation before the merge.

This adds documentation, renames some types, and allow setting some more global states.

I would like someone to comment on if the documentation makes sense.

I am unsure about the rename and would be willing to revert it.  Before ARTS::Var::WorkspaceIndex was a thing.  To be more consistent, it is now called ARTS::Var::Index.  The latter does not interfere with the global Index, and everything inside the Var namespace knows it has to use a Var::Index.  Everywhere else, it is anyways necessary to use Var::Index, so this does not interfere either with Var::Index.  It thus made more sense to me to have as common a tongue as we can have here.  The hope is anyways that when modules are a thing that there will be no access to the global Index, so then the namespaces having their own similarly named types makes a lot more sense than to have the longer WorkspaceIndex abomination of a name.  Of course, this all applies to all Groups.

### DIFF
--- a/doc/uguide/cpp_api.tex
+++ b/doc/uguide/cpp_api.tex
@@ -25,23 +25,16 @@ as an automatically generated ARTS namespace to your project.
 
 To link the public interface, you need to \verb|add_subdirectory(arts)|
 anywhere in your CMake project, where the \verb|arts| directory should contain
-a current version of ARTS.  You also need to manually link OpenMP again.
+a current version of ARTS.
 
 An example (in your projects \verb|CMakeLists.txt|):
 \begin{verbatim}
-###########################################################
-# OpenMP Library
-include (FindOpenMP)
-if (NOT ${OPENMP_FOUND})
-  message(FATAL_ERROR "No OPEN MP found, ARTS requires it")
-endif()
-
-# ARTS Library
-add_library (arts_interface STATIC interface.cpp)
+######################################################
+# ARTS Custom Executable
+add_executable(arts_interface STATIC interface.cpp)
 target_link_libraries(arts_interface PUBLIC
-                      public_arts_interface
-                      OpenMP::OpenMP_CXX)
-###########################################################
+                      public_arts_interface)
+######################################################
 \end{verbatim}
 
 At this point, your \verb|interface.cpp| must have
@@ -105,9 +98,10 @@ The purpose of these types is to pass input to the functions of
 Method and AgendaMethod.  The main way to generate
 instances of these variables is by their corresponding \verb|*Create|
 function or by simply accessing them via their common Workspace
-space name.  The only constructor that is recommended to use is the
+names.  The only constructor that is recommended to use is the
 construction from the corresponding Group, as this can simplify
-the access to several methods.
+the access to several methods.  The other two constructors risk
+accessing out-of-bound memory, or to deference a null-pointer.
 
 It is highly recommended to not discard created variables, as they
 will still occupy memory in the Workspace until the end of the program
@@ -122,9 +116,8 @@ Examples:
   Var::Index y = Var::stokes_dim(ws);
   
   // Create a new index on the workspace
-  Var::Index z{Var::IndexCreate(ws,
-                                            Group::Index{1},
-                                            "new_index_name")};
+  Var::Index z{Var::IndexCreate(ws, Group::Index{1},
+                                "new_index_name")};
 \end{verbatim}
 Note that \verb|x| will not work as an input to AgendaMethod functions
 but it will work as an input to Method functions.  It does not append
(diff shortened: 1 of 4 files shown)


olemke:
No problem. We should make more use of the `in progress` label for PRs to indicate that they're not ready yet to avoid that.
`ARTS::Var::Index` sound good to me and more intuitive than `WorkspaceIndex`.
I'll have a read through the documentation.

doc/uguide/cpp_api.tex
```diff
@@ -0,0 +1,223 @@
...
+include (FindOpenMP)
+if (NOT ${OPENMP_FOUND})
+  message(FATAL_ERROR "No OPEN MP found, ARTS requires it")
+endif()
+
+# ARTS Library
+add_library (arts_interface STATIC interface.cpp)
+target_link_libraries(arts_interface PUBLIC
+                      public_arts_interface
+                      OpenMP::OpenMP_CXX)
```
olemke, line 43: With PR #237 it should not be necessary anymore to call `FindOpenMP` and specify OpenMP::OpenMP_CXX when linking against public_arts_interface.

How about using `add_executable` instead of `add_library` in this example? It's probably more likely that someone wants to build an external program that links against the ARTS library than another library?

doc/uguide/cpp_api.tex
```diff
@@ -0,0 +1,223 @@
...
+
+As a last note.  Several inputs above automatically generates inputs from standard C++ types, such as \verb|VectorNLogSpace|
+generating two \verb|Var::Numeric| from two doubles.  This is a convenient way to use the methods but the user should
+be aware that these methods will end up deleting variables in the end, so some care has to be taken when the scope of such
+automatic variables is long.
+
+\subsection{AgendaMethod, AgendaDefine, and AgendaExecute}
+%--------------------------------------------------------------------------
+\label{sec:cpp_api:usage:agenda}
+The Agenda namespaces deals with setting and defining Workspace Agendas.
```
olemke, line 185: deal

doc/uguide/cpp_api.tex
```diff
@@ -0,0 +1,223 @@
...
+It is only possible to set Workspace Agendas that have been defined as part
+of the Workspace at compilation time.  The AgendaMethod namespace contains
+the same functions as the Method namespace but returns an Internal::MRecord.
+The user of this interface is expected to never manually construct an MRecord.
+Instead, this type is meant to only be used when Agendas are defined in the
+AgendaDefine namespace.  The AgendaDefine namespace defines a variadic 
+function per Agenda in the common Workspace and expects a list of MRecord
+to set this Agenda's methods.  Finally, AgendaExecute exist to execute a single
```
olemke, line 193: exists

doc/uguide/cpp_api.tex
```diff
@@ -0,0 +1,223 @@
...
+Agenda.  Normally, this is not preferred since Agendas should generally just
+be used inside methods, but the option still exist.
```
olemke, line 195: exists

olemke (approved):
Documentation looks good to me. Just marked some typos. Thanks for providing documentation for the interface.

doc/uguide/cpp_api.tex
```diff
@@ -25,23 +25,16 @@ \section{Linking the public interface}
...
-include (FindOpenMP)
-if (NOT ${OPENMP_FOUND})
-  message(FATAL_ERROR "No OPEN MP found, ARTS requires it")
-endif()
-
-# ARTS Library
-add_library (arts_interface STATIC interface.cpp)
+######################################################
+# ARTS Custom Executable
+add_executable(arts_interface STATIC interface.cpp)
```
olemke, line 34: The `STATIC` has to be removed since it's not an option for executables.

olemke:
From my side this PR is ready for master after you merge riclarsson#8 to fix the non-OpenMP build issue. Anything else pending from your side?

riclarsson:
There is plenty of small things that could be done to improve this but there are no show-stoppers.  Some are modular and some are not.  A short list:

- Add `namespace ARTS::Constant { using namespace ::Constant; }` to autoarts.h to ensure that external users always can use the same exact constants.
- Add ARTS::Group::Internal for things like MRecord and Absorption::SingleLine to have a life within the ARTS namespace.  This one is very manual but would allow setting up most of your calculations quite easily from within the ARTS namespace.
- Add all the methods without any defaults and with full control over in-and-out parameters to ARTS::Method taking a true Group type and to ARTS::AgendaMethod taking a Var type. This is easy since Verbosity is never input normally, so the signature would be unique.  (But verbosity is ignored by AgendaMethod, so it is ugly.)
- Add all the methods with partial defaults to ARTS::Method taking Group type and to ARTS::AgendaMethod taking Var type.  This one is more tricky since the signature might get weird overlaps with the GIN/GOUT only functions.  A variadic function would have to be setup and the minimum number of arguments it has to take is to cover all GIN without defaults.  The rest is possible with structs with default arguments but I can imagine it getting complicated.
- Make the signatures work with named inputs.  This is very tricky and I don't think it is possible without a lot of ugliness or meta-programming tricks.  Perhaps the same struct trick can be used again but using the C-default argument style of C99/C++20.  This means the functions have to be called using some very ugly tricks like `VectorNLogSpace(ws, {.out=Var::f_grid(ws), .nelem=2, .start=1e5, .end=1e-2});`...

If any of these things are "easy" and "directly useful now", I would have added them.  Presently, there is no modules so they are not pressing, and the variadic stuff is just too complicated.

PR #241, opened 2020-10-09
Cleaning up CHANGES file for release

olemke:
For the release I'd like to improve our `CHANGES-2.3` file a bit more (it had its last commit in 2018 ;-) ). The `NAME CHANGES` section is for sure incomplete, but we should at least try to advertise new key features.

I've added PyARTS and the new line-by-line stuff to key features and the ARTSCAT-6 to general.

Please have a look at what's there and let me know in a comment to this PR of any things to add or change.

The list of "key features" I'd also like to re-use for announcing the new version on the homepage / mailing list.

### DIFF
--- a/CHANGES-2.3
+++ b/CHANGES-2.3
@@ -25,7 +25,7 @@ General
 - Absorption/RT overhaul
   - Support for new lineshapes
   - Performance improvements
-- New system for defining particle size distributions
+- New and extended system for defining particle size distributions
 - DOIT improvements
   - Optimized pressure grid
   - Convergence acceleration
--- a/CHANGES-2.3
+++ b/CHANGES-2.3
@@ -30,6 +30,8 @@ General
   - Optimized pressure grid
   - Convergence acceleration
   - Optional precalculated first-guess field
+- New sensor setup for passband-type, meteorological millimeter
+  instruments (sensor_responseMetMM)
 
 
 Other Changes
--- a/CHANGES-2.3
+++ b/CHANGES-2.3
@@ -2,9 +2,17 @@ Key features
 ============
 - PyARTS: Python bindings for ARTS
 
+- New implementation of the infrastructure for the line-by-line data.
+  Data that is shared between often hundreds of lines is separated from the
+  pure line-by-line data for more efficient memory and computational resource
+  usage.
+
 
 General
 =======
+- New ARTSCAT-6 catalog format to accomodate the new line-by-line
+  infrastructure.  A description can be found in the ARTS Theory Guide.
+
 - New format of scattering meta data (see online doc of scat_meta_single).
 
 - Cleaner separation of scattering data of the different scattering species:
@@ -55,9 +63,6 @@ General
   polarisation was obtained when using the sensor part to extract the V Stokes
   element.
   
-- New ARTSCAT-6 catalog format.  Easier-to-read and modify line data.
-  A description can be found in the ARTS Theory Guide.
-
 - backend_channel_responseGaussian now takes Vectors instead of Numerics as
   inputs making it possible to define different responses for each channel.
 
--- a/CHANGES-2.3
+++ b/CHANGES-2.3
@@ -1,15 +1,39 @@
 Key features
 ============
-- PyARTS: Python bindings for ARTS
 
-- New implementation of the infrastructure for the line-by-line data.
-  Data that is shared between often hundreds of lines is separated from the
-  pure line-by-line data for more efficient memory and computational resource
-  usage.
+- New improved format for line-by-line data
+- Non-LTE (pure-rotational non-overlapping, and non-chemical cases)
+- Dedicated methods for heating rate calculations
+- Basic simulations of radars (both single and multiple scattering)
+- Radio link calculations not supported in this version
+- Interfaces to DISORT and RT4 scattering solvers
+- Jacobian for new quantities:
+  - spectroscopic variables
+  - particle properties (approximative)
+- OEM type inversions inside ARTS
+- PyARTS: Python bindings for ARTS
 
 
 General
 =======
+
+- Radiative transfer code (except MC) totally revised, including:
+  - Higher consistency between modules
+  - Higher calculation efficiency
+  - Jacobian of atmospheric variables now fully analytical
+- Absorption/RT overhaul
+  - Support for new lineshapes
+  - Performance improvements
+- New system for defining particle size distributions
+- DOIT improvements
+  - Optimized pressure grid
+  - Convergence acceleration
+  - Optional precalculated first-guess field
+
+
+Other Changes
+=============
+
 - New ARTSCAT-6 catalog format to accomodate the new line-by-line
   infrastructure.  A description can be found in the ARTS Theory Guide.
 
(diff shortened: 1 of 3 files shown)


riclarsson:
- A big change not mentioned anywhere is that the Jacobian is now analytical instead of perturbed (even if 2.2 did this in-place).

- A minor change is that the computations of Jacobian and RT are faster because of using Eigen instead of allocating small matrices to heap (using RadiationVector and TransmissionMatrix classes).

- Another minor change is that we use better linear algebra for inversion and exponent of propagation matrices.  Again, this is just minor speed-ups.

- Another minor change is that the entire LBL routine has been changed (not just the catalog).  We can support new line shapes because of this.

- Another minor change is that we have the CPP API.

- Another big change is that we now do NLTE.  Forward calculations should all work but NLTE inverses work with pure-rotational non-overlapping, and non-chemical cases (which, I agree, is limiting).

- Another big change is that we have changed the definition of most RT so that emission/transmission/hybrid types of RT uses the same syntax (i.e., that the emission term never assumes it is only LTE absorption, but that it can be either scattering or NLTE).  I am not sure how different the radar stuff is here though for RT.

Those are the short list that I can think of in quick.  The big ones are considered big because they change the theoretical framework.  The minor ones are considered minor because they mostly change usage.

olemke:
@riclarsson Thanks a lot for your list. I'll incorporate those into the file. The C++ API I had of course on the radar, but since it was just added, I considered holding off until ARTS 3 to really advertise it. But since you seem okay with it, I'll add it, too.

riclarsson:
@olemke Sure, I think we can leave the API out for 2.4.  I think it works as it is and I am starting to use it.  But you are aright, it is very new.

erikssonpatrick:
Oliver, very good that you bring this up. This is of course an important part of the release. Here we have some work to do. The list is far from complete, eg OEM and DOIT/RT4 interfaces are missing. This all I have time to say now. We should join forces to make the file complete. And we should try to separate completely new features
 from under-the-hood changes.

erikssonpatrick:
Hi again. Decided to look at this again because I am quite busy next week. And no time now to make a full suggestion, but here is at least a suggestion for the start (with no guarantee that key features list is complete):

Key features
============

- New improved format for line-by-line data

- Non-LTE (pure-rotational non-overlapping, and non-chemical cases)

- Dedicated methods for heating rate calculations

- Basic simulations of radars (both single and multiple scattering)

- Radio link calculations not supported in this version 

- Interfaces to DISORT and RT4 scattering solvers

- Jacobian involving particle scattering (approximative) 

- OEM type inversions inside ARTS

- PyARTS: Python bindings for ARTS


General
=======

- Radiative transfer code (except MC) totally revised, including:
  - Higher consistency between modules
  - Higher calculation efficiency
  - Jacobian of atmospheric variables more analytical

- Absorption ...

- New system for defining particle size distributions

...

erikssonpatrick:
Some comments.

We have since ARTS 1 said that we have semi-analytical calculations for atmospheric variables, so we have to be careful with what we write.

The question is where we draw the line for General. If we should be as detailed as the comments in the present file, I think we will end up with a very long list. And I wonder if the present comments are the most important ones. And this is especially tricky for scattering data and species as we still have both old and new PSD system in place. It would of course be excellent to have a full list of such changes, but a challenge to really make it complete. So maybe we have to give up on this ...

riclarsson:
It's not "more" analytical, it's analytical.  The Jacobian 'can' now be computed analytically.  There are still some limitations of course, but only that our interp-functions still don't return a way to compute the partial derivatives.  In the past, the Jacobian for VMR were as analytical as they are today (assuming the no line-shape version).  All other partial derivatives were completely perturbed but at a smart code-depth.  I don't think we go too far by saying that the Jacobian is analytical now even with the limitation, but in the past you clearly thought it was too much.

This is for Stokes Dim 1.  I am not sure how good the 2.2 version was at polarized Jacobian, since there was a manual approximation instead of analytical solution to the Matrix-exponent derivative.

olemke:
Here's what it looks like now. I've added DOIT and Absorption/RT overhaul in the General section.

In the announcements I would only post this list, but leave the more detailed changes in the file under a heading "Other changes". It might still be useful for someone.

@erikssonpatrick I changed "more analytical" to "fully analytical". If you think that's too much and we should keep "more" instead, please let me know.

Key features
============

- New improved format for line-by-line data
- Non-LTE (pure-rotational non-overlapping, and non-chemical cases)
- Dedicated methods for heating rate calculations
- Basic simulations of radars (both single and multiple scattering)
- Radio link calculations not supported in this version
- Interfaces to DISORT and RT4 scattering solvers
- Jacobian for new quantities:
  - spectroscopic variables
  - particle properties (approximative)
- OEM type inversions inside ARTS
- PyARTS: Python bindings for ARTS


General
=======

- Radiative transfer code (except MC) totally revised, including:
  - Higher consistency between modules
  - Higher calculation efficiency
  - Jacobian of atmospheric variables now fully analytical
- Absorption/RT overhaul
  - Support for new lineshapes
  - Performance improvements
- New system for defining particle size distributions
- DOIT improvements
  - Optimized pressure grid
  - Convergence acceleration
  - Optional precalculated first-guess field

erikssonpatrick:
@olemke I am fine with "fully". And all look good, but we could change to 

New and extended system for defining particle size distributions

An impressive list but still not sure we are complete here. Can you check when we added FASTEM, TELSEM and TESSEM interfaces? Aren't at least the two last ones after 2.2?

erikssonpatrick:
@riclarsson The interp issue was exactly why I hesitated to write fully analytical, but I have no problem of ignoring that.

olemke:
@erikssonpatrick Got catch with T+T. Added to the list. Fastem has been around for longer though. Also added the "new" MetMM setup to the list.

erikssonpatrick:
Yes, MetMM should also be on the list.

olemke:
This is the final version I would put in the announcements. Any change requests?

Key features
============

- New improved format for line-by-line data
- Non-LTE (pure-rotational non-overlapping, and non-chemical cases)
- Dedicated methods for heating rate calculations
- Basic simulations of radars (both single and multiple scattering)
- Radio link calculations not supported in this version
- Interfaces to DISORT and RT4 scattering solvers
- Jacobian for new quantities:
  - spectroscopic variables
  - particle properties (approximative)
- OEM type inversions inside ARTS
- TELSEM and TESSEM surface models
- PyARTS: Python bindings for ARTS


General
=======

- Radiative transfer code (except MC) totally revised, including:
  - Higher consistency between modules
  - Higher calculation efficiency
  - Jacobian of atmospheric variables now fully analytical
- Absorption/LBL revised
  - Support for new lineshapes
  - Performance improvements
- New and extended system for defining particle size distributions
- DOIT improvements
  - Optimized pressure grid
  - Convergence acceleration
  - Optional precalculated first-guess field
- New sensor setup for passband-type, meteorological millimeter
  instruments (sensor_responseMetMM)


Known Issues
==========

- TestSpectroscopy fails on macOS

riclarsson (approved):
Looks good.

PR #242, opened 2020-10-09
Remove leftover mentions of scat_species_XXX_field

olemke:
Question @m-brath, @erikssonpatrick : I've removed some leftover mentions of scat_species_XXX_field in `methods.cc`, but I'm struggling with the documentation for [cloudboxSetAutomatically](https://atmtools.github.io/arts-docs-master/docserver/methods/cloudboxSetAutomatically.html). It also refers to `scat_species_XXX_field` in several sentences. I'm especially unsure how to correct the second paragraph that ends with:

> For the latter, the WSM needs to be called once per each scat_species_XXX_field, with previously determined cloudbox_limits provided through *cloudbox_limits_old*.

Does "the latter" here refer explicitly to `scat_species_XXX_field` and if so, does that mean that `cloudbox_limits_old` has no use anymore?


erikssonpatrick:
Seems that the WSM is still using the old PSD system, so I have to refer to @m-brath. 

However, I would say that the text here is not critical for v2.4, but the WSM and its text should be updated to v3 as part inside issue #206. 

In line with what I write for CHANGES, the question is if we bother to document all around the old PSD system, as we plan to remove it.

olemke:
Updated docs for `cloudboxSetAutomatically` after discussion with @m-brath

PR #246, opened 2020-10-14
Interp

riclarsson:
This is a very basic and initial merge for a new interpolation routine.  I was reminded because of the 2.4 release about this issue.  I have only use of squashing and linear interpolations, so I don't have good ways to test other types of data.

The code sets up Lagrange polynominals in the most basic sense possible.  The usage mimics in part the old code.  You setup a Lagrange around some new grid-point
```cpp
const Interpolation::Lagrange lag(x, x_grid, order);
```
You compute the interpolation weights (which for 1D is already computed in Lagrange)
```cpp
const Vector iwlag = interpweights(lag);
```
And finally you get the value
```cpp
const Numeric y = interp(y_grid, iwlag, lag)
```
The new thing is that you can compute the interpolation weights derivative (which for 1D is also already computed in Lagrange)
```cpp
const Vector diwlag = dinterpweights(lag);
```
And finally you get the interpolated derivative value
```cpp
const Numeric dy = interp(y_grid, diwlag, lag)
```

I have only tested the 1D cases, so I am sure there is plenty missing for the larger cases.  All the code except the core Lagrange class is auto-generated.

I am committing this because there is no other motion to fix the lack of derivative output from the current interpolation routine. I suspect a lot of this should be turned off before it is fully tested.  I note from my own tests in test_interpolation that this routine differs from interpolation_poly ever so slightly, but I don't know why and how, or if this is expected because interpolation_poly is being smart.


olemke:
Github now offers to mark PRs as a Draft. I thought I try that here. :-)

riclarsson:
So I have now tested the Numeric interp and the full reinterp.  The former is easy to confirm that it works because it is simple.  The second just calls the first a bunch of times for different local interp.  I need to check the second for squashing-like operations.

Now it struck me that these are the only two functions required for a interpolation routine.  I am considering removing the options to directly generate, say, a Tensor4 from a Tensor5.  Instead, this would be a reinterp of a Tensor5 into another Tensor5 and then just a selection of a Tensor4View to return to a Tensor4.  I am very much unsure we can do that without extra copies?  Is it possible to do this in ARTS:
```cpp
// I want this to not copy the main data, because if it does then a whole lot more complicated function signatures have to exist
const Tensor4 a{Tensor5(N, 1, M, K, L, 0.0)(joker, 0, joker, joker, joker)};
```
Do you know, Oliver?  I have been assuming that `Tensor4 a = fun(...)` works, since it is a much simpler case.

riclarsson:
Oh, does GCC8 not support C++17 algorithms?

olemke:
> Oh, does GCC8 not support C++17 algorithms?

std::reduce is part of Parallelism TS which is not fully available in GCC8 according to https://en.cppreference.com/w/cpp/compiler_support#cpp17

erikssonpatrick:
@riclarsson First a general comment. Please check with us others before you start a bigger project like this. This is one reason to have the ARTS video confs. 

The present interpolation routines are by @stefanbuehler. There are also some poly-interpolation functions. @simonpf talked about making some new interpolation routines. And now you have started this. While we rather want to have a single interpolation function set.

My suggestion is to put this on hold until at least the video conf tomorrow. I think Stefan said that he will not join tomorrow, so we can probably not take a decisoon tomorrow. But  we can have a basic discussion.

This is all I have time to read and write today.

olemke:
> ```c++
> const Tensor4 a{Tensor5(N, 1, M, K, L, 0.0)(joker, 0, joker, joker, joker)};
> ```

This invokes one copy. `Tensor5(N, 1, M, K, L, 0.0)(joker, 0, joker, joker, joker)` evaluates to a `Tensor4View` and the constructor call for `Tensor4 a(Tensor4View)` needs to make a copy. The data of a `Tensor4View` cannot be moved  to a `Tensor4` because the View doesn't "own" the data.

riclarsson:
@olemke Does that mean that I need to remove code using the new algorithms to support GCC8 or that GCC8 is to be removed from the build check?

Are you opposed to having a move-and-destroy friend function then?  Like this: `template <Index dim1, Index dim2> Tensor3 reduce_rank(Tensor5&& t5) {...}`, where the dim-values are used to choose where the 0th index are taken from.  The function could throw if the total number of elements is wrong, and it can set mdata to nullptr.  I think delete[] will ignore nulldata, though I am not sure.  (I hope there's no name collisions with the template-form, because they would all need different number of template arguments or different inputs).

@erikssonpatrick It's not much work but no one is doing it.  Lagrange polynominals are repetitive, so a simple string-manipulator can generate all the main code for you.  The main reason to put code behind ideas is that otherwise it doesn't get written, and the idea is presented with the code not without it because ideas without solutions cannot be merged in the future.  The current code should supports all use-cases today.  I want to simplify the interface from what we have today and remove some direct features in favor of indirect solutions though, because I think it simplifies things.

The new interface is:
```cpp
// Create equivalent to GridPos (2 functions only, one is a class):
const auto lag_interp = Interpolation::Lagrange(new_numeric_x, x_grid, order);
const auto lag_reinterp = Interpolation::LagrangeVector(new_x_grid, x_grid, order);

// Create equivalent to interpweights (Vector-like; 4 functions)
const auto iw_interp = interpweights(lag_interp);
const auto iw_reinterp = interpweights(lag_reinterp);
const auto dw_interp = dinterpweights(lag_interp);
const auto diw_reinterp = dinterpweights(lag_reinterp);

// Create equivalent to interpweights (higher order-like; 24 functions)
const auto iw_interp = interpweights(lag_interp1, ..., lag_interpN);
const auto iw_reinterp = interpweights(lag_reinterp1, ..., lag_reinterpN);
const auto diw_interp_d2 = interpweights(lag_interp1, ..., lag_interpN, 2);
const auto diw_reinterp_d2 = interpweights(lag_reinterp1, ..., lag_reinterpN, 2);

// Compute the interpolation (7 functions)
const auto new_numeric_y = interp(y_field, iw_interp, lag_interp1, ..., lag_interpN)
const auto dnew_numeric_y = interp(y_field, diw_interp, lag_interp1, ..., lag_interpN)

// Compute the reinterpolation (7 functions)
const auto new_y_field = reinterp(y_field, iw_reinterp, lag_interp1, ..., lag_interpN)
const auto dnew_y_field = reinterp(y_field, diw_reinterp, lag_interp1, ..., lag_interpN)

// Reduce some fields:
// This is still in the airs, it would be nice to have reduce_rank..., otherwise you must use the MatpackView to access sub-type, which is cumbersome, or make a copy, which is potentially slow
```

olemke:
We still need GCC 8 because that's what's used on our cluster here.

Friend functions should be avoided because they break encapsulation rules exposing internals of the object to outside functions. If we decide to introduce `reduce_rank`, it should be part of the matpack classes themselves. It would still require the Tensor classes to befriend each other but at least it stays in the family. ;-)
Why is the `reduce_rank` function using templates? Could just take the dimension to drop as function arguments?

`delete` can be called with null as documented here: https://en.cppreference.com/w/cpp/language/delete

riclarsson:
Right, so then there's quite a lot of C++17 that is out of reach because of your clusters.  I really like the algorithms of C++17, they make a lot of code easier to write.

I had reduce_rank as a template because it has to be known at compile time what ranks are to be reduced (since it takes ownership).  It also allows you to demand that each dimension is reasonable with static_assert and a nice warning message (even demanding that dim0 is smaller than dim1 and so on).  I don't know how to make reduce_rank the easiest way, but the content of the functions should basically just call `MatpackView::MatpackView(Numeric* data, const Range& Nr, ... const Range& rc)` from within a `Matpack` of a higher rank and then set its own `mdata=nullptr`.  I don't know the best practice for giving other types access to each-other's internals.

riclarsson:
@olemke 

Is this an acceptable way to implement reduce_rank?  Here is an example that goes from Tensor7 ro Tensor3.  I had to implement size() for all matpack types and expose the view-constructor.  The latter I think it would be nice if it was not exposed but just accessed in a friendly manner (though I don't know much about C++ inheritance and friend-requests.)

  ```cpp
  /*! Reduce a Tensor7 to a Tensor3 and leave this in a bad state */
  template <std::size_t dim0, std::size_t dim1, std::size_t dim2>
  Tensor3 reduce_rank() {
    static_assert(dim2 < 7, "Bad Dimension, Out-of-Bounds");
    static_assert(dim0 < dim1, "Bad Dimensions, dim1 must be larger than dim0");
    static_assert(dim1 < dim2, "Bad Dimensions, dim2 must be larger than dim1");
    
    Range r0(0, dim0 == 0 ? nlibraries() : dim0 == 1 ? nvitrines() : dim0 == 2 ? nshelves() : dim0 == 3 ? nbooks() : npages());
    Range r1(0, dim1 == 1 ? nvitrines() : dim1 == 2 ? nshelves() : dim1 == 3 ? nbooks() : dim1 == 4 ? npages() : nrows());
    Range r2(0, dim2 == 2 ? nshelves() : dim2 == 3 ? nbooks() : dim2 == 4 ? npages() : dim2 == 5 ? nrows() : ncols());
    
    Tensor3 out(mdata, r0, r1, r2);
    if (size() not_eq out.size()) throw std::runtime_error("Can only reduce size on same size input");
    mdata = nullptr;
    return out;
  }

```

**Comment by olemke** [comment]
I think it's a viable solution. Having the constructor exposed is not a big deal in my opinion. It might even come in handy if we ever want to create a Tensor from data in memory. Having it hidden would require all higher tier Tensor classes to be friends with the lower tier Tensors. That seems less clean to me than this solution. It should be mentioned in the documentation of the constructor that the Tensor will take care of freeing the memory of the pointer that is passed in. That's an important information for the caller of the constructor.
One other little nitpick is the wording in the documentation for reduce_rank: "and leave this in a bad state" This sounds like it leaves the original Tensor in a broken state which it doesn't (and should be considered a bug otherwise). Why not write something like: "and sets this empty"

**Comment by riclarsson** [comment]
So now this draft can be merged as it adds features to ARTS.  I don't think it should be yet though, unless you accept a new interface for interpolations.

The interface here is slightly different than before because C++11 move semantics.  As written above, the interface is:
```cpp
// Create equivalent to GridPos (4 functions only, two are classes):
const auto lag_interp = Interpolation::Lagrange(new_numeric_x, x_grid, order);
const auto lag_reinterp = Interpolation::LagrangeVector(new_x_grid, x_grid, order, extrapol);
const auto lag_interp = Interpolation::FixedLagrange<order>(new_numeric_x, x_grid);
const auto lag_reinterp = Interpolation::FixedLagrangeVector<order>(new_x_grid, x_grid, extrapol);

// Create equivalent to interpweights (Vector-like; 4 functions)
const auto iw_interp = interpweights(lag_interp);
const auto iw_reinterp = interpweights(lag_reinterp);
const auto dw_interp = dinterpweights(lag_interp);
const auto diw_reinterp = dinterpweights(lag_reinterp);

// Create equivalent to interpweights (higher order-like; 24 functions)
const auto iw_interp = interpweights(lag_interp1, ..., lag_interpN);
const auto iw_reinterp = interpweights(lag_reinterp1, ..., lag_reinterpN);
const auto diw_interp_d2 = interpweights(lag_interp1, ..., lag_interpN, 2);
const auto diw_reinterp_d2 = interpweights(lag_reinterp1, ..., lag_reinterpN, 2);

// Compute the interpolation (7 functions)
const auto new_numeric_y = interp(y_field, iw_interp, lag_interp1, ..., lag_interpN)
const auto dnew_numeric_y = interp(y_field, diw_interp, lag_interp1, ..., lag_interpN)

// Compute the reinterpolation (7 functions)
const auto new_y_field = reinterp(y_field, iw_reinterp, lag_interp1, ..., lag_interpN)
const auto dnew_y_field = reinterp(y_field, diw_reinterp, lag_interp1, ..., lag_interpN)

// Reduce the reinterpolation (part of Matpack)
const Tensor[N-1] new_y_field = reinterp(y_field, iw_reinterp, lag_interp1, ..., lag_interpN)).reduce_rank<i>()
const Tensor[N-2] new_y_field = reinterp(y_field, iw_reinterp, lag_interp1, ..., lag_interpN)).reduce_rank<i, j>()
const Tensor[N-3] new_y_field = reinterp(y_field, iw_reinterp, lag_interp1, ..., lag_interpN)).reduce_rank<i, j, k>()
const Tensor[N-4] new_y_field = reinterp(y_field, iw_reinterp, lag_interp1, ..., lag_interpN)).reduce_rank<i, j, k, l>()
const Tensor[N-5] new_y_field = reinterp(y_field, iw_reinterp, lag_interp1, ..., lag_interpN)).reduce_rank<i, j, k, l, m>()
const Tensor[N-6] new_y_field = reinterp(y_field, iw_reinterp, lag_interp1, ..., lag_interpN)).reduce_rank<i, j, k, l, m, n>()
```

The question I don't know how to answer is if there is any speed degradations in the new code.  This code performs more calculations since the derivatives are computed.  I don't think this will be an issue because allocations rather than computations should be the main speed issue.  There is the same amount of allocation in GridPosPoly and Lagrange, whereas FixedLagrange does not allocate at all.  GridPos occupies 24 bytes whereas FixedLagrange<1> occupies 40 bytes, so the latter might be slower than GridPos.  FixedLagrange<1> can be used entirerly in constexpr sense though, so depending on use case, runtime can be made shorter.

All of this is before derivatives are needed.  When derivatives are needed the new routine should be a lot faster since there are no re-computations.  It should also be more accurate since the calculations are guaranteed to be in-place.

The present code can replace ALL of the current interpolation routines, but making that switch is not something I will do easily, since I am not sure there exists good tests for the full width of the current interpolation routines.  Are there any exhaustive tests available?  I can make the code exchange over time if tests exists but this will take some time since I need to have time.  I don't mind if this pull request just hangs around here until all the remaining interpolation routines are switched.  If I make this change, the functions that I have not been involved with but need the derivatives will still compute the interpolation twice since the results changes otherwise.

olemke:
I'm currently working on an updated implementation of the Hitran crosssection implementation. There, we heavily rely on the interpolation code to map the crosssection data onto the current `f_grid`. This is taking a considerable amount of time, so I took this as a test to see how the new routines compare to the old ones. Here is the code of the old and the new interpolation I'm using:

```cpp
if (0) {
        const auto lag_interp = Interpolation::LagrangeVector(
            f_grid_active, data_f_grid_active, f_order, 0.5);
        const auto itw = interpweights(lag_interp);
        xsec_interp = reinterp(fit_result, itw, lag_interp);
      } else {
        // Find frequency grid positions:
        ArrayOfGridPosPoly f_gp(f_grid_active.nelem()), T_gp(1);
        gridpos_poly(f_gp, data_f_grid_active, f_grid_active, f_order);

        Matrix itw(f_gp.nelem(), f_order + 1);
        interpweights(itw, f_gp);
        interp(xsec_interp, itw, fit_result, f_gp);
      }
```
In my current test, `f_grid_active` has ~220000 points, `data_f_grid_active` has ~23000 points. The interpolation is executed several times during `yCalc`. Here's the execution time measured for yCalc:
```
...
- yCalc
- timerStop
- Print
  * CPU time  total: 13.09  user: 11.07  system: 2.02
               real: 13.10  99.92%CPU
...
```
Approximately 4 seconds out of the total time can be attributed to the interpolation. And here the execution time with the new interp routines:
```
...
- yCalc
- timerStop
- Print
  * CPU time  total: 197.37  user: 196.58  system: 0.79
               real: 197.48  99.94%CPU
...
```
This shows a huge degradation in performance.

For further testing, I've added a simplified test case (test15 in test_interpolation.cc) in this branch. It directly compares the old vs. the new interpolation on the same data. Timings here also don't look good unfortunately:
```
 ➜ src/test_interpolation
========== New interp ==========
  * CPU time  total: 1.85  user: 1.85  system: 0.00
               real: 1.86  99.46%CPU

========== Old interp ==========
  * CPU time  total: 0.02  user: 0.02  system: 0.00
               real: 0.02  100.00%CPU

Max difference new-old interpolation: 1.42109e-13
```

Before digging any deeper into this issue, my first question would be whether I'm doing something completely wrong in using the new routines?

riclarsson:
Hi Oliver,

Ow, that's a huge degradation.  I haven't spent much time thinking about the vector case.  Just looking at the code again, I see that there are probably plenty of improvements to be made...

One that mostly fixes the issue you have above was to do this in interpolation_lagrange.cc:
```c++
std::vector<Lagrange> LagrangeVector(const ConstVectorView xs,
                                     const ConstVectorView xi,
                                     const Index polyorder,
                                     const Numeric extrapol) {
  std::vector<Lagrange> out;
  out.reserve(xs.nelem());
  bool has_one=false;
  for (auto x : xs) {
    if (has_one) {
      out.emplace_back(out.back()).update(x, xi, extrapol);
    } else {
      out.push_back(Lagrange(x, xi, polyorder, extrapol));
      has_one = true;
    }
  }
  return out;
}
```

It has the unfortunate side-effect that it demands that xs is sorted.  From that change alone, I get most of the differences to go away.  Is it part of the current interp to assume that the new grid is also sorted?  Otherwise I need to rethink a bit here...

I will have a look at Grid later.  It might be bad presently, and could probably be improved significantly.

olemke:
The new grid does not have to be sorted in the current interpolation routines. But a sorted grid gives better performance.

riclarsson:
The new code has the following effect now:

The original test_interpolation.cc, the new code is still a lot slower
```
========== New interp ==========
  * CPU time  total: 1.34  user: 0.63  system: 0.71
               real: 0.08  1675.00%CPU

========== Old interp ==========
  * CPU time  total: 0.01  user: 0.01  system: 0.00
               real: 0.02  50.00%CPU

Max difference new-old interpolation: 1.42109e-13
```

Just switching order, it looks like the new code is still slower but with somewhat less of an impact.
```
========== Old interp ==========
  * CPU time  total: 1.28  user: 0.57  system: 0.71
               real: 0.07  1828.57%CPU

========== New interp ==========
  * CPU time  total: 0.09  user: 0.06  system: 0.03
               real: 0.03  300.00%CPU

Max difference new-old interpolation: 1.42109e-13
```

OK, there are some more things that needs doing on the code to support the old routine in full.  I need to add support for negatively sorted and unsorted new grids.  I also need to add support for negatively sorted original grids.  Can the original grid be unsorted?

Before that, I will have a look to try and fix grid not making so many copies so that your new test case shows comparable performance.

olemke:
The current interpolation routines require a sorted grid.

riclarsson:
Updated test results.  I am now using the old and new routine optimally as I can see it.

These are the times:
```
========== New interp ==========
Median time: 0.0182007 seconds
========== Old interp ==========
Median time: 0.00747708 seconds
```

When the derivatives are commented out from the new routine (not removing the memory allocations) it reads:
```
========== New interp ==========
Median time: 0.0149465 seconds
========== Old interp ==========
Median time: 0.00740572 seconds
```

(If the median loop calculator also allocates the matpack-types, the results are 0.0129308 seconds for the old routine.)

So the new routine is 40% slower than a dry run of the old routine and 140% slower than a rerun of the old routine.  I should probably introduce a flag for not doing the derivatives, in which case the same respective numbers are 15% for dry run and 100% for rerun.

riclarsson:
Now I have to leave this for the coming days.  I don't know how to improve this anymore by just looking at memory.  I suspect there is a much smarter way to compute the start position than to just loop over everything as I am doing presently.  Anyways, the numbers now are for dry runs:

```
========== New interp ==========
Fastest time: 0.0128591 seconds
Median  time: 0.0131081 seconds
Max-5%  time: 0.0136703 seconds
========== Old interp ==========
Fastest time: 0.0107804 seconds
Median  time: 0.0110069 seconds
Max-5%  time: 0.0116489 seconds
```

And for dry runs without derivatives:
```
========== New interp ==========
Fastest time: 0.0122821 seconds
Median  time: 0.0125142 seconds
Max-5%  time: 0.0131663 seconds
========== Old interp ==========
Fastest time: 0.010747 seconds
Median  time: 0.0109394 seconds
Max-5%  time: 0.0115633 seconds
```

So the short of it is that if you need derivatives then the new code will be 40% faster but if you don't need them it is going to be 20% slower.  This could be improved by adding complications to just a 15% degradation.

stefanbuehler:
Hi Richard,

this is very interesting stuff! My original version was not strictly optimised for performance. For example, I made the design choice to allocate the polynomial grid positions and interpolations weights objects at runtime, since they have different size depending on the interpolation order. That is perhaps the most serious performance limitation for the higher order interpolation, since the allocation is slow. 

I also did not check whether there are numerically more efficient alternatives to my quite naive implementation. I see on the German Wikipedia page (https://de.wikipedia.org/wiki/Polynominterpolation) that there are quite a few numerical schemes for polynomial interpolation around.  From a quick read, I think the one that I have implemented is 
the barycentric interpolation formula. They say there that others can be more efficient, mentioning the Newton and Neville-Aitken algorithms. (Strangely, the English  version of the Wikipedia article is quite different and does not mention these, but perhaps other interesting aspects).

I would therefore expect a re-implementation to be more efficient than my original version.

Aside from the performance issues, how do you plan to go about to eliminate the existing interp and interp_poly systems? Can there be new frontend functions that are a drop-in replacement for the existing interface? - This seems to be almost a necessity, because otherwise you would have to do a lot of reprogramming across large parts of ARTS.

Concerning testing, I am not so worried. If the existing tests run through with a new ARTS version with replaced interpolation core, I would consider that as a strong proof that they are ok. What I mean is that any critical issues would very likely show up on the existing tests, I think, either as wrong result values or as degraded computation time.

stefanbuehler:
> Now I have to leave this for the coming days. I don't know how to improve this anymore by just looking at memory. I suspect there is a much smarter way to compute the start position than to just loop over everything as I am doing presently. 

My original gridpos does two things, which are both simple:
- To find the first start position, a first guess is computed by assuming that the original grid is evenly spaced. The code line doing this is
   Index current_position = (Index)rint(frac * (Numeric)(n_old - 2));

- From then on, indices are just counted up or down, as needed. For each new element of the new grid, we start with the index of the previous element. This is efficient if the new gridpoints are sorted, and particularly efficient if they are so close together that several of them are inside the same interval of the original grid, since then the index does not have to be changed at all.

One more thing about gridpos: Note that it probably is used in some places (ppath routines?) for things other than just interpolation.

And another thing: I always thought the gridpos part is pretty generic, and that any conceivable interpolation function would somehow need this step. So, I have not yet understood why you want to replace it.

Unfortunately, I cannot join the meeting this Friday, but perhaps we could have an in-depth discussion of interpolation at the next one on November 13?

erikssonpatrick:
Yes, the gridpos-es are used quite heavily in the Jacobian calculations (as mentioned at last video meeting).  So they have to stay (and be consistent with actual interpolation). Or a lot of complicated recoding to be done.

riclarsson:
@stefanbuehler I don't want to give the impression that the current code is optimized.  It is not.  It is simply the Lagrange Polynom Interpolation formula+derivatives taken directly from Wikipedia and MathWorld in a way that allows both fixed and runtime definitions.  This ends up with code that except for the startup cost of your rint should be about the same speed as GridPosPoly.

My vector-case is however just a patchwork on top, and does not remember at all the previous state.  This makes your GridPosPoly much more efficient, and this is probably where the main differences between the codes come from.  I should probably have dedicated classes rather than independent entries in a vector here.

In the test case that Oliver submitted, for the FixedLagrange (which does not allocate) the code is already some 70% faster than GridPosPoly (once the same minor hiccups as Lagrange had were fixed) but they are not comparable since in one case you have to select ahead of time your interpolation order.  This is without removing derivatives.

I think that to eliminate current interp-routines, you are right that there must probably be an interface that is the same as today.  I am not entirely sure what that interface is though.  What is the width of the interface?

Anyways, the latest numbers now that I simplified the interface is with derivative:
```
========== Fix interp ==========
Fastest time: 0.00307922 seconds
Median  time: 0.00310053 seconds
Max-5%  time: 0.0036057 seconds
========== New interp ==========
Fastest time: 0.0127339 seconds
Median  time: 0.0128778 seconds
Max-5%  time: 0.0136001 seconds
========== Old interp ==========
Fastest time: 0.0107313 seconds
Median  time: 0.0108736 seconds
Max-5%  time: 0.0114658 seconds
```

Without derivative (for new):
```
========== Fix interp ==========
Fastest time: 0.00307207 seconds
Median  time: 0.00310176 seconds
Max-5%  time: 0.00367645 seconds
========== New interp ==========
Fastest time: 0.0115446 seconds
Median  time: 0.011786 seconds
Max-5%  time: 0.0123211 seconds
========== Old interp ==========
Fastest time: 0.0108058 seconds
Median  time: 0.0109223 seconds
Max-5%  time: 0.0115055 seconds
```

stefanbuehler:
@riclarsson 

Richard, the interfaced is described in quite some detail in the developer guide. Have you checked that? If not, please do, and then let's discuss the remaining questions. :-)

riclarsson:
Right, I have looked at the code and the documentation but still cannot make heads and tails out of it.

In short, is the current interface to interpolate from all higher orders to all lower orders?

Can Tensor7 as an original grid can be regridded as Tensor7, Tensor6, Tensor5, Tensor4, Tenso3, Matrix, Vector and Numeric?
Can Tensor6 as an original grid can be regridded as Tensor6, Tensor5, Tensor4, Tenso3, Matrix, Vector and Numeric?
And so on?  Or is it the same as the simpler version I made where you can only interpolate to Numeric and regrid as the same type?  It looks like it is something half-way between this.

The code I have can only interpolate to Numeric and reinterpret to its own size.  I can probably add methods that reduces these to the lower orders if needs be.  Something like
```c++
template <std::size_t... dims>
auto interp(...) {
  return reinterp(iy, iw, ...).reduce_rank<dims...>();
}
```

For each matpack class should be enough together with the already existing interp(...) function to achieve interpolation to any lower rank, perhaps with a special.

The additions will be that I need to add methods that takes the views rather than returns the value.  This might make the code a lot slower.  Which would be a shame because I think I just found one of the last major slowdowns here so that the times (now on laptop) are now looking OK:

```
========== Fix interp ==========
Fastest time: 0.00288559 seconds
Median  time: 0.00301507 seconds
Max-5%  time: 0.00879443 seconds
========== New interp ==========
Fastest time: 0.012182 seconds
Median  time: 0.012334 seconds
Max-5%  time: 0.0128254 seconds
========== Old interp ==========
Fastest time: 0.012926 seconds
Median  time: 0.0132332 seconds
Max-5%  time: 0.0158345 seconds
```

stefanbuehler:
Hi Richard,

the gridpos step is always the same, since it acts independently on each 
dimension.

For the other steps, there are three different cases, called after 
colours, just to give them a name:

Green: Dimension stays the same (Tensor7 is regridded to Tensor7)
Blue:  New dimension is one (Tensor7 is regridded to Vector)
Red (not mentioned in uguide): The special case that the Vector has 
length 1, so we are extracting only a single point. This case actually 
performs more poorly than the others, since there is an extra copy 
involved.

In practise, the green case is hardly used. The blue case is used a lot, 
and also the red case, somewhat unfortunately, due to its poor 
implementation (it is just a wrapper around the blue case functions).

> Or is it the same as the simpler version I made where you can only 
> interpolate to Numeric and regrid as the same type?  It looks like it 
> is something half-way between this.

Well, the only extra feature is that you can have not just a single 
Numeric, but a whole stack of them at the same time (the blue case). If 
you need more than a single interpolated point, this is considerably 
more efficient for gridpos than looping over a scalar function. (Because 
you can start the search from the last point, which is efficient if the 
points are sorted.)

However, I just realise that for the two remaining steps (determining 
the weights and applying them) there probably is no big difference to 
both approaches. One could explicitly loop over the gridpos objects, and 
do the interpolation for each individually. Perhaps it would be a bit 
slower, because more function calls, but probably not much.

So, I could have made an alternative design, where interpweights and 
interp are scalar functions, and the user has to do the looping over the 
different interpolation points himself. Perhaps that would have been 
better, since quite often the standard cases do not fit.

If one were to introduce such a design now, one would have to keep the 
existing vector interpweights and interp functions as wrappers. They 
would then just contain appropriate loops over the scalar functions, so 
would be much easier to understand. In this way the existing code in 
user space would not have to be changed.

> The code I have can only interpolate to Numeric and reinterpret to its 
> own size.

I think that’s ok. But I don’t yet understand how your code maps 
onto the three elementary steps of interpolation (1. find grid 
positions, 2. calculate weights, 3. apply weights). Does it? Or is it 
applying several steps in one go? That would be problematic, both for 
efficiency reasons and for the sake of not having to reprogram all the 
user space code.

> The additions will be that I need to add methods that takes the views 
> rather than returns the value.

In the light of what I wrote above. Perhaps a core method that just 
returns the scalar value is the cleanest and most efficient solution. 
Then wrappers around this for the different existing cases (that part is 
easy, I think).

I would recommend that you use the existing GridPos and GridPosPoly 
classes, to be compatible with all the other non-interpolation methods 
that use them.

All the best

Stefan

erikssonpatrick:
I really think it is time to take one step back here and be clear about what the aim is? And if there any constraints (such as degradation of calculation efficiency is not accepted). Otherwise the risk is high that a time is wasted. And when that is clear, discuss how to most easily solve it.

erikssonpatrick:
That said, some comments to potentially save time.

I have the impression that Richard you started this to get derivatives. I don't know exactly what derivatives you need/want. But I find it likely that the present scheme can be extended to give the derivatives you lack. The output of the interpolation is a weighted avergare of original data and the derivative of this expression can be derived automatically. (did I mention this on the video meeting or it just passed my mind?)

If someone happen to not understand some code, that is not an imediate reason to rewrite. It could just reverse the problem. You will understand this part of ARTS better, while others will likely understand it less. When it comes to the interpolation routines, this is probably the best documented part of ARTS' internal functions, so to improve understanding is a big challange for this part. In a situation like this, a better solution is likely to join forces to solve the problem with least overall effort. And that is one of the reasons to why we have the ARTS video meetings.

riclarsson:
@stefanbuehler Hi Stefan,

> Hi Richard, the gridpos step is always the same, since it acts independently on each dimension. For the other steps, there are three different cases, called after colours, just to give them a name: Green: Dimension stays the same (Tensor7 is regridded to Tensor7) Blue: New dimension is one (Tensor7 is regridded to Vector) Red (not mentioned in uguide): The special case that the Vector has length 1, so we are extracting only a single point. This case actually performs more poorly than the others, since there is an extra copy involved. In practise, the green case is hardly used. The blue case is used a lot, and also the red case, somewhat unfortunately, due to its poor implementation (it is just a wrapper around the blue case functions).

The new routine can now mimic the VectorView reduction case you call "Blue".  The "Red" and the "Green" case are the base cases in this new setup.

> Or is it the same as the simpler version I made where you can only interpolate to Numeric and regrid as the same type? It looks like it is something half-way between this.
> Well, the only extra feature is that you can have not just a single Numeric, but a whole stack of them at the same time (the blue case). If you need more than a single interpolated point, this is considerably more efficient for gridpos than looping over a scalar function. (Because you can start the search from the last point, which is efficient if the points are sorted.) However, I just realise that for the two remaining steps (determining the weights and applying them) there probably is no big difference to both approaches. One could explicitly loop over the gridpos objects, and do the interpolation for each individually. Perhaps it would be a bit slower, because more function calls, but probably not much. So, I could have made an alternative design, where interpweights and interp are scalar functions, and the user has to do the looping over the different interpolation points himself. Perhaps that would have been better, since quite often the standard cases do not fit. If one were to introduce such a design now, one would have to keep the existing vector interpweights and interp functions as wrappers. They would then just contain appropriate loops over the scalar functions, so would be much easier to understand. In this way the existing code in user space would not have to be changed.

I don't understand what you mean by "the search"?  Is this related to the dimensions of your interpweights?

> The code I have can only interpolate to Numeric and reinterpret to its own size.
> I think that’s ok. But I don’t yet understand how your code maps onto the three elementary steps of interpolation (1. find grid positions, 2. calculate weights, 3. apply weights). Does it? Or is it applying several steps in one go? That would be problematic, both for efficiency reasons and for the sake of not having to reprogram all the user space code.

It follows those steps exactly.  Lagrange/FixedLagrange (runtime/compile time) are the GridPos.  My hope was that FixedLagrange<1> should be binary compatible for the first 24 bytes as GridPos, but I am not sure if this is the case still. And I need to fix it for negative grids.  Both versions can output a vector version for more grid points and the standard way of doing this will be somewhat optimized for positions.

These Lagrange/FixedLagrange and their vectors are used in "interpweights" and "dinterpweights" functions to generate appropriate grids of interpolation or their derivatives.  These two output a matpack-like type of the same size as the original data in the runtime version for the single interpolation case.  A Tensor3-like interpolation weight of sizes "[3, 2, 1]" means that the page-dimension is square-interpolated, the row-dimension is linearly interpolated and the column interpolation is nearest neighbor*.  To deal with higher dimensions there is a Grid class that can hold any data up to any dimension and have it accessed by "base& operator()(dims...)".

Finally "interp" can be called to get a Numeric, and "reinterp" can be called to get the same type as the original at a new grid.  To support the blue case, I just added "reinterp_reduce", which reduces the dimensions to a VectorView to be easier to deal with the "Blue" case.  The intended original way to do this was otherwise to directly use "reduce_rank" because then you can get a Matrix from a Tensor4 and really any combination you might want.

* With caveat that I need to test this a bit more.

> The additions will be that I need to add methods that takes the views rather than returns the value.
> In the light of what I wrote above. Perhaps a core method that just returns the scalar value is the cleanest and most efficient solution. Then wrappers around this for the different existing cases (that part is easy, I think). I would recommend that you use the existing GridPos and GridPosPoly classes, to be compatible with all the other non-interpolation methods that use them.

Using GridPosPoly would be difficult.  I am actually not sure at all what it holds and how.  I see no need for holding multiple indices.  GridPos should be the same as FixedLagrange<1>, but the latter has the derivatives so it is 16 bytes larger.

riclarsson:
@erikssonpatrick I agree that the original code might be possible to adapt.  Please do it.  Nothing has happened with these derivatives for many years so there is evidently no interest or no time in changing the old routines to do this.

riclarsson:
@erikssonpatrick 
There is no degradation anymore.  I repost the numbers:

```
========== Fix interp ==========
Fastest time: 0.00303733 seconds
Median  time: 0.00327445 seconds
Max-5%  time: 0.00796405 seconds
========== New interp ==========
Fastest time: 0.0117237 seconds
Median  time: 0.0119856 seconds
Max-5%  time: 0.012658 seconds
========== Old interp ==========
Fastest time: 0.0126558 seconds
Median  time: 0.0127205 seconds
Max-5%  time: 0.013373 seconds
```

The new code is the same speed as the old code within reasonable noise.  Unless you know your interpolation order, in which case the new code is significantly faster.

erikssonpatrick:
Richard, please don't take this the wrong way. Honestly, I am trying to help. 

As reply to your last comment, I just wanted to be clear about what I meant with constrain. I used calculation efficiency as example. I did not say that it was an issue now. If the interpolation can be made faster, I will be very happy. I said already at the video meeting.

riclarsson:
So now negative sorted grid works and I added a test to check this.

Output of test:
```
========== Fix interp ==========
Fastest time: 0.00269166 seconds
Median  time: 0.00270482 seconds
Max-5%  time: 0.00323985 seconds
========== New interp ==========
Fastest time: 0.00977629 seconds
Median  time: 0.00989372 seconds
Max-5%  time: 0.0102461 seconds
========== Old interp ==========
Fastest time: 0.0108235 seconds
Median  time: 0.0109795 seconds
Max-5%  time: 0.0114546 seconds

2.25: 0 0.75 0.25  -1   1 - 1 0.25 0.75   1  -1  xstart [asc des]: 2 3
3.25: 1 0.75 0.25  -1   1 - 0 0.25 0.75   1  -1  xstart [asc des]: 3 4
2.35: 0 0.65 0.35  -1   1 - 1 0.35 0.65   1  -1  xstart [asc des]: 2 3

2.25: 0 0.75 0.25 -1 1 - 1 0.25 0.75 1 -1  xstart [asc des]: 2 3
3.25: 1 0.75 0.25 -1 1 - 0 0.25 0.75 1 -1  xstart [asc des]: 3 4
2.35: 0 0.65 0.35 -1 1 - 1 0.35 0.65 1 -1  xstart [asc des]: 2 3
```

There is an issue here.  The FixedLagrange<1> has reversed order of the grid-weights compared to GridPos.  I have looked at these indices, and it seems like GridPos can be used for 0-long grids?  And that the exact definition of these variables matters there?  It is not a lot of work to just switch index otherwise but I need to have a look at this.

stefanbuehler:
> I don't understand what you mean by "the search"?  Is this related to 
> the dimensions of your interpweights?

I meant the search for the right position in the original grid, by 
gridpos or gridpos_poly.

>> The code I have can only interpolate to Numeric and reinterpret to 
>> its own size.
>> I think that’s ok. But I don’t yet understand how your code maps 
>> onto the three elementary steps of interpolation (1. find grid 
>> positions, 2. calculate weights, 3. apply weights). Does it? Or is it 
>> applying several steps in one go? That would be problematic, both for 
>> efficiency reasons and for the sake of not having to reprogram all 
>> the user space code.
>
> It follows those steps exactly.  Lagrange/FixedLagrange 
> (runtime/compile time) are the GridPos.  My hope was that 
> FixedLagrange<1> should be binary compatible for the first 24 bytes as 
> GridPos, but I am not sure if this is the case still. And I need to 
> fix it for negative grids.  Both versions can output a vector version 
> for more grid points and the standard way of doing this will be 
> somewhat optimized for positions.
>
> These Lagrange/FixedLagrange and their vectors are used in 
> "interpweights" and "dinterpweights" functions to generate appropriate 
> grids of interpolation or their derivatives.  These two output a 
> matpack-like type of the same size as the original data in the runtime 
> version for the single interpolation case.  A Tensor3-like 
> interpolation weight of sizes "[3, 2, 1]" means that the 
> page-dimension is square-interpolated, the row-dimension is linearly 
> interpolated and the column interpolation is nearest neighbor*.  To 
> deal with higher dimensions there is a Grid class that can hold any 
> data up to any dimension and have it accessed by "base& 
> operator()(dims...)".
>
> Finally "interp" can be called to get a Numeric, and "reinterp" can be 
> called to get the same type as the original at a new grid.  To support 
> the blue case, I just added "reinterp_reduce", which reduces the 
> dimensions to a VectorView to be easier to deal with the "Blue" case.  
> The intended original way to do this was otherwise to directly use 
> "reduce_rank" because then you can get a Matrix from a Tensor4 and 
> really any combination you might want.

Ok, can you remind me what is the advantage of the reimplementation?

> Using GridPosPoly would be difficult.  I am actually not sure at all 
> what it holds and how.  I see no need for holding multiple indices.  
> GridPos should be the same as FixedLagrange<1>, but the latter has the 
> derivatives so it is 16 bytes larger.

GridPosPoly is very simple. It is explained in Section 5.10 of the 
developer guide. Which part of that explanation is unclear? Perhaps the 
use of the term weights?

What GridPosPoly stores are the weights for the 1D case. For linear 
interpolation, these are just one minus the fractional distances that 
would be stored in the ordinary GridPos object. In hindsight, it would 
have been better to store weights in the ordinary GridPos as well, but 
it was implemented much earlier.

The weight concept is clearer than the fractional distance concept, 
because one weight is associated with every gridpoint in the original 
grid that is used for interpolation. (Whereas fractional distances get a 
bit wonky for the extra grid points that are used for higher order 
interpolation.)

What does your FixedLagrange<1> store?

Best wishes,

Stefan

stefanbuehler:
Hej again,

> There is an issue here.  The FixedLagrange<1> has reversed order of 
> the grid-weights compared to GridPos.  I have looked at these indices, 
> and it seems like GridPos can be used for 0-long grids?  And that the 
> exact definition of these variables matters there?  It is not a lot of 
> work to just switch index otherwise but I need to have a look at this.

You seem to move along a similar path as I did back then. Yes, for the 
linear case GridPosPoly contains the same values as GridPos, just their 
places are switched (see the other comment I gave just before). Because 
for Poly they represent interpolation point weights, whereas for GridPos 
they represent the distance from that point. The smaller the distance, 
the larger the weight, the former is 1 - the latter.

/Stefan

riclarsson:
Hi Stefan,

> Ok, can you remind me what is the advantage of the reimplementation?

That the derivatives are computed inline when needs be.  So if you use say the lookup-table for temperature Jacobian the calculations should be faster and more accurate than today.  If you know the size of your polynominal (even above 1) the calculations will be even faster, for some 300% speedups for pure interpolation of higher orders.  [I have not checked if the current GridPos is faster than FixedLagrange<1> yet but I don't think there should be much difference.]

> GridPosPoly is very simple. It is explained in Section 5.10 of the developer guide. Which part of that explanation is unclear? Perhaps the use of the term weights? What GridPosPoly stores are the weights for the 1D case. For linear interpolation, these are just one minus the fractional distances that would be stored in the ordinary GridPos object. In hindsight, it would have been better to store weights in the ordinary GridPos as well, but it was implemented much earlier. The weight concept is clearer than the fractional distance concept, because one weight is associated with every gridpoint in the original grid that is used for interpolation. (Whereas fractional distances get a bit wonky for the extra grid points that are used for higher order interpolation.) 

I did not understand why you needed a vector of indices.  I see now that the multiple indices are used in cyclic interpolation, I missed that before.  I don't have that as a case yet so this needs fixing (I think it is relatively easy to fix, but the cyclic case might be a bit slower to compute in the solution I imagine, so I need to check that the expected slowdown is at an acceptable level, e.g. by comparing to GridPosPoly cyclic interpolation).

> What does your FixedLagrange<1> store?

FixedLagrange<1> stores the zeroth position in the original grid, the l_0 Lagrange term, the l_1 Lagrange term, the dl_0 Lagrange term derivtive [which is -1 or 1 depending on grid order], and the dl_1 Lagrange term derivative [opposite sign].  In that order.  Your GridPos stores the zeroth position, the l_1 Lagrange term, and the l_0 Lagrange term.  In that order.

For higher poly-order the FixedLagrange<n-1> stores the l_2 to l_n terms before dl_0, and after dl_1 are the dl_2 to dl_n terms.  Of course in these cases the derivatives will be fractional to some degree.

The Lagrange stores the zeroth position in the original grid first, then it stores a Vector to the l:s and then a Vector the the dl:s.  Your GridPosPoly stores a vector of indices and then a Vector to what I think is identical to the l:s that I have if I understand the code properly (which I probably don't).  These storage orders are independent of poly-order since Vector and vector don't actually have any data next to them.

riclarsson:
New speedup, there is no good reason to use matpack types to store the weights, so now the Lagrange is some 32 bytes smaller and is finally smaller than GridPosPoly (8 + 24 + 24 for Lagrange, 24 + 40 for GridPosPoly in bytes).  Before Lagrange was 88 bytes.  The new code is probably a little bit faster than the old code now:

```
========== Fix interp ==========
Fastest time: 0.00269933 seconds
Median  time: 0.00271295 seconds
Max-5%  time: 0.00309088 seconds
========== New interp ==========
Fastest time: 0.00886349 seconds
Median  time: 0.00892188 seconds
Max-5%  time: 0.0093345 seconds
========== Old interp ==========
Fastest time: 0.0106754 seconds
Median  time: 0.0107874 seconds
Max-5%  time: 0.0112582 seconds
```

stefanbuehler:
Hej Rickard,

> FixedLagrange<1> stores the zeroth position in the original grid, the 
> l_0 Lagrange term, the l_1 Lagrange term, the dl_0 Lagrange term 
> derivtive [which is -1 or 1 depending on grid order], and the dl_1 
> Lagrange term derivative [opposite sign].  In that order.  Your 
> GridPos stores the zeroth position, the l_1 Lagrange term, and the l_0 
> Lagrange term.  In that order.
>
> For higher poly-order the FixedLagrange<n-1> stores the l_2 to l_n 
> terms before dl_0, and after dl_1 are the dl_2 to dl_n terms.  Of 
> course in these cases the derivatives will be fractional to some 
> degree.
>
> The Lagrange stores the zeroth position in the original grid first, 
> then it stores a Vector to the l:s and then a Vector the the dl:s.  
> Your GridPosPoly stores a vector of indices and then a Vector to what 
> I think is identical to the l:s that I have if I understand the code 
> properly (which I probably don't).  These storage orders are 
> independent of poly-order since Vector and vector don't actually have 
> any data next to them.

Hm. I don’t fully understand but get the impression, that the data 
stored are quite similar, except for the additional derivatives.

Why not simply extend the existing GridPosPoly with the additional data?

/Stefan

riclarsson:
Well, I have been hoping someone can deal with this for many years now.  Since it never happens I assume that there is no interest.  If someone was willing to extend the old code, I would be happy.

It is clear that all 6 ways to create GridPosPoly need to take a "do_derivatives" argument and then do the appropriate calculations.  This means navigating the calculations and know exactly when and where to do the updates.  Then, each of the interpweights functions will need to have a derivative version.  Since these are macro-heavy, they will have to be re-written from scratch.

For me, there is a lot more work to rewrite the old stuff than to just replace the interface with something I can work with.  I would be happy if someone extended the old interface but I don't see it happening.

stefanbuehler:
Dear Richard,

I don’t see this as negatively as you express, I think you are almost 
there (to understanding the existing interface).

Ah, I think we have reached a point where it is better to discuss at a 
real (Zoom) meeting. The one on November 13?

As you probably gathered, I’m not unsympathetic to changing the 
implementation. (If the new one is cleaner, faster, and equally well 
documented.) Somewhat less sympathetic to changing the outer interface.

The overwhelming majority of interpolations in ARTS are linear, so that 
is the biggest issue. For the higher order cases alone, adapting to a 
new interface is more feasible. For the linear case, I think it is an 
almost impossible task. I’m still not sure what exactly your plans 
are, but let’s discuss in person.

I’m in a very overloaded work situation now, with the teaching period 
starting and teaching three different courses, plus some introduction 
work, plus some other crazy stuff. So will have to tune out of ARTS 
issues until then.

All the best,

Stefan

riclarsson:
> Dear Richard, I don’t see this as negatively as you express, I think you are almost there (to understanding the existing interface). 

Right.  I might be able to understand the current interface but changing the code to do the derivatives is another challenge entirely.

> Ah, I think we have reached a point where it is better to discuss at a real (Zoom) meeting. The one on November 13?

I agree.  I had some time yesterday so I added the cyclic interpolation that was missing for a 1-to-1 replacement.  The speeds are the same.  But of course, it is not as well tested beyond 1D.  There is now Linear, Log, and Cyclic interpolations possible in the new interface.

> As you probably gathered, I’m not unsympathetic to changing the implementation. (If the new one is cleaner, faster, and equally well documented.) Somewhat less sympathetic to changing the outer interface.

The interface has been left the same bar the types.  I see that this can be an issue in ppath since the interpweights MatpackType is created or used manually there.  To switch such code might be difficult since if is not well-documented what the goal is.  Adding simple flatten or reshaped outputs is not that big of a deal though.

> The overwhelming majority of interpolations in ARTS are linear, so that is the biggest issue. For the higher order cases alone, adapting to a new interface is more feasible.  For the linear case, I think it is an almost impossible task. I’m still not sure what exactly your plans are, but let’s discuss in person. I’m in a very overloaded work situation now, with the teaching period starting and teaching three different courses, plus some introduction work, plus some other crazy stuff. So will have to tune out of ARTS issues until then.

Right, I take this to mean that GridPos must remain because it is used manually to mean other things than interpolation positions.  But I agree, the 13th should be OK for me.  I have no Internet at home because that is apparently the system here, so if they decide on harsher lock-downs next Wednesday, I might miss the meeting.

riclarsson:
So I would really like this to be merged at some point since it is touching so many files that maintaining it out of branch will be a mess.  What is your ETA on looking over the code and merging it, @stefanbuehler @olemke  ?

olemke:
I'll have a final look over the code until Friday. @stefanbuehler is reviewing the documentation.

riclarsson:
Thanks!

I just added some last code to the ENUMCLASS because I just found out that there is a constexpr string-type in C++17.  If it does not compile on GCC8, it can be removed.

stefanbuehler (commented):
Looks good to me overall!

Minor questions:

For developer guide, could you also explain what reinterp does, please?

For guide and code: I find the naming of the option "Linear" for LagrangeType rather confusing, since the term linear often is used for interpolation order 1. How about calling it "Direct" instead?

A programming question: You now consistently use Vector and Matrix return values. I assume this is efficient because of the new constructors you added. Correct? This seems quite a paradigm change. It would be good to have a discussion whether we want to generally encourage that also in other areas. Or not. The pros and cons are not clear to me at this point.

And finally a plea: Might I ask you to look for the functions "polint" and "interp_poly", both defined in the old interpolation.cc, and eliminate them? I mean, delete them and replace their calls (in doit.cc and m_doit.cc) by the new Lagrange routines.

Those functions were added ad hoc by Claudia Emde when she wrote the DOIT part, so they are older than my polynomial interpolation routines. I missed to replace them when I introduced my routines. Luckily, they are used in very few places as far as I can see. It would be great if we could use this opportunity to get rid of them.

olemke:
> A programming question: You now consistently use Vector and Matrix return values. I assume this is efficient because of the new constructors you added. Correct? This seems quite a paradigm change. It would be good to have a discussion whether we want to generally encourage that also in other areas. Or not. The pros and cons are not clear to me at this point.

This depends strongly on the use case. Using return values is not penalised by a copy being made anymore thanks to the move constructor introduced in C++11. However, if the caller wants to write the result into a View type, using a return value still requires an additional copy in contrast to passing the target View as an input parameter.

stefanbuehler:
Ok, so a downside is loss of flexibility when it comes to views.

On the upside, output as return value allows nesting of function calls, 
as also used by Richard now, making the calling code a bit more compact. 
My feeling is, though, that this is particularly beneficial for small 
generic classes that are used a lot. Arithmetik operations being the 
best example, of course. For a “heavy” function (lots of input, 
substantial computation time) it seems to me not to have a strong 
benefit.

Are there other up- or downsides?

riclarsson:
Hi Stefan,

I don't want to touch the `polint` and `poly_interp` functions for now.  I agree that they can be removed but I think this should wait until the DOIT update.  There's a bit of complication removing them because they have an adaptive interpolation order.  I could add a function that computes this interpolation order if it is useful to have but I don't think I should mess too much with code that might be removed.  Please let me know when these changes are done and I can do the update to remove them then.

I've updated the documentation of `reinterp`.  It is basically Green interpolation.  I don't offer Blue interpolation, but Green+reduce does the exact same job.  I also changed the name from `LagrangeType` to `GridType` instead of from `Linear` to `Direct`.  The former was the poorly named construct, I think, since the grid-type is most important.

About the programming question.  As Oliver says, the move operations makes sure you don't do any copies returning something that allocates anymore.  I want to add that you also have enforced return value optimization (RVO) in C++17.  So if a function always returns the same named variable, or only returns unnamed variables, in all its return-statements, then it is the same as if that variable was created in the calling function and the move-constructor isn't even called.

The main benefit of returning rather referencing is that you can chain functions for some shorter commands, as you noted.  Note that this can mean move-initialization in the function call, so that there's even some additional benefits there (mainly if it is a local function variable that is not a reference by default, so perhaps not that important unless it's an Index or Numeric or std::array<Numeric, 4> that should probably never be const-references anyways).

Another benefit is if we want to have list of values.  You can use the, also somewhat new, emplace_back instead of push_back in Array<> to not have to copy variables into Array<> but rather construct them in place using your returning function.

Lastly, I think it also makes reading the code easier.  More like a math-statement than something where I have to remember the position of inputs.  You can even mark local variables as constant and just have them in the local loop that they are used in.  I am not sure if there are additional benefits to constant, but according to my understanding of the C++ standard, the compiler is allowed to just reuse the same memory block in the next call to new inside a loop.  I don't think we can make use of that though, since the constructors of our matpack-types are in the matpackX.cc-files.

olemke:
@riclarsson I've pushed two commits to simplify the conversion from `string_view` to `String`. Hope that's okay with you. The `string_view_to_string` function seemed not necessary at all, it could've been replaced with `std::string(sv)` instead. But with the string_view conversion constructor added to `my_basic_string` it's now even cleaner.

riclarsson:
Oliver, this is much better.  I did not know about that constructor.  I must have just tried String(sv)...

stefanbuehler:
Dear Richard,

> I don't want to touch the `polint` and `poly_interp` functions for 
> now.  I agree that they can be removed but I think this should wait 
> until the DOIT update.  There's a bit of complication removing them 
> because they have an adaptive interpolation order.  I could add a 
> function that computes this interpolation order if it is useful to 
> have but I don't think I should mess too much with code that might be 
> removed.  Please let me know when these changes are done and I can do 
> the update to remove them then.

Ok, you have a point there. I’m Cc-ing in Freddy here. Freddy, when 
you have some time over, could you please look for the deprecated 
“poly_interp” functions in the DOIT part, so that you are also aware 
of them? I hope their calls are in parts that you plan to make obsolete. 
Otherwise, we have to adapt the calls to use the new Lagrange 
funtiontions.

> I've updated the documentation of `reinterp`.  It is basically Green 
> interpolation.  I don't offer Blue interpolation, but Green+reduce 
> does the exact same job.

I’m sorry, but I still don’t get how this works. Could you add a 
simple example?

Also, the reduce functions themselves need documentation. Please 
describe them in Chapter 3 (Vectors, matrices, tensors, and arrays), 
including a simple usage example.

My C++ is quite rusty, unfortunately. What does the
   template <std::size_t dim0>
line do in front of the reduce fuctions?

> I also changed the name from `LagrangeType` to `GridType` instead of 
> from `Linear` to `Direct`.  The former was the poorly named construct, 
> I think, since the grid-type is most important.

I agree that GridType is much better than LagrangeType. However, I still 
think “Linear” as an option is both mathematically wrong and 
practically misleading. Wrong, because there is no such thing as a 
linear grid. (Or rather, perhaps one could use the term for a grid that 
is linearly *spaced*, but that is not how it is meant here.) The grid 
can be anything, the option just says that there should not be any 
transformation applied to the grid for the computation of the weights. I 
therefore still think that “Direct” would be more correct. And it 
avoids the cognitive dissonance caused by the keyword “linear” 
appearing in the arguments of the function for nonlinear (= higher 
order) interpolation. If you don’t like the word “Direct”, how 
about “Plain”, “Native”, “Standard”, or something in that 
direction?

I think this is the option that most people will want most of the time 
anyway. Would it make sense to make it the default value for type so 
that it can be omitted?

Cheers

Stefan

olemke:
> What does the template <std::size_t dim0> line do in front of the reduce fuctions?

Instead of passing the dim0 as an input parameter for the function, the function here is implemented as a template function. This has the advantage that the correctness of the passed dimension can already be evaluated at compile time by the use of 'static_assert'. The template also enables the compiler to generates a specifically optimised version of the function for each value of dim0.

stefanbuehler:
Ok, that’s approximately what I figured. But what’s the point here? 
The only thing that is static is the
“static_assert(dim0 < 3, "Bad Dimension, Out-of-Bounds”)”
statement (for reduce_rank of tensor3). That costs practically nothing. 
As far as I can see, everything else is as if dim0 was a normal input 
argument.

Richard, there is something else that I don’t understand about 
reduce_rank: Why is the object left in an empty state, and what does 
that even mean?

Intuitively, I would have thought reduce rank gives an alternative view 
of the same data, so would leave the data themselves intact.

And, can a Vector, Matrix or Tensor have “an empty state”? I don’t 
remember. It will be empty if one of the dimensions is 0. But if the 
dimensions are nonzero, then it is not empty. Or?

riclarsson:
The new type takes the `Numeric *` that the old type previously owned.  If the old's `Numeric *` is deleted, e.g. by the object going out-of-scope, then the new type's `Numeric *` would also be deleted if the old type still had a `Numeric *`.  So it cannot have that `Numeric *`, it must be left empty, i.e., pointing at `nullptr`.

The problem is perhaps made clear in code like this

```cpp
Vector a;
{
  Matrix b;
  ...
  a = b.reduce_rank<0>();
}
// a now still has a value.  If b was not left empty, the out-of-scoping of b would have deleted a's Numeric * otherwise
```

In shorter forms:

```cpp
const Vector a = func_returning_matrix(...).reduce_rank<0>();
//  The returned value only exists as an rvalue so is immediately deleted.  Thus a would have no value if the Matrix still had a Numeric *.  So the Matrix must be left empty
```

stefanbuehler:
I get it.

But wow, that’s quite some side effect. Reduce_rank destroys the 
original object for which it is called.

riclarsson:
Yes.  It is only supposed to be used in cases where you expect that the object is an rvalue.  The first example above should never be allowed to pass code-review.  At the very least, `std::move(b)` must be minimum so it is clear that `b` is invalid.

I am not sure if there is some C++ keyword to ensure that an operation only works when the object is in rvalue-state though...

olemke:
I think by adding '&' in the function definition, it'll ensure that it can't be called on a temporary:

```cpp
  template <std::size_t dim0>
  Vector reduce_rank() & {..}
```

olemke:
> I think by adding '&' in the function definition, it'll ensure that it can't be called on a temporary:
> 
> ```c++
>   template <std::size_t dim0>
>   Vector reduce_rank() & {..}
> ```

Just realized that my previous answer was the wrong way around. To enforce the object being an rvalue, you need && instead:

```cpp
  template <std::size_t dim0>
  Vector reduce_rank() && {..}
```

riclarsson:
Thanks Oliver,  I will add the && on every reduce_rank.  That way it is clear we are destroying the use of the original object.

olemke:
> Ok, that’s approximately what I figured. But what’s the point here? The only thing that is static is the “static_assert(dim0 < 3, "Bad Dimension, Out-of-Bounds”)” statement (for reduce_rank of tensor3). That costs practically nothing. As far as I can see, everything else is as if dim0 was a normal input argument.

You're right that performance is not the issue here. I would argue for convenience and error safety. Having `dim0` as a template argument enables the use of `static_cast` which produces a compile time error. Compile errors are easy to identify and impossible to miss compared to runtime errors (except if you have 100% test coverage which we don't ;-) ).

stefanbuehler:
Ok, I buy that.

As I wrote earlier, I think a usage example (in connection with 
reinterp) would help a lot.

riclarsson:
Stefan,

Several examples are added to the documentation.  This includes `reduce_rank<>()` initializations and uses for both standard matpack types and for interpolation.  I also added standard move documentation to chapter 3 and named the types of matpack initializations.  There was actually a small error in the documentation because of relative logic in the examples.

Now also because of Oliver showing how, it is impossible to compile ARTS with `reduce_rank<>()` in code where the object is named.  Unless you explicitly call `std::move()` on the named object, in which case you anyways expect that the object can no longer be used.

I also changed the code to have GridType::Standard and made the defaults arguments to most of the interpolation routines more in line with how old interpolate_poly works.

riclarsson:
Oliver,

I will make a nonstd::abs inside the Interpolation namespace for now that is constexpr.  I would rather keep the constexpr-ness than not here.  I think std::abs will never be made constexpr by the standard because it is probably a C-function in the back, making it impossible to make constexpr since it has to set and look at so many global variables.

riclarsson:
I think that is that for the comments.  Did I miss anything?

There is one more small thing about reduce_rank but I don't know if we have a solution.  There was talks about having an ARTS_ERROR that is a runtime_error when build with debug-symbols and completely removed when built in Release mode.  Is this a thing?  It would make sense to have this on the `size() not_eq out.size()` call because if there is an error there, it should be caught pretty much the first time you run the code on any non-zero data.  Which is what you do first so it is debug mode.

stefanbuehler:
Thanks for adding the documentation. It makes things a lot clearer. :-)

If I understand the example correctly, then reinterp+reduce_rank is not 
the same as blue interpolation. In blue interpolation, the interpolation 
is done to a sequence of points. Think a batch of 1-point 
interpolations, as shown in Figure 5.1. I think it could be implemented 
as a loop over your interp function.

If I understand your example right, then the reinterp+reduce_rank combo 
just gets rid of one of the existing dimensions. I think the same 
behavior could be achieved with the linear interpolation functions if 
one were to use an appropriate view in the input argument, together with 
a green interpolation.


Some minor cosmetics in other areas:

Page 20:
Vector i=func(...;
Is missing the closing bracket. (I don’t think I can push directly to 
your fork, otherwise I would have tried to just insert it.)

Page 53:
Having the example is great! But it is right now unnecessarily 
confusing, since you are mixing the steps of the linear interpolation 
and the O(1) Lagrange interpolation in the code. I like the idea to have 
both for comparison, but suggest you arrange like this:

// Simple linear interpolation:
GridPos gp;
…

// General Lagrange interpolation, special case interpolation order 1:
const LagrangeInterpolation lag(0, x, og);
…

It will be much easier to see the similarities and differences.

stefanbuehler:
Ah, I think that would require more discussion. We never reached a good 
conclusion on the errors, if I remember right.

riclarsson:
Stefan,

This should be the last of the fixes?

> If I understand the example correctly, then reinterp+reduce_rank is not the same as blue interpolation

I misunderstood the blue name.  I never had to use it to replace `interpolation_poly`, so it was unused.  Are you OK not having it?

As you say, doing these interpolations with a loop is very straightforward, and I think it will be slightly faster.  It definitely has a smaller total memory footprint.  Also, compilers are allowed to not call `new` every time a `const` variable is reinitialized by elision inside a loop as far as I understand it, so it might just be one allocation in total. (I don't know how to test this... and maybe having our constructors in the `.cc`-files hinders this optimization?)

Another way to do this is to have some `reinterpweights` and `dreinterpweights` that returns `Grid<MatPackType, 1>` and have an `interp` function that accepts this type and returns a `Vector` for all type of rank `Matrix` and higher when they see this type.  It's not that much work to add, but I would rather wait adding this for when it has some use.

> We never reached a good conclusion on the errors, if I remember right.

We should reach some agreement about this at some point.  I have too many opinions about this to list them all here.  In short, I would like to compile ARTS entirely without exceptions when I have something that works.  This version should ideally also set "noexcept(true)" on all of its functions.  I would also rather have a version of ARTS that throws `runtime_error` whenever anything goes wrong rather than sometimes doing `runtime_error` and sometimes `assert`.  This should of course have `noexcept(false)` all over the place.  Only on functions that demonstrably cannot go wrong (like boolean tests, and clear numerics that can at worst give `inf` or `nan`) can perhaps be made noexcept(true) still.

olemke:
> > We never reached a good conclusion on the errors, if I remember right.
> 
> We should reach some agreement about this at some point. I have too many opinions about this to list them all here. In short, I would like to compile ARTS entirely without exceptions when I have something that works. This version should ideally also set `noexcept(true)` on all of its functions. I would also rather have a version of ARTS that throws `runtime_error` whenever anything goes wrong rather than sometimes doing `runtime_error` and sometimes `assert`.

That was the consensus in the last discussion. Turning assertions into exceptions and replace them in time critical parts (e.g. boundary checks in `operator[]`), with an `ARTS_ASSERT(condition)` macro that uses exceptions. These can be completely removed in release builds for performance. In all other places `runtime_errors` should be used, possibly provided by additional convenience macros such as `ARTS_CHECK(condition)` or `ARTS_CHECK_WITH_MESSAGE(condition, message)`.

> This should of course have `noexcept(false)` all over the place.

Not necessary to put `noexcept(false)` all over the place since that's the default behaviour and doesn't need to be specified explicitly.

> Only functions that demonstrably cannot go wrong (like boolean tests, and clear numerics that can at worst give `inf` or `nan`) can perhaps be made `noexcept` still.

Agreed.

riclarsson:
I think we should move the error discussion elsewhere since this will soon be closed and therefore lost to history without any code changes.  One last thing though so I am clear:

> Not necessary to put noexcept(false) all over the place since that's the default behaviour and doesn't need to be specified explicitly.

If we had `ARTS_NOEXCEPT` as a macro that is either `noexcept(true)` or `noexcept(false)` depending on build options, then we could still but the tag on functions like:

```cpp
 // Const index operators:
  /** Plain const index operator. */
  Numeric operator[](Index n) const ARTS_NOEXCEPT {  // Check if index is valid:
    ARTS_ASSERT(0 <= n);
    ARTS_ASSERT(n < mrange.mextent);
    return get(n);
  }
```

And the function is now `noexcept` in Release build but throws a runtime_error i `Debug` mode.

stefanbuehler:
Dear Richard,

the main reason for doing it for a whole stack of points at once is that 
you do not have to start the grid position search from scratch. Instead, 
you take the position of the previous point as the start of the search. 
This is a lot more efficient, since searching is expensive.

You already take the starting point of the search as an input variable. 
The only missing part is to be able to feed the position of the previous 
point into this. If that was possible, then it would indeed be a more 
general and perhaps better solution than my blue interpolation.

So, I guess you would just need a public Lagrange::Pos() operator.

I noticed one more thing in the code: Please add Doxygen documentation 
for pos, lx and dlx, the data members of Lagrange.

All the best,

Stefan

riclarsson:
Stefan,

The LagrangeVector takes the previous position.  So what I think you would do is to set all of the ArrayOfLagrangeInterpolation you need once, and then use these LagrangeInterpolation inside your for-loop.

Also, the LagrangeInterpolation class is a struct.  pos and lx and dlx are all available to you by just standard struct access methods.  So you could if you wanted to store the last position outside and not even allocate an ArrayOfLagrangeInterpolation and feed best-guess position into it.

I will add the Doxygen documentation tomorrow.

stefanbuehler:
Ah, I see, the LagrangeVector class. Please add doxygen documentation 
also for that.

Then I think we have everything we need, you just have to explain the 
concept in the developer guide. (Calculate LagrangeVector in one go, 
then loop over its elements and apply interpolation for each.)

Right?

/Stefan

riclarsson:
Hi Stefan,

Do you want the LagrangeVector to be a class?  It is currently just an Array<Interpolation::Lagrange>, but I could just use Simon's trick and make it into a class so that the one way to construct it is via the current Interpolation::LagrangeVector method.  It seems a bit specialized, since there might be other ways to do this.

stefanbuehler:
Does it matter? Class is not necessary if functionality is the same. 
Main thing is that philosophy is explained clearly in developer guide 
and code comments.

riclarsson:
If it is a class instead of a typedef, it appears higher up in the doxygen documentation.  I am mostly asking because your comment about documenting LagrangeVector.  It is documented as a function.  It is also documented in the developer guide.  So the remaining thing would be to make it more visible in doxygen.

riclarsson:
I've added the discussion about blue interpolation now.  I am very much not sure which option is faster.  You have these two options (which I hope are clear from the names):

1) Using Interpolation::LagrangeVector
```cpp
Vector new_data(new_grids_nelem);
const auto t_lags = Interpolation::LagrangeVector(new_t_data, old_t_data, 1);
const auto p_lags = Interpolation::LagrangeVector(new_p_data, old_p_data, 1, 0.5, false, Interpolation::GridType::Log);
for (Index i=0; i<new_grids_nelem; i++) {
  new_data[i] = interp(old_data, interpweights(t_lags[i], p_lags[i]), t_lags[i], p_lags[i]);
}
```

2) Using LagrangeInterpolation with memory
```cpp
Vector new_data(new_grids_nelem);
std::array<Index, 2> pos0={start_pos_finder(new_t_data[0], old_t_data),
                           start_pos_finder(new_p_data[0], old_p_data)};
for (Index i=0; i<new_grids_nelem; i++) {
  const Lagrange t_lag(pos0[0], new_t_data[i], old_t_data, 1);
  const Lagrange p_lag(pos0[1], new_p_data[i], old_p_data, 0.5, false, Interpolation::GridType::Log);
  new_data[i] = interp(old_data, interpweights(t_lag, p_lag), t_lag, p_lag);
  pos0 = {t_lag.pos, p_lag.pos};
}
```

stefanbuehler:
Ok, then. I must have simply missed it, looking in the source code.

stefanbuehler:
For Option 1, due you internally do something similar? So, not start the 
search from scratch for each new interpolation point? I think that is 
important.

Do I understand right that you think Option 2 could be more efficient, 
because there is no need to allocated memory for the LagrangeVector?

riclarsson:
Yes on both options.

I use the last position as the first guess in the next iteration in `LagrangeVector`.  It was the thing that fixed the most of the issues with Oliver's original time comparison when this code was very slow.

And yes, it is potentially more performant to not allocate the two ArrayOfLagrangeInterpolation.  However, you are still allocating the Vector inside the loops (i.e., the lx and dlx, so it might do nothing).

riclarsson:
I'm pinging this again.

My feeling is that everything is sorted and it can be merged.  Best-use cases cannot be sorted without using the functionality in-place, I think.  Decent enough (as in slightly faster than previous interpolation routines) are available and documented now.

Since it touches so many files it is important that this get sorted quickly.  Many kinds of other commits will create issues here.  (Probably both the other active commits I have got up will have issues before this is done.)

olemke (approved):
Looks ready to be merged from my side.

stefanbuehler:
I’m also ok with the merge. /Stefan

PR #250, opened 2020-10-16
Minor fixes

erikssonpatrick:
Hej, 

With v2.4 out of the way, here is my recent work that should get into master. A mix of things. most finihed, but second order integration not. Richard has to look at it, but thinks it is simplest if my code gets into master.

/P


olemke:
I'll have a look at the failing test. It's caused by changes in master since your branch is not fully up to date.

Was the addition of the file `transmissionmatrix.h.bu` intentional?

PR #253, opened 2020-10-21
Add minimal documentation for the ARTS classes interface

riclarsson:
Seems like a bad thing that this is not autogenerated but I guess that's what it is.

This is just copying one of the other rst files and adding what it didn't do properly to the list.  It is possible that some things are missing.

### DIFF
--- a/python/doc/source/modules.rst
+++ b/python/doc/source/modules.rst
@@ -29,7 +29,7 @@ ARTS Data Types
 
 
 ARTS Class Interface
----------------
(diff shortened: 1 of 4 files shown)


olemke:
Sphinx feature for fully automatic generation of documentation lists is unfortunately very limited to say the least.

riclarsson:
Ok, well I hope this is enough to at least show the interface for anyone interested.  There's not much more to this as an interface.

PR #254, opened 2020-10-21
Temporary solution to over-shoot

riclarsson:
This is code trying to solve a problem of the second order integration routine having much larger source terms than expected.  It now falls back to first order if the first order term is smaller than the second order term.  This is partly in line with the ref I am using here but seems to have some other issues.

### DIFF
--- a/controlfiles/artscomponents/wfuns/TestTjacStokes4.arts
+++ b/controlfiles/artscomponents/wfuns/TestTjacStokes4.arts
@@ -16,8 +16,6 @@ INCLUDE "general/continua.arts"
 INCLUDE "general/agendas.arts"
 INCLUDE "general/planet_earth.arts"
 
-StringSet( rt_integration_option, "second order" )
-
 
 # Agenda for scalar gas absorption calculation
 Copy(abs_xsec_agenda, abs_xsec_agenda__noCIA)
@@ -152,8 +150,6 @@ yCalc
 #
 VectorCreate( yref )
 ReadXML( yref, "yREF4.xml" )
-Print(y)
-Print(yref)
 Compare( y, yref, 1e-4,
          "Calculated *y* does not agree with saved reference values." )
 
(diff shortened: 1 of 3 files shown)


riclarsson:
@erikssonpatrick Here's some code that makes the failing controlfile almost work.  I am unsure whether we need to have a lot more tests in place at a higher level of the code than this.  I also simplified the output a bit, since C++17 allows some code duplication to be removed.

erikssonpatrick:
@riclarsson Sorry for the late reply, but just too many things going on now. And no time for testing right now. "Almost work"? What are the differences in Tb now?

Right now I am don't need the Jacobian. So accuracy for Tb main concern. Of course good if also Jacobian can be obtained for second order. Seem that you are on it? Checks work after the last commit?

riclarsson:
The Jacobian are way wrong at this point.  I also don't have much time for this bit.

The other bit produces 
```
Max allowed deviation set to: 0.0001
but the vectors deviate with: -0.000451282
```

erikssonpatrick:
Seems that Tb-s are totally fine. So my suggestion, for now, is to add a check in iyEmission that if second order and jacobian_do is true, raise an error. And then make a pull request. I will then ask Vasilis to test second order in his comparison to RTTOV.

riclarsson:
The Tb are fine because it returns to use the first order approximations whenever it is smaller than the second order approximations.  This is what I understand they do in the ref you sent, but it makes little sense to me so it might be completely wrong.  I would like to see the derivation of the approximation sometime to make a better judgement myself...

riclarsson:
Anyways, I mark this for review since it does what you want it to do.  Hopefully I don't have too many errors introduced.

erikssonpatrick:
OK from me.

@riclarsson Maybe Freddy can help you to understand why the min between first and second order shall be used. He has been testing the scheme for DOIT quite carefully.

Issue #258, opened 2020-10-25
Second order integration


erikssonpatrick:
@riclarsson I made a test with the latest version of second order integration in iyEmissionStandard. I got basically identical results with first and second order integration. I assume this means that the max criterion has the result that the algorithm falls back to first order more or less for each RT step. Maybe this is what you meant in some comment you made. Anyhow, as it is now second order ends up pretty much useless as it does not improve accuracy, but still use more calculations.

But was it really necessary to introduce the max criterion? My memory is that the large deviation for y I reported about did not involve I, but the later Stokes elements. Maybe my memory is wrong, but maybe there was just a problem for higher stokes_dim and polarised situations?

erikssonpatrick:
With the comment above I wanted to give the input I can provide. For the coming weeks, I will have no free time at all and must all non-urgent things to the side. 

@m-brath I added you here as you have looked at the second order integration algorith most carefully among us. Could you try to help Richard? What do you say, is the max criterion necessary for clear sky (where maybe the absorption can vary strongly ver a RT step, but not the Planck function)? How should the expression be implemented for stokes_dim > 1?

riclarsson:
There was a problem for all Stokes dimensions in the test case you sent me.  We don't have the propagation matrix in this part of the calculations, so the max-condition must always be active to not encounter this issue.

As I have told you, we would have to change the RT setup to fix this, so that the "update_radiation_vector" function takes the propagation matrix too to test the limit.  This means its derivative and the distance plus its derivatives should also be sent in.

But before I will touch the code to do this, I need a full derivation of the second order source terms.  Because it is very strange to me that such a limit can at all be required.  The presented equation clearly allows for it this to happen but the image they show of the problem does not allow for this to happen.  I don't have the time for this derivation myself for the near future, but I won't touch the code before I have it.

m-brath:
> @m-brath I added you here as you have looked at the second order integration algorith most carefully among us. Could you try to help Richard? What do you say, is the max criterion necessary for clear sky (where maybe the absorption can vary strongly ver a RT step, but not the Planck function)? How should the expression be implemented for stokes_dim > 1?

I think for a pure clear sky atmosphere on earth the max criterion is mostly not needed, because the temperature and the absorption gradients over one rt step are mostly not strong enough. Nonetheless I would incorporate the max criterion within the solver. As I could not understand it from the code, how do you exactly define the max criterion?  For higher stokes dim, I have no answer yet. I have to think and read about it.

[...]

PR #259, opened 2020-10-26
Secordint

riclarsson:
This makes the code compliant with the source for second order integration (except using first order calculations rather than the delta-s version of Qmax; the source function never defines delta-s so I assume it is the distance but cannot know for sure).  Note that I leave a lot of fixmes around here.  This is not good code but it solves the problem.

The new output is close to the results from before but the second order integration case is activated more often.  This improves the forward results of TestTjacStokes4.arts, for example, because the output using second order integration is closer to the values if the pressure grid is ten times as dense of first order integration.

Note that I don't think this or the previous code should have been merged before we have a proper derivation of the equations and can redesign the entire absorption-to-RT step accordingly.  All of our RT algorithms are turning into spaghetti code at this rate.  For instance, one of the steps in these equations would work rather better if they had alpha and j rather than S and K.  The latter are not necessarily the same according to how we have defined them, so I am unsure whether or not this code can at all work with the iy-transmission code...  we currently throw away j and alpha after computing S.

### DIFF
--- a/src/m_fos.cc
+++ b/src/m_fos.cc
@@ -1909,6 +1909,15 @@ void iyHybrid2(Workspace& ws,
                             tot_tra[ip],
                             dlyr_tra_above[ip + 1],
                             dlyr_tra_below[ip + 1],
+                            PropagationMatrix(),
+                            PropagationMatrix(),
+                            ArrayOfPropagationMatrix(),
+                            ArrayOfPropagationMatrix(),
+                            Numeric(),
+                            Vector(),
+                            Vector(),
+                            0,
+                            0,
                             RadiativeTransferSolver::Emission);
   }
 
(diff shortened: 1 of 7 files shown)


riclarsson:
@erikssonpatrick and @m-brath This is code for #258 .  I had time to do this but if you look at the additions, it is extremely ugly code.  There could also be several bugs in this code because the indexing we are using is not very nice to work with.  I think the absorption-to-RT step should be changed such that the inner loops outputs more relevant information for the RT step, such as an array of Q and dQ rather than K and S.

riclarsson:
It would be fine to merge this.  The only difference potentially I would like is to make the solver a template argument because then there are several values that could have default inputs (and the functions will be slightly smaller for different RT schemes).  But I don't think this is necessary, because we have quite long function calls all over the place anyways, and the amount of calculations in this one is large enough that I think it is fine.  (The derivatives might still be very wrong.)

PR #263, opened 2020-11-12
Minor things

riclarsson:
These are many minor changes to some code.  The main thing is removing a very large function dealing with the setup of the Jacobian calculations because it wasn't really being used.  The rest are just constifying some variables and using more accurate maths in some of the derivatives.


olemke:
I'll remove the test_omp since it's outdated and not needed anymore.

PR #267, opened 2020-11-16
Switch to new iyHybrid solver.  Remove some old code from jacobian.cc/h

### LINKED ISSUE #205: Finish iyActive2
Make some checks and when done make this iyActive

### LINKED ISSUE #264: iyHybrid or iyHybrid2
There is a replacement of iyHybrid using the new interface.  If this is merged as the default, get_diydx in jacobian.h/jacobian.cc can be removed.

I made the interface but @erikssonpatrick  told me he or someone else would setup a test.  I no longer remember the state of this but would like to remove get_diydx since I had a minor look at jacobian.h/cc today.

Anyways, what is the status?  I think a decision needs to be made before ARTS 3.

riclarsson:
This closes #264.  According to @erikssonpatrick , it also closes #205.

I am protesting against closing #205.  We still have a `iyActiveSingleScat2`.  Also `iyActive` is still there using pure matpack-types instead of Eigen-types for calculations.  The latter is mostly its own thing though, so I am not sure if much can be gained*.  Anyways, Patrick wants #205 closed after this code is merged.

* It even accepts some `iy_unit` that is not supported by `iyApplyUnit`.  The latter's documentation must be updated to say what unit conversions are accepted, since it does not support all unit conversions there are in ARTS


erikssonpatrick:
Hej,

Yes, correct that we also have iyActiveSingleScat2. That got lost somewhere. For me perfectly OK to remove old version and make new version iyActiveSingleScat. As for iyHybrid, most important right now is to get old code removed so we can focus on what's left. This is e.g. important for Simon's big revision. Anyhow, we at Chalmers can take care of checking iyHybrid and iyActiveSingleScat. 

If you attack iyActiveSingleScat in this pull request or if you do it in a later, I leave to you to decide.

Good point with iyApplyUnit. I will look at that. Right now considering to introduce different iy_unit for emission and radar. Would make sense as we will mix the two observation types and will have two units now and then.

riclarsson:
I'll update this so that the other "2" is removed before merging so that this commit truly does close #205 .

riclarsson:
This can be merged pending working checks. I changed my original comment so that the auto-close github feature should work for #205 .

As I said in the commit:  I needed to remove artscomponents/radar/TestIyActive_wfuns.arts from the active tests (pun not intended).  There is likely some error in my handling of the indices.  It is mostly manifest in the last columns of the Jacobian.  Commit #24fbe29 should be corrected at some point by your tests at Chalmers.  Good luck, I hope the code is understandable.

erikssonpatrick:
Do I get it right, that the jacobian part of iyActiveSingleScat is not working? That's OK for the moment. But can I just suggest then to make sure that the user get an error if trying to calculate the Jacobian with that method.

For the rest OK from me.

riclarsson:
OK, I have added a throw when jacobian_do is true.  I am not sure how wrong this is but it is the last index that is different.

erikssonpatrick:
I will try to look at the Jacobian issue some day. We take this and can close an issue!

That is, OK from me.

riclarsson:
@erikssonpatrick I think that since I have asked for you as a reviewer, the formal way to ensure @olemke knows he can merge this is for you to press the "Accept" or "OK" button at the place where it says that you are the reviewer.

Issue #136, opened 2020-03-02
ECS Model needs fixes


olemke:
This issue keeps track of fixing the issues in commit 08de9d4b3a66199d3b6b66841a42707a1cd56a16. If the problems cannot be resolved, the code needs to be removed again at some point.

olemke:
See caveats mentioned in https://github.com/atmtools/arts/pull/281#issuecomment-762235812

PR #272, opened 2020-12-03
Makarov2020 o2

riclarsson:
This is a bit of a mess of a commit.  Three things: 1) I am adding a few more ways to plot, mostly multiple lines, and showing an example in test_gui.cc.  2) I am changing how to set line shape parameters from within a controlfile.  3) I am attempting again to implement Makarov etal 2020 O2 line mixing ECS model.  I think I am very close this time, and that it is probably possible to use the model without much hassle.  I still need to confirm this but it is a good step towards finishing #136 .

Anyways, as a test for how to use the debug GUI, I added a simple test_gui build that shows the below plot:
Screenshot from 2020-12-03 16-37-44

The shown plot are the differences between my implementation of ECS and MPM and pure line-by-line based on Makarov etal 2020.


riclarsson:
I think this can be merged.  There are probably still some bugs but nothing that should be a show-stopper before it is tried out.

The code attempts to solve the GUI debug documentation thing and it mostly solves the ECS implementation errors.

For the latter, there are still differences between online ECS and my own.  I don't know where they are from but suspect that I am either coupling too few lines or that the online version has a bad Wigner-implementation (I have seen the latter in some old Fortran code that in my previous tests turned out to be very inaccurate, so this is why I am suspecting it.)  I hope it is the former because otherwise we have to use slower and less accurate Wigner symbols in ARTS to use online results.

olemke:
I will remove the 9d87062 commit (from #267) once more from this PR before merging to keep things clean(I did this this morning, but you just force-pushed it back in ;-) ).

olemke:
I will remove the 9d87062 commit (which should only be in once more before merging (I did this this morning, but you just pushed it back in.

riclarsson:
Thanks, Oliver!  I still don't understand fully how this platform works.  Force pushing is the only thing that seems to work or there is just a big warning message from git.

PR #273, opened 2020-12-07
My work for v3

erikssonpatrick:
Some smaller additions and bug fixes. The last commits are the bug fixes and why I select to make a pull request. 

The bug fix should also get into v2.4. That is, the changes in m_microphysics.cc and psd.cc should be copied to v2.4. Can you fix that, or shall I do it? (No need to add the new test to v2.4). Getting this into v2.4 should hopefully fix the problems reported by Renish.


olemke:
Yes, I will backport those to 2.4.

PR #277, opened 2020-12-10
Add custom check-deps target

olemke:
'make check' did not trigger a build of test_cpp_api. This commit
introduces a custom target 'check-deps' to which essential dependencies
for checks can be added to. Currently contains arts and test_cpp_api.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -411,25 +411,27 @@ if (CMAKE_GENERATOR STREQUAL "Xcode")
   set (CTEST_MISC_OPTIONS -C \${CONFIGURATION})
 endif (CMAKE_GENERATOR STREQUAL "Xcode")
 
+add_custom_target(check-deps)
+
 add_custom_target(check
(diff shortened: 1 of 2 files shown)


simonpf:
Hi Dave,


Nice manuscript! I liked the bit on the history of satellite observations in the

introduction. Below are my comments on the figures:


* Fig. 1:
  - Not quite convinced that this needs to be a figure
  - Consider adding a legend
  - Having this as figure may be more justified if you would
    add the time dimensions as well.
* Fig. 2:
  - IMO this should be a bar plot because channels are not
    a continuous quantity
  - Using a broken axis or two axes may be
    a good idea here: https://matplotlib.org/3.1.0/gallery/subplots_axes_and_figures/broken_axis.html
* Fig. 3, 4, 5:
  - The only thing I would probs change is to put the grid behind
    the bars using. Should work by just adding the zorder=30 to
    the bar command.
* Fig. 6
  - Labels are too small
  - Lines should be thicker
  - Conf. intervals in a and c are hardly visible
  - You could try to instead fill the range between intervals
    with a relatively low alpha value
* Fig. 7
  - See Fig. 3, 4, 5
* Fig. 9
  - You seem to use latex to render unit, which is why they appear
    as italics. Formally it would be more consistent to use \text
    or \unit commands, but that may require to modify the latex
    preamble in you matplotlib_rc. Here's what I use:

    text.latex.preamble : \usepackage{amsmath}\usepackage{lmodern}\usepackage{units}

I thin for consistency it would be nice to use the same background for all plots
but that could also just be me being pedantic.



________________________________

PR #283, opened 2020-12-15
Methods for errors in debug.h and matpackI.cc/h

riclarsson:
This adds some code that makes it possible to use ARTS specific terms for error handling.  This comes after discussion we had a long time ago that `assert` is not really desirable because it terminates.

This commit therefore includes 5 MACROS:
1) `ARTS_CHECK(condition)`
2) `ARTS_CHECK_WITH_MESSAGE(condition, ...)`
3) `ARTS_ASSERT(condition)`
4) `ARTS_ASSERT_WITH_MESSAGE(condition, ...)`
5) `ARTS_NOEXCEPT`

The conditions has to be true for `ASSERT` and false for `CHECK` statements.  If not, the functions may throw a `runtime_error`.  The message of this `runtime_error` will be `"Failed Check/Assertion: " #condition "\n"` in the basic case.  In the variadic expressions (`*WITH_MESSAGE`) the triple dots represents any number of arguments and these arguments will all be parsed in order by their respective `operator<<` he resulting `String` is also output to the `runtime_error`.  In essence the same as:

```cpp
if (condition) {  // For CHECK
  std::ostringstream os;
  os << ...;  // Though with one call to << per argument, i.e., use '\n' as a separate argument for newline
  throw std::runtime_error(String("Failed Check/Assertion: " #condition "\n") + os.str());
}
```

The last MACRO, `ARTS_NOEXCEPT` is just there so that functions that cause runtime_error during development can still be `noexcept` in the end if there's nothing that is allowed to go wrong after release.

To be very clear, if you build ARTS with `-DCMAKE_BUILD_TYPE=Release/Native` then the first 4 MACROS do nothing and `ARTS_NOEXCEPT` is defined as `noexcept(true)`.  If you build with the standard build type or some other way define `NDEBUG` while building ARTS, you will have the `runtime_error` with the first four and `ARTS_NOEXCEPT` is defined as `noexcept(false)`.

I have gone over `matpackI.cc/h` to add `ARTS_NOEXCEPT` where I think it can be added just to show how it is used.  Note that most `std`-interfaces and C-functions don't allow it so it should not be used.  In short though, if you use it is fine because the code will likely be as fast, it might be faster, as before.  It is just that you will get terminate instead of `runtime_error` in release builds.  (I checked only `make check` and it was at most some 0.1 s faster with all the new `ARTS_NOEXCEPT` so I don't think it does anything at all there, but perhaps `matpackI` is underrepresented in the tests.)

The last thing to discuss here is whether we should also have an additional two MACROS:

6) `ARTS_USER_CHECK(condition)`
7) `ARTS_USER_CHECK_WITH_MESSAGE(condition, ...)`

These would be defined the same as the other two `CHECK`-MACROS but would only be disabled, i.e., be ignored, if `-DARTS_USER_INPUT_CHECK=1` is set while building ARTS (and ARTS is built without `NDEBUG`?).  I suggest these should be included as a way to disable user errors in our methods.

Some methods do quite extensive tests.  For example, the Zeeman code performs some 20 tests every time it is called (every layer).  These vary from simple index checks that will never slow down the code execution to some more advanced tests that ensures that the minimum value of a `Tensor4` is positive or that all species follow the expected pattern.  The latter can be non-trivial.  (`min freq > 0`, for instance, is perhaps unnecessarily slow when `f_grid` gets close to 100'000 numbers.  At least after the first check.  And since most spectrometers I have have some 65k channels, I am doing at least a few needless computations.  However, `propmat_clearskyAddZeeman` is a workspace method so each variable have to be checked separately and independently from what you would normally expect from them.)


stefanbuehler:
I really like your system, especially that one can still build up complex error messages. Very elegant!

Is it possible to somehow make the message arguments optional?

Then we could get away with only two macros, ARTS_ASSERT (for all internal errors) and ARTS_USER_ERROR_IF  (for errors triggered by the user), both taking the optional message input. Usage example for the two typical use cases:

ARTS_ASSERT { a<5 }

ARTS_USER_ERROR_IF { b > a,
	                                  "The value of b (", b, ") is too large." }

It is not completely logical that for internal errors we throw when the condition is false, and for user errors when the condition is true. But on the other hand we are very used to this so I don't really see an advantage in trying to be more general.

In "debug" mode both of these would be enable, in "release" mode only the second, and in "reckless" mode none.

stefanbuehler:
Another idea:

If the error functions were defined as class members (in the .h file), would not the compiler optimise the calls completely away in case the body is empty? I think if the problem can be solved with a regular class instead of macros, it would be preferable. With a class "artserror" it could look like this:

Old assert:
artserror.internal_ifnot( a<5 )

Old runtime error:
artserror.user_if( b>a, 
                             "The value of b (",  b,  ") is too large."  )

Would it work to simply modify the body of these functions, based on the build setup? If all the error handling goes through a single class, it would be a nice unified design.

riclarsson:
Functions would work.  As Oliver says, the namespace+function rather than class+static function seems better.  I disagree with Oliver on empty function bodies being a problem.  With [[maybe_unused]] one issue is gone.  GCC also optimizes away empty function calls on O2/O3 when it can see the function body, which it can since we have to have the function body in the header with the variadic templates required here.  They would still be compiled into the binary, but that's not too big of an issue, I think.

I still prefer macros for other reasons.  Here is the ARTS_ASSERT call in my current implementation:
```cpp
ARTS_ASSERT(n < 0)
...
Failed Internal Assert: n < 0
This is a bug! Bug is found at:
	/home/larsson/Work/arts/src/test_gui.cc:15
Please contact ARTS developers so we can fix our error(s) via:
	github.com/atmtools/arts
```

For an equivalently short call as a function:
```cpp
ArtsError::internal_if(n < 0)
...
Failed Internal Assert.  This is a bug!
Please contact ARTS developers so we can fix our error(s) via:
	github.com/atmtools/arts
```

The two advantages of the macro-message comes from the fact that it can use `__LINE__`, `__func__`, `__FILE__`, and `#condition` to form the message.  The function will only have a bool, and the other three pre-defined macros (that are part of the C++-standard), they would point at the error-function rather than to the function where the error occurs.

stefanbuehler:
Yes, knowing the place in the code where the assert is is a big 
advantage. I rest my case and vote also for the macro solution.

olemke:
@riclarsson Good point on the code location.

riclarsson:
Technical notes follows.

Now these four macros are defined:

1) `ARTS_ASSERT(condition, ...)`
2) `ARTS_NOEXCEPT`
3) `ARTS_USER_ERROR_IF(condition, ...)`
4) `ARTS_USER_NOEXCEPT`

The first two are defined as expected when `NDEBUG` is set/unset.  The latter two are defined as expected if the new CMake flag `-DNO_USER_ERRORS` is set as something that CMake `if()` evaluates as true.  The flag `NO_ARTS_USER_ERRORS` is passed as C and CPP flags to ARTS compile-time, it is 0 when the CMake statement evaluates as false and 1 otherwise.  Inside `debug.h`, this flag is used to define the macros.  I am not against changing the `NO_ARTS_USER_ERRORS` to something that demands only existence instead of being a specific value, but I went with values because I find them easier.

The 1) and 3) macros can technically be called with any number of additional arguments to form their message.  This is not what we want for 3); we want it to have at least some error message. To fix that issue, I added a `static_assert` inside its macro checking that there is at least one commend following the original.  To count the number or arguments, I had to use tuple because I could not make any other version of this work as I wanted it to work.  This means `debug.h` now depends on `<tuple>`.  If there's a way to count the arguments without `<tuple>`, that would be great.  The other way to make 3) fail is to simple add `__VA_ARGS__` as the last argument to the variadic template that builds the message.  The error message will then be such a complete mess that no one can possible understand what is going wrong, though, so the `static_assert` is really useful.  Note that I do not know of any `constexpr` way to form a char-array out of integers, so the static assert only reports on the error file.  My compiler/IDE gives me the exact line both of the macro and of the `static_assert`, so it is known and possible to debug, but it would be neat to have to `__LINE__` as well.

I did not use `__func__` to give more information about the errors.  I wanted to use it, but it caused me some issues with compiler errors I did not understand.

riclarsson:
Here follows some non-technical note.

I updated the main CMake file to give information about build type as the last status message.  You will see 

`Building ARTS in RelWithDebInfo/Release/Native mode`

when you run `cmake` now.  It will report only one of the modes.

The error messages from `ARTS_ASSERT` gives the filename and line number of the error.  It also gives a link to here (github.com/atmtools/arts), calling the error a bug and claiming that it is our fault and that they should contact us to try and find out what went wrong.  Is this OK?

The error messages from `ARTS_USER_ERROR_IF` also gives the filename and line number of the error.  This is perhaps not needed?  I like it because it helps me when I am setting up controlfiles, but it does not help non-developers at all.  The error message also strongly claims that the error is entirely the user's fault and that the error message from `__VA_ARGS__` is there to help them solve the issue.  Should we also report an email or even the github issues page?  I think the latter needs to be updated so that there is a user-question templates.  (I think these templates should also have a feature-request template, since I was told by email that there is a bug in xml-data earlier today.)

stefanbuehler:
> The error messages from `ARTS_ASSERT` gives the filename and line 
> number of the error.  It also gives a link to here 
> (github.com/atmtools/arts), calling the error a bug and claiming that 
> it is our fault and that they should contact us to try and find out 
> what went wrong.  Is this OK?

That’s great.

>
> The error messages from `ARTS_USER_ERROR_IF` also gives the filename 
> and line number of the error.  This is perhaps not needed?  I like it 
> because it helps me when I am setting up controlfiles, but it does not 
> help non-developers at all.  The error message also strongly claims 
> that the error is entirely the user's fault and that the error message 
> from `__VA_ARGS__` is there to help them solve the issue.  Should we 
> also report an email or even the github issues page?

Reporting the place in the code, although not strictly logical for this 
type of error, cannot hurt, I think. I would vote against the issues 
page link, though, since these errors normally should not be related to 
ARTS issues. Better not point users in that direction.

> I think the latter needs to be updated so that there is a 
> user-question templates.  (I think these templates should also have a 
> feature-request template, since I was told by email that there is a 
> bug in xml-data earlier today.)

That was me. When I created the issues, it was first labeled “bug”, 
then I changed it to “documentation”. Perhaps I just didn’t use 
the interface in the best way.

riclarsson:
And related for the remaining of this pull request draft before merging.

I would like to replace all `assert` with `ARTS_ASSERT` before merging this.

@olemke Is it possible to add a meta-class to array.h for the printing here?  I think something like:

```cpp
template <typename ArrayType> struct ColumnPrinter {
const ArrayType& ref;
constexpr ColumnPrinter(const ArrayType& inref) noexcept : ref(inref) {}
friend std::ostream& operator<<(std::ostream& os, const ColumnPrinter<ArrayType>& arr) {
  for (auto& x: arr.ref) os << x << '\n';
  return os;
}
};
```

would be useful.  (I haven't tested this yet but it would solve quite a few problems with printing lists.)  I just want to know if you are against the concept.  I don't also know where the `ColumnPrinter` class would be placed, perhaps array.h is not the best?  Making it such a generic template means that it is possible to use it with `Vector` as well.

@stefanbuehler I need your input about the changed error messaged.  The `ColumnPrinter` struct above would help here to restore the old error messages.

riclarsson:
> That was me. When I created the issues, it was first labeled “bug”, then I changed it to “documentation”. Perhaps I just didn’t use the interface in the best way.

I know.  The only default template to report anything is to report a bug.  The way to open a new request is to not use the big green "Get Started" button but instead press the Open a blank issue blue text hidden away at the bottom.

We are mostly using these issues to ask for features, like you did wanting to have the documentation improved.  I think having at least two templates, one called "Documentation request" and one called "Feature request" with relevant subtexts would help to ensure that the right tags are set.  I have no idea how this works though, so I am not going to go around playing with the `.github` stuff

olemke:
> @olemke Is it possible to add a meta-class to array.h for the printing here? I think something like:
> [snip]
> would be useful. (I haven't tested this yet but it would solve quite a few problems with printing lists.)

Why is the default `operator<<` for `Array` not sufficient? Which problems are you trying to solve by this?

>I just want to know if you are against the concept. I don't also know where the `ColumnPrinter` class would be placed, perhaps array.h is not the best? Making it such a generic template means that it is possible to use it with `Vector` as well.

Best would be to put it in a separate header file, e.g. `output_formatters.h`. As the class is independent of other ARTS includes, having it separately ensures it can then be included without pulling in unneeded dependencies. Note: As the class is generic, `Iterable` would be a more fitting typename instead of `ArrayType`.

olemke:
> We are mostly using these issues to ask for features, like you did wanting to have the documentation improved. I think having at least two templates, one called "Documentation request" and one called "Feature request" with relevant subtexts would help to ensure that the right tags are set. I have no idea how this works though, so I am not going to go around playing with the `.github` stuff

I'll add templates for those cases. We can iterate further on the details when I open that PR.

riclarsson:
> Why is the default `operator<<` for `Array` not sufficient? Which problems are you trying to solve by this?

Lines 573-580 of `m_abs.cc` in the pull request from 20 hours ago (new code).  The old way of throwing errors allowed you to say that you know the meta data of the array-variables when you format the printing.  The new way assumes that the variables can print themselves.  This means that there is more clear information in the old error message.  (It's an old design decision to have many separate variables for doing a single thing - the meta-information is not part of the printing but only something that a developer can know.  Perhaps having that format-printing is too much and it would instead be better to have a continuum meta-data `meta2String()`?)

> Best would be to put it in a separate header file, e.g. `output_formatters.h`. As the class is independent of other ARTS includes, having it separately ensures it can then be included without pulling in unneeded dependencies. Note: As the class is generic, `Iterable` would be a more fitting typename instead of `ArrayType`.

OK.  First I need Stefan's input on the above.

stefanbuehler:
These template functions are so hard to understand! :-)

I can see how a ColumnPrinter can be useful, so I would be ok with adding it. But then with good documentation (particularly important are usage examples), to encourage reuse.

If it is just needed to solve this particular error message, an alternative solution would be, as you suggest, to make a dedicated function that does the same thing, instead of the template function.

I think I prefer this simpler solution, simply because I think the general one will not be used enough to justify the complication. It also has the advantage of complete flexibility (in other places, perhaps it is better to present the mix of data and meta-data in a different way, and then there can be dedicated functions also for that).

olemke:
I agree with Stefan on having a dedicated function. The generic ColumnPrinter would not fully solve the issue for this particular error anyway. While it would print the values column-wise like in the original error code, it would still not output the additional index and meta data name for each item. A dedicated function would solve that.

riclarsson:
I went through all the files and made assert into `ARTS_ASSERT`.  I also went through the files to make a few throw-statements into `ARTS_USER_ERROR_IF`.

I agree with both of you and did not go with the idea of using a `ColumnPrinter` class.

I think this can be merged if it goes through the tests.

Note that it might also be a good idea to have one test compile this in release mode on at least one compiler.  The `noexcept` marking changes the function signature, so if there is a mismatch between a `.cc` file and a `.h` file with this, then this will be a compile-time bug.  I am not sure how the layout of release versus debug builds are with the checks on github.  These changes makes the debug build run significantly slower, meaning it is more recommended than ever to use ARTS in Release build mode when running code that needs to be quick.

riclarsson:
Also, should I revert the additions of `[[maybe_unused]]`?  There are too many places that would need these that there is no point in updating all of ARTS to take this into account.

stefanbuehler:
I didn’t get what problem you want to fix with the `[[maybe_unused]]` 
additions.

riclarsson:
I know many people claim that exceptions are very slow.  Even having them in the code at all makes the code slower, even if the actual exception call is never invoked.  GCC claims that having exceptions makes the code some 7% larger.  Testing the make check versus master-branch, I get some (23.34 s) vs (27.27 s).  The slower being the new code.  I would guess that matpack gets significantly larger this way, with all the additional exceptions.  So much so that it creates these issues.  In release, the differences are not there (actually, the new code is faster in my tests but I don't get why so I assume something bad on my computer).

I originally added it in m_abs.cc to remove compiler warnings when using `-DNO_USER_ERRORS`.  But too many functions would require it for it to be worth it to add this all over ARTS.  So I am asking if I should revert the additions to `m_abs.cc` or if it is fine to leave it in?

stefanbuehler:
> I know many people claim that exceptions are very slow.  Even having 
> them in the code at all makes the code slower, even if the actual 
> exception call is never invoked.  GCC claims that having exceptions 
> makes the code some 7% larger.  Testing the make check versus 
> master-branch, I get some (23.34 s) vs (27.27 s).  The slower being 
> the new code.  I would guess that matpack gets significantly larger 
> this way, with all the additional exceptions.  So much so that it 
> creates these issues.  In release, the differences are not there 
> (actually, the new code is faster in my tests but I don't get why so I 
> assume something bad on my computer).

I suspect the issue is something different. A quick googling turned up 
this quote: “[…] Exceptions are slow on the exceptional path, but 
they are otherwise quicker than explicit checks (if strategy) in 
general.”

So perhaps the problem is not an overhead for exceptions, but rather the 
explicit if statement in the definition of the ARTS_ASSERT macro. 
Perhaps the standard implementation of ASSERT is doing something that is 
faster than a normal if statement?

Well, either way, I guess this can’t be helped. And there is the 
release mode (or even the reckless mode) option for anyone that really 
needs speed.


About `[[maybe_unused]]`:

> I originally added it in m_abs.cc to remove compiler warnings when 
> using `-DNO_USER_ERRORS`.  But too many functions would require it for 
> it to be worth it to add this all over ARTS.  So I am asking if I 
> should revert the additions to `m_abs.cc` or if it is fine to leave it 
> in?

I still don’t understand what is causing the warnings and how 
maybe_unused helps to avoid them.

riclarsson:
In the reckless mode, a lot of variables in a lot of functions are not going to be used at all since if-statements are thrown out.  So gcc says so in it's warning messages. I thought this would be a small issue when I tested it in m_abs.cc.  When I then triggered it again just now, the reckless build that is, there are too many warnings to fix.

It's the reckless mode, so seeing warnings might be good,. My question was if I should revert this in m_abs.cc or if I can leave it in.

riclarsson:
About exceptions. Your comment is true if you deal with the error. In the past, we simply didn't deal with the error. So it's not really a comment about the slowdown I see here.  Anyways, we do have release so I don't think it's an important issue.

olemke:
> I didn’t get what problem you want to fix with the `[[maybe_unused]]` additions.

@stefanbuehler Some function parameters are only passed to be used in user checks. If the checks are removed, the compiler throws a warning that those function parameters are unused.

@riclarsson I added a commit that turns unused parameter and unused variable warnings off if NO_USER_ERRORS is in use. Don't know if it's really worth to care about clean compiler output in reckless mode, since it's a half-broken mode by causing undefined behaviour of the resulting ARTS binary anyway. ;-)

There are some other warnings triggered by NO_USER_ERRORS. Mainly caused by `if..else if...else` statements where the check is in the else statement. With the check removed, this causes `non-void function does not return a value in all control paths [-Wreturn-type]` and `variable 'X' is used uninitialized whenever 'if' condition is false  [-Wsometimes-uninitialized]`. I left those on for now.

stefanbuehler:
> It's the reckless mode, so seeing warnings might be good,. My question 
> was if I should revert this in m_abs.cc or if I can leave it in.

I guess, since this is solved more generally by Olivers configuration 
adjustment, and for consistency, probably better remove the statements. 
(But I have no strong opinion on this, since they don’t do any harm.)

riclarsson:
@olemke OK, that makes more sense.  I like it.

@stefanbuehler I tried compiling `matpack` with `-fno-exception` and it gets some 3% smaller with no significant changes in runtime of make check so you might be correct that exceptions are fast enough to not notice.  (My change was in Release build mode, so there might be some additional issues.)  So I still don't know why debug is so slow now.  Anyways, I removed the added maybe_unused attributes.

Finally, I added some documentation in `developer.tex`, changing the assertion section and adding a short if-then-throw section.

This still requires significant changes in quite a few places of ARTS to be fully incorporated.  The `assert` → `ARTS_ASSERT` change is done.  However. I cannot alone fix all the `runtime_errors` as `ARTS_USER_ERROR_IF`.  There are too many places where this is used, and not always is the logic as clear as if-then-throw.  There are also many cases where I find a if-then-throw statement that has simply been commented away, or where there is some sort of deep nested logic between the unction working normally and it throwing an error.

Therefore I cannot fix these things alone.

The types of errors I have noted that are difficult to fix are these:
1) User string option (long series of if-then-else-if-then-else-throw.  The ones I found before giving up were given `ARTS_USER_ERROR_IF(true, ...)` as their throwing routine.
2) Intermixed errors.  In several places there are if-then-else-if-then-throw-else.  I flattened the ones I found but this easily makes a mess of the original logic.
3) In some places the ostringstream is filled up conditionally.
4) Using ostringstream as a conditional cout-output.

The problems here are in order:
1) This is not nice since it will read "User Error: true", before giving the rest of the error message.  I think user options should generally be handled nicer.  Strings options should be turned into constant expression views in std::array(s).  This would make it straight forward to check the user input because "it has to be in the list".  But this was not enforced in the past so it would be too difficult to fix now.
2) These interchanged-logic errors are difficult to deal with without having written the function.  I think I failed flattening something like this in `m_jacobian.cc`, which is when I gave up translating files and reverted it.
3) These places I have used the `bool ? if_true : if_false` with `var_string(...)`, the variadic string creator I added here, to keep the same logic.  This is significant work and since we don't generally throw I am not sure I am adding some errors here.
4) tmatrix is one of these but I also found it elsewhere.  It's weird to not separate error checking cleanly from correct code, the logic becomes difficult to deal with when you did not write the original code.  (calcSingleScatteringDataProperties is really weird to look at, it looks like it claims that the calculations have failed before even starting to perform any calculations at all!  That or it is using try-catch blocks instead of if-else blocks...)

riclarsson:
If you accept the documentation and that the updates to use `ARTS_USER_ERROR_IF` fully must be gradual (and enforced for new features/when touching old features), then this is ready to merge now.

olemke:
About the speed differences: It all comes down to the additional amount of code in the if-block. I assume it affects how the compiler is optimizing the overall code of the function where the assert is used. The additional overhead might impact performance particularly for smaller, inlined functions such as access operators. Another factor could be that the compiler has to add more state handling code if the function can actually return, instead of just immediately abort execution at that point.

Here are some benchmarks for `make check-all-controlfiles` in the master branch and for different variations of the implementation of ARTS_ASSERT in the artserrors branch:

```
master:
RelWithDebInfo: 190s
Release: 165s

artserrors:
RelWithDebInfo: 215s
RelWithDebInfo (modified ARTS_ASSERT to use cerr+abort() instead of throw): 205s
RelWithDebInfo (modified ARTS_ASSERT to no error output+abort() instead of throw): 190s
Release: 165s
Reckless: 165s
```

Looks like this is just the price we have to pay for better error handling.

@riclarsson A note on your exception test. I think something is fishy there. `-fno-exceptions` turns off exception support in the compiler and every throw/try/catch statement present in the code will generate a compiler error. Therefore it is not possible to compile ARTS with that option.

riclarsson:
Oliver, I turned them off manually, the try-catch and throw-blocks that is.  That's how I checked it.  There's not that many, though you do get quite a few warnings since new can return nullptr...

olemke:
> Finally, I added some documentation in `developer.tex`, changing the assertion section and adding a short if-then-throw section.

Looks good. 👍 

> This still requires significant changes in quite a few places of ARTS to be fully incorporated. The `assert` → `ARTS_ASSERT` change is done. However. I cannot alone fix all the `runtime_errors` as `ARTS_USER_ERROR_IF`. There are too many places where this is used, and not always is the logic as clear as if-then-throw. There are also many cases where I find a if-then-throw statement that has simply been commented away, or where there is some sort of deep nested logic between the unction working normally and it throwing an error.

I think we need to cover cases where a throw can occur without including an explicit condition attached. We should consider adding an ARTS_USER_ERROR macro that throws without any condition attached. This could then also be used in the catch block of try...catch constructs to re-throw the exception with additional error information, e.g. [like here](https://github.com/atmtools/arts/blob/194939162abe2ca73327126c09b74ab43a31a98f/src/xml_io_compound_types.cc#L1227)

I also agree that it is too much to do all those replacements as part of this PR. This is better to be done incrementally for different parts of ARTS once this has been merged.

> 
> Therefore I cannot fix these things alone.
> 
> The types of errors I have noted that are difficult to fix are these:
> 
> 1. User string option (long series of if-then-else-if-then-else-throw.  The ones I found before giving up were given `ARTS_USER_ERROR_IF(true, ...)` as their throwing routine.
> 2. Intermixed errors.  In several places there are if-then-else-if-then-throw-else.  I flattened the ones I found but this easily makes a mess of the original logic.
> 3. In some places the ostringstream is filled up conditionally.
> 4. Using ostringstream as a conditional cout-output.

All these might also benefit from an ARTS_USER_ERROR macro.

> 
> The problems here are in order:
> 
> 1. This is not nice since it will read "User Error: true", before giving the rest of the error message.  I think user options should generally be handled nicer.  Strings options should be turned into constant expression views in std::array(s).  This would make it straight forward to check the user input because "it has to be in the list".  But this was not enforced in the past so it would be too difficult to fix now.
> 2. These interchanged-logic errors are difficult to deal with without having written the function.  I think I failed flattening something like this in `m_jacobian.cc`, which is when I gave up translating files and reverted it.
> 3. These places I have used the `bool ? if_true : if_false` with `var_string(...)`, the variadic string creator I added here, to keep the same logic.  This is significant work and since we don't generally throw I am not sure I am adding some errors here.

I think avoiding these for now is better than potentially breaking them. Refactoring them should be discussed separately to bring this pull request to an end first.

> 4. tmatrix is one of these but I also found it elsewhere.  It's weird to not separate error checking cleanly from correct code, the logic becomes difficult to deal with when you did not write the original code.  (calcSingleScatteringDataProperties is really weird to look at, it looks like it claims that the calculations have failed before even starting to perform any calculations at all!  That or it is using try-catch blocks instead of if-else blocks...)

These are special cases with parallelized loops. Remember that you cannot throw from inside an OpenMP loop. This is undefined behaviour. Exceptions must be caught inside the loops, the error messages are accumulated into an ArrayOfString and the loop is exited gracefully. After the loop, if an error occurred, it is then re-thrown. ARTS_USER_ERROR_IF is not appropriate to replace any of the error handling code inside the loop, only for the re-throw after the loop.

riclarsson:
> I think we need to cover cases where a throw can occur without including an explicit condition attached. We should consider adding an ARTS_USER_ERROR macro that throws without any condition attached. This could then also be used in the catch block of try...catch constructs to re-throw the exception with additional error information, e.g. [like here](https://github.com/atmtools/arts/blob/194939162abe2ca73327126c09b74ab43a31a98f/src/xml_io_compound_types.cc#L1227)

I have added the macro and changed all true-cases to use it.  I considered adding a ARTS_ASSERT_UNREACHABLE, but I decided that that is not as important since an assert is strong enough anyways.

> I also agree that it is too much to do all those replacements as part of this PR. This is better to be done incrementally for different parts of ARTS once this has been merged.

Great.

> > ...
>
> All these might also benefit from an ARTS_USER_ERROR macro.

The first case does.  The other three are unrelated.  I think the only way to deal with the others is to flatten the logic for (2), to use_var_string for (3), and to completely rewrite the error handling for (4).

> > ...
>
> I think avoiding these for now is better than potentially breaking them. Refactoring them should be discussed separately to bring this pull request to an end first.

I have double-checked what I changed here and I don't think the logic is messed up anywhere else now after dealing with these things.

> > tmatrix ...
>
> These are special cases with parallelized loops. Remember that you cannot throw from inside an OpenMP loop. This is undefined behaviour. Exceptions must be caught inside the loops, the error messages are accumulated into an ArrayOfString and the loop is exited gracefully. After the loop, if an error occurred, it is then re-thrown. ARTS_USER_ERROR_IF is not appropriate to replace any of the error handling code inside the loop, only for the re-throw after the loop.

Two things.

1)  So long as the logic is sound already, then this PR cannot mess up with OpenMP protections.  If a throw is acceptable, then the catch will fill up an ArrayOfString as I see it.  I am never re-throwing in a catch if there was no rethrow already there.  So the new error system is appropriate, so long as the inner logic throws runtime error and the outer loop catches the throw.
2)  TMatrix logic is especially weird here.  It seems to me that TMatrix interface returns an error code, even if that error code is a char-array (it shouldn't have to be, since the true error comes from user input, so it should be checked elsewhere).  That error code is thrown in the interface function instead of returned as a simple EXIT_FAILURE/EXIT_SUCCESS or the equivalent.  The wrapper function in ARTS is then using try-catch block to act as the if-else logic of what essentially the error code return could be doing.  Not only that, regardless of if there is an error or not, an ostringstream is filled with a warning about the potential error ahead of time.  This is a very convoluted way to work with the logic of the errors, making it difficult to adopt this code.  In these kinds of weird logic, I am unable to tell how to handle the adaptation to the new system.

## One last thing:

Is it more appropriate to use `logic_error` instead of `runtime_error` for `ARTS_ASSERT`?  Might it mess with some wrappers that require runtime errors?  It would help mostly for the aesthetics of things, but I am also imagining writing a CPP-API auto-generated "test" function that calls all functions of ARTS with standard inputs.  If the error is of type `runtime_error` such a test-function could accept the throw, whereas if the error is of type `logical_error`, then it knows that there is at least something wrong with the function.  This would of course not fix all errors, but at least it could potentially catch a few.

olemke:
> Two things.
> 
> 1. So long as the logic is sound already, then this PR cannot mess up with OpenMP protections.  If a throw is acceptable, then the catch will fill up an ArrayOfString as I see it.  I am never re-throwing in a catch if there was no rethrow already there.  So the new error system is appropriate, so long as the inner logic throws runtime error and the outer loop catches the throw.

Makes sense.

> 2. TMatrix logic is especially weird here.  It seems to me that TMatrix interface returns an error code, even if that error code is a char-array (it shouldn't have to be, since the true error comes from user input, so it should be checked elsewhere). That error code is thrown in the interface function instead of returned as a simple EXIT_FAILURE/EXIT_SUCCESS or the equivalent. 

The problem is that we're relying on 3rd party FORTRAN code here. The TMatrix algorithm can fail to converge depending on varying combinations of input values which is not always predictable. Hence we cannot check that beforehand.
On the EXIT_FAILURE/EXIT_SUCCESS: It is quite helpful to the user that the FORTRAN code actually returns a string hinting at what caused the error.

> The wrapper function in ARTS is then using try-catch block to act as the if-else logic of what essentially the error code return could be doing.  Not only that, regardless of if there is an error or not, an ostringstream is filled with a warning about the potential error ahead of time.  This is a very convoluted way to work with the logic of the errors, making it difficult to adopt this code.  In these kinds of weird logic, I am unable to tell how to handle the adaptation to the new system.

Since the OpenMP parallelization of the loop has been removed at some point in the past because the FORTRAN code is not threadsafe anyway, the logic can be simplified now. I'll submit a separate PR to clean that up.

> 
> ## One last thing:
> Is it more appropriate to use `logic_error` instead of `runtime_error` for `ARTS_ASSERT`? Might it mess with some wrappers that require runtime errors? It would help mostly for the aesthetics of things, but I am also imagining writing a CPP-API auto-generated "test" function that calls all functions of ARTS with standard inputs. If the error is of type `runtime_error` such a test-function could accept the throw, whereas if the error is of type `logical_error`, then it knows that there is at least something wrong with the function. This would of course not fix all errors, but at least it could potentially catch a few.

Throwing a `logic_error` would indeed require every function up the call chain that catches `runtime_errors` to also have an additional catch block for the `logic_error` to avoid uncaught program termination. I don't think `logic_error` is necessarily a good fit for ARTS_ASSERT in general. The description on the (C++ Reference) for logic_error reads:

*This class defines the type of objects thrown as exceptions to report errors in the internal logical of the program, such as violation of logical preconditions or class invariants.*

Maybe an `ArtsAssertion` class that inherits from `std::runtime_error` would be more fitting.

I see your point of being able to distinguish between asserts and user errors. But before continuing down that road, let's first finish up this PR.

riclarsson:
I will leave the dealing with tmatrix errors up to you then.  It just really sticks out as a weird example.  The new error handling isn't that fancy, it should be able to be adopted to most use cases, so if it isn't then I am suspecting the original error handling as convoluted.  (This might be a bad assumption, but I think it works for now.)

> Throwing a `logic_error` would indeed require every function up the call chain that catches `runtime_errors` to also have an additional catch block for the `logic_error` to avoid uncaught program termination. I don't think `logic_error` is necessarily a good fit for ARTS_ASSERT in general. The description on the (C++ Reference) for logic_error reads:
> 
> _This class defines the type of objects thrown as exceptions to report errors in the internal logical of the program, such as violation of logical preconditions or class invariants._
> 
> Maybe an `ArtsAssertion` class that inherits from `std::runtime_error` would be more fitting.

That would work.  However, I would also like to argue that if a user ever reaches an `ARTS_ASSERT`, then the internal logic of the program is wrong.  They should have been stopped by a user error, not an assertion error.  So I don't see the issue.  Other than what you rightly say that the functions above all have to catch the error properly, I don't think it is against the spirit of things.  Though I do see people claim that the existence of "logical_error" is wrong to begin with, so perhaps that's another reason not to use it.

> 
> I see your point of being able to distinguish between asserts and user errors. But before continuing down that road, let's first finish up this PR.

OK, what more do you want me to fix before then?  I think things are caught up now.  No pun intended.

olemke (approved):
Looks all good. Great work!

PR #285, opened 2020-12-20
Some bug and documentation fixes

erikssonpatrick:
Stuff I found when setting up some new calculations. Should also get into v2.4. As far as I remember no previous changes in affected files that should cause problems to merge with v2.4.


olemke:
I'll backport these to 2.4 post-holidays in January.

PR #288, opened 2021-01-15
Use ENUMCLASS

riclarsson:
This is a small commit to use ENUMCLASS in a few more places.  Since I also found string_view useful and some of the old classes can have quite a few more constexpr functions I decided to review them as well.

Note that old binary catalogs might not load anymore.  I am not sure what the policy should be about this?  On one hand, the exact binary layout cannot be guaranteed from one user to another, so we cannot distribute binary data. On the other hand it is nice to distribute some data binary.

### DIFF
--- a/controlfiles-python/raw/corr.py
+++ b/controlfiles-python/raw/corr.py
@@ -60,9 +60,3 @@ ws.ybatchTimeAveraging(time_step="24 h", disregard_first=1, disregard_last=1)
 ncfn = os.path.splitext(__file__)[0]+'_ref.nc'
 # pyarts.classes.netcdf.save(ncfn, [ws.covmat_sepsbatch, ws.ybatch])
 refs = pyarts.classes.netcdf.load(ncfn)
-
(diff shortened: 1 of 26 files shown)


olemke:
> Note that old binary catalogs might not load anymore. I am not sure what the policy should be about this?

Right now, I don't think that's an issue as it is still in an experimental stage for catalogs at this point.

> On one hand, the exact binary layout cannot be guaranteed from one user to another, so we cannot distribute binary data. On the other hand it is nice to distribute some data binary.

Are they not written through `bofstream` to ensure cross-platform binary compatibility as done for the other data types? Or is there a different issue?

riclarsson:
They are written through `bofstream`.  But as far as I know, we should support `float` as `Numeric`, since otherwise there would be no point in having `Numeric`.  So I will just presume that distributing binary data is not a good idea unless you control the build environment.

olemke:
Numeric values are converted to double by `bofstream` and converted back to the selected Numeric type by `bifstream`, thus keeping the output files compatible independent of the build setting.

PR #295, opened 2021-01-26
Additional issue templates

olemke:
This PR adds new issue templates for documentation, feature and enhancement related topics.
You can look at the new templates in action on my fork:

https://github.com/olemke/arts/issues/new/choose

Since we have to separate labels defined for "feature" and "enhancement", I've also added two separate templates to cover those cases.

Any comments / feedback / improvement ideas on these templates?

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,6 +1,6 @@
 ---
 name: Bug report
-about: Create a report to help us improve
+about: Create a report to help us improve ARTS
 title: ''
 labels: bug
 assignees: ''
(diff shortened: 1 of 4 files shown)


riclarsson:
I think a feature is not added because of frustration but because of some practical use case.  Like:

"I am trying to solve problem X.  To do that in ARTS, I have to first run function Y to prepare my data for variable Z.  I think adding a feature similar to function Y would be useful."

olemke:
@riclarsson Both very good points. The original Github issue template I used doesn't distinguish between feature and enhancement issues and I missed to remove that example from our feature template.
Also added a bullet point regarding to the level of commitment to implement the new feature.

stefanbuehler:
Looks great to me! Perhaps add the point about level of commitment also for the other types of request. Good to make the point that help is welcome.

PR #299, opened 2021-02-08
Don't check Sleep timer against upper bound

olemke:
Removed the upper bound test for arts.pyarts.fast.classes.TestTime which
was randomly failing depending on system load and other factors.

From https://www.cplusplus.com/reference/thread/this_thread/sleep_for/ :

"The execution of the current thread is stopped until at least rel_time
has passed from now."


riclarsson:
My mistake, l like the fix.

riclarsson (approved):

PR #301, opened 2021-02-09
Backport continua fixes

olemke:
Cherry-picked from #300


erikssonpatrick:
Hi,

Maybe I should have clarified that I have been emailing on the side about this issue with Stefan and Richard. In short, I found NaNs in my simulations. Richard tracked down the reason. Checked my proposed solution with Stefan. 

Anything else to discuss?

P

erikssonpatrick:
Hi again,

Now see that this just referred to v2.4. So I assume all OK,

P

olemke:
All good. This is just the backport to get the fix into the stable ARTS version.

PR #302, opened 2021-02-09
This is a mixed bag of things to close of ARTS 3 for my part (according to issue tracker)

riclarsson:
This introduces some basic raw functionality.  `level0_data` and `level0_time` are made workspace variables.  The only real way to use them are via `ybatchCAHA`.  If the latter is set, it outputs sensor_time, otherwise it uses the `GIN` of hot and cold temperatures to perform textbook calibration of a CAHA-style system.  I also introduced a few other simple manipulative functions in `raw.cc/raw.h`.  The basic functionality is: calibrate (`ybatchCAHA`), time averaging (`ybatchTimeAveraging`), correction (`ybatchTroposphericCorrectionNaiveMedianForward`), and focus (`ybatchDoublingMeanFocus`).  These are enough to simplify and use most sensor ground data that I am generating.  There is in addition a masking functionality that I use in my own setups (`ybatchMaskOutsideMedianRange`) to remove some bad data points by `NaN`-ifying them.  All of the methods has statistics that operates on "`NaNs` and `Infs` are ignored", to allow bad data to be masked in that way.  Note that this makes it important which order you do your operations, as masking is easier after tropospheric correction ("I will have a line strength around 2 K above a 3 K baseline, so values below 0 or above 6 K are probably from a bad ADC and not just noise").  There's an example of `test_raw` built if you enable the GUI.  It shows each step pedagogically (though with terminal text; I am considering how to allow setting titles and so on in the debug plot).

Other changes in this commit:
sensor_time is now an `ArrayOfTime`.  I have made it so you can set it with standard `Vector` functions where it was applicable in controlfiles.  There are internal support functions, both called `time_vector`, that turns an `ArrayOfTime` into a Vector or vice versa based on Unix-time.  This is perhaps not optimal and it would be better to simply use the individual time stamps without conversion to Vector, but it allowed a very simple translation of the original functions using `sensor_time` to happen.  Note that this `ArrayOfTime` is not strictly following C++17 standard (it's technically C++20 to demand Unix time), but I understand it's fine in Clang and GCC on both Linux and MacOS.  If there are problems with this, I can probably work around them.

I decided to flatten `nlte_source` and make `dnlte_source_dx` a single output of `propmat_clearsky` agenda.  I also changed the internals of `propmat_clearsky` agenda to operate on `jacobian_quantities` instead of on an internal representation of it that fits with how the calculations are done there.  This might slow down some code and speed up others.  It means less allocations, but of very small `Arrays`, whereas it might mean more allocations of other stuff.  I tried to compensate by changing `get_stepwise_clearsky_propmat` to actually use `jacobian_do`, which it didn't do so it computed things needlessly over-and-over.  The last function also because of these changes is now only allocating a temporary `propmat_clearsky` as output of the `propmat_clearsky_agendaExecute` call.  The rest are now used as allocated by the function calling `get_stepwise_clearsky_propmat`.  Note that I did not dare change the complicated logic around what and where things are overwritten, so it remains but some is empty now.  Also note that this change means that NLTE will never support LTE species-tag special derivatives, since `nlte_source` is a purely calculated.  I am thinking of changing `propmat_clearsky` agenda once more after this to also flatten `propmat_clearsky`, so that the species-tag derivatives are set properly in the agenda itself.  This will take significant more efforts though, as I want to refine the matching-properties a bit.  For future development here, make use of something like `if (not propmattype_index(derivatives_data, iq)) continue;` when you are developing `propmat_clearsky` agenda related stuff.  This will skip things you don't need to compute.

Since I was complaining about unclear options handling in ARTS for user errors, I also made an `Options` namespace in `constants.h`.  This I will use henceforth for user-facing options together with the generated `to##ENUMTYPE##OrThrow`.  Enums now also keep a simple `std::array` of all enums in and about since I needed to loop over all options to check them in one place.  I hope others will use this way to deal with options but I of course know none will.

The plotting debug-tools are slightly refined.  I just squashed them a bit to allow easier plotting of some basic data types.  I might still refine them more as I find them very useful for inspecting temporary data.  I note that they don't work very well in multithreading for instance.  I also had to reorder the build a bit.  I don't know what happened but I couldn't link with the XML-writing as it was...

I removed `abs_coef` and company as outputs.  These have not made sense to be user-facing for a long time already.  No controlfiles were using them.  This brought with a few documentation changes.

I made a simple way to allow creating an `Array<X>` from a `std::array<Y>` so long as `Y` is convertible to `X`.  Before, `X` and `Y` had to be the exact same type, but in my mind it makes sense to allow this as long as they can be created from one-another.  This makes for a bit more dangerous error messages about weird constructors, so I made the constructor print a static error message.

I made several of the propagation matrix to `Eigen` functions templated to allow more sensible ways of calling them.

These methods are changed:
`AbsInputFromRteScalars` (removed)
`abs_coefCalcFromXsec` (removed)
`yMaskOutsideMedianRange`
`ybatchMaskOutsideMedianRange`
`yDoublingMeanFocus`
`ybatchDoublingMeanFocus`
`nlte_sourceFromTemperatureAndSrcCoefPerSpecies` (removed)
`propmat_clearskyAddFromAbsCoefPerSpecies` (removed)
`ArrayOfTimeNLinSpace`
`ArrayOfTimeSetConstant`
`ybatchCAHA`

These workspace are changed:
`abs_coef` (removed)
`abs_coef_per_species` (removed)
`counts` (removed)
`covmat_sepsbatch` (removed)
`dsrc_coef_dx` (removed)
`dnlte_dx_source` (removed)
`nlte_dsource_dx` (removed)
`dnlte_source_dx`
`level0_data`
`level0_time`
`nlte_source_field` (removed)
`src_coef` (removed)
`src_coef_per_species` (removed)


olemke:
Two tests in check-all are failing (they work in master):

```
➜ ctest -E python -R arts.ctlfile.slow.artscomponents.wfuns.TestWfuns
Test project /Users/olemke/Hacking/uhh/arts/build
    Start 115: arts.ctlfile.slow.artscomponents.wfuns.TestWfuns
1/1 Test #115: arts.ctlfile.slow.artscomponents.wfuns.TestWfuns ...***Failed   13.62 sec
O3 rel analytical
O3 rel perturbation
Run-time error in controlfile: /Users/olemke/Hacking/uhh/arts/controlfiles/artscomponents/wfuns/TestWfuns.arts
Run-time error in method: Compare
Ja-Jp FAILED!
Max allowed deviation set to : 0.005
but the matrices deviate with: -2.95812
```

```
➜ ctest -E python -R arts.ctlfile.xmldata.artscomponents.cia.TestCIA
Test project /Users/olemke/Hacking/uhh/arts/build
    Start 165: arts.ctlfile.xmldata.artscomponents.cia.TestCIA
1/1 Test #165: arts.ctlfile.xmldata.artscomponents.cia.TestCIA ...***Failed    3.16 sec
Run-time error in controlfile: /Users/olemke/Hacking/uhh/arts/controlfiles/artscomponents/cia/TestCIA.arts
Run-time error in method: Compare

abs_xsec_per_species[10] (100,1) and abs_xsec_per_species_reference[10] (0,0) do not have the same size.
Mismatch at array index: 10

abs_xsec_per_species[11] (100,1) and abs_xsec_per_species_reference[11] (0,0) do not have the same size.
Mismatch at array index: 11

abs_xsec_per_species[12] (100,1) and abs_xsec_per_species_reference[12] (0,0) do not have the same size.
Mismatch at array index: 12
```

src/enums.h
```diff
@@ -51,48 +72,87 @@ std::array<std::string_view, size_t(EnumType::FINAL)> enum_strarray(
...
+ \endverbatim
  */
 #define ENUMCLASS(ENUMTYPE, TYPE, ...)                                    \
   enum class ENUMTYPE : TYPE { __VA_ARGS__, FINAL };                      \
                                                                           \
   namespace enumstrs {                                                    \
   constexpr auto ENUMTYPE##Names = enum_strarray<ENUMTYPE>(#__VA_ARGS__); \
+  }                                                                       \
+  namespace enumtyps {                                                    \
+  constexpr auto ENUMTYPE##Types = enum_typarray<ENUMTYPE>();             \
```
olemke, line 155: This definition needs to be preceded by `[[maybe_unused]]` to silence these warnings:

```
/Users/olemke/Hacking/uhh/arts/src/linemixing.cc:120:1: warning: unused variable 'SpecialParamTypes' [-Wunused-const-variable]
ENUMCLASS(SpecialParam, char,
^
/Users/olemke/Hacking/uhh/arts/src/enums.h:155:18: note: expanded from macro 'ENUMCLASS'
  constexpr auto ENUMTYPE##Types = enum_typarray<ENUMTYPE>();             \
                 ^
<scratch space>:26:1: note: expanded from here
SpecialParamTypes
^
1 warning generated.
```

```
/Users/olemke/Hacking/uhh/arts/src/m_absorptionlines.cc:410:1: warning: unused variable 'HitranTypeTypes' [-Wunused-const-variable]
ENUMCLASS(HitranType, char,
^
/Users/olemke/Hacking/uhh/arts/src/enums.h:155:18: note: expanded from macro 'ENUMCLASS'
  constexpr auto ENUMTYPE##Types = enum_typarray<ENUMTYPE>();             \
                 ^
<scratch space>:23:1: note: expanded from here
HitranTypeTypes
^
1 warning generated.
```

erikssonpatrick:
That was a lot. I focus on the level0 part. CAHA? I assume C and H stand for cold and hot but still don't understand the acronym. I suggest going for a longer and then more clear name.
(Don't understand how you keep track of hot, cold, and atmospheric observation with just level0_data and level0_time, but that just a comment, assuming the description of the variables clarifies this.)

riclarsson:
> That was a lot. I focus on the level0 part. CAHA? I assume C and H stand for cold and hot but still don't understand the acronym. I suggest going for a longer and then more clear name.
> (Don't understand how you keep track of hot, cold, and atmospheric observation with just level0_data and level0_time, but that just a comment, assuming the description of the variables clarifies this.)

@erikssonpatrick I agree.  The functions so far takes extra input as GIN.  This is a good way for simplicity.  However, it doesn't map 1-to-1 with the data I read.

Would you rather that this operates in a way where the `level0_data` has all the information it needs?  I can make it a map of some sort of generic type.  I'm not sure how to support proper array-types (size known at compile-time), but I guess it's fine if it only supports dynamic types?

riclarsson:
> That was a lot. I focus on the level0 part. CAHA? I assume C and H stand for cold and hot but still don't understand the acronym. I suggest going for a longer and then more clear name.
> (Don't understand how you keep track of hot, cold, and atmospheric observation with just level0_data and level0_time, but that just a comment, assuming the description of the variables clarifies this.)

Ah, and CAHA is cold-atm-hot-atm repeated until the instrument crashes.  I'll make the name longer.

erikssonpatrick:
About the GIN part. I was just surprised about the few variables involved. I can not say anything about GIN or not without getting into the details and no time for that. So keep it as it is for now.

riclarsson:
Hmm, it's enough I think to have this few variables.  After this I just need to fit a baseline, but that's the OEM's responsibility in theory (and practice for my setup).

I might make a little bit of a more generic data type anyways.  Our measurements are either in "ints" or in "float", by the bit counts depending on sensor, so I think I need some variable that can be a vector of either of these anyways or the memory footprint becomes too large to handle many days of data (a single day here is about 16 Gb of data with double, and if I am to make time-series of this I will probably want it to 'overlap' by a few hours in the morning and night to confirm it's not too bad a fit).  Then smaller original data would be useful.

riclarsson:
@olemke I fixed both cases, I think.  As well as adding a maybe unused attribute.  I never saw that warning, so you must check first if this actually removes it.  I also changed the name to be clear about the cyclicity of the data in accordance with Patrick's suggestion.

riclarsson:
This now also closes #303

olemke:
> @olemke I fixed both cases, I think. As well as adding a maybe unused attribute. I never saw that warning, so you must check first if this actually removes it.

Looks all good now. 👍

PR #308, opened 2021-02-15
Make Jacobian target store keys in usable manners

riclarsson:
Also add a few more Options to constants.h for dealing with Jacobian.
Some of these had spaces in them before.  These are now removed since
enums cannot have spaces in their names.  Would it be better to allow
spaces again?  I think it's fine not to, and it would require a few
more steps to allow this.


olemke:
@riclarsson About the changes in CMakeLists.txt: I happened to discover the cmake "option" macro last week and was about to submit a patch similar to what you did in this PR. I added one commit from my branch to this PR that shows the options for disabling Fortran modules only if Fortran support is turned on.

olemke:
No strong opinion on the spaces. I'm fine with not allowing them if it keeps the ARTS enum implementation code cleaner. Only downside is that it breaks old controlfiles once more. But the changed parameters names where probably not widely used anyway yet. If no one else has objections I'm happy with it.

riclarsson:
@olemke Great minds and all that.  I also found out about this "option" last week.  I didn't search more and just added the arguments because auto-complete was nice.  Your version is better.

I also think we should be allowed to make breaking changes from now until version 3.  Isn't the entire point of v.2.5 that we are cleaning up things added for features in 2.3 that we now consider we can improve?  Keeping spaces in options might be a hard requirement though, because perhaps not all options are as clean as the two I just broke, and it would be good to just bite that apple now.  (There's probably a bunch of MIT-licensed algorithms available that matches strings ignoring whitespaces, so I don't believe it's difficult to add such a thing here.  Though I also prefer if options are kept simple without spaces since it forces simple names.)

PR #309, opened 2021-02-16
Make Time easier to interact with

riclarsson:
Also added a minimizer for wave-like features (which seems to work
according to the test I did in test_linalg.  Maybe the test should
be moved out of there and into its own test-style however...)

### DIFF
--- a/python/pyarts/classes/Time.py
+++ b/python/pyarts/classes/Time.py
@@ -3,6 +3,7 @@ from pyarts.workspace.api import arts_api as lib
 
 from pyarts.classes.io import correct_save_arguments, correct_read_arguments
 from pyarts.classes.ArrayBase import array_base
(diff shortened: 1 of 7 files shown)


riclarsson:
Question for another PR:

Is there any way to make the `pyarts.workspace.Workspace::variable::value` method return a `pyarts.classes.from_workspace(pyarts.workspace.Workspace::variable)` instead of throwing when the type is not explicitly available in the `Workspace`?

The `classes` supports most of the other stuff but perhaps not all.  At least for simply things like assignments, file IO, simple math, and these kinds of things it should be good enough to return a `from_workspace` variable.

olemke:
Since your question needs further discussion and I'm afraid it gets lost in here after closing this pull request, please open a separate issue for that.

PR #310, opened 2021-02-17
Add native classes support to .value

olemke:
When accessing WSV with .value that don't have a native Python class implementation, fall back to use Richard's native classes to pass the value from ARTS to Python, e.g. ws.time.value works now.

### DIFF
--- a/python/pyarts/workspace/variables.py
+++ b/python/pyarts/workspace/variables.py
@@ -411,6 +411,8 @@ class WorkspaceVariable:
             workspace.
         """
         from pyarts.xml import load
+        from pyarts.types import classes as arts_classes
+        from pyarts import classes as native_classes
 
         if not self.ws:
             raise Exception("Cannot retrieve the value of a variable without "
@@ -418,7 +420,12 @@ class WorkspaceVariable:
         with tempfile.TemporaryDirectory() as tmpdir:
             tfile = os.path.join(tmpdir, 'wsv.xml')
             self.ws.WriteXML("binary", self, tfile)
-            v = load(tfile)
+            if self.group in arts_classes:
+                v = load(tfile)
+            else:
+                cls = getattr(native_classes, self.group)
+                v = cls()
+                v.readxml(tfile)
 
         return v
 
--- a/python/pyarts/workspace/variables.py
+++ b/python/pyarts/workspace/variables.py
@@ -305,6 +305,7 @@ class WorkspaceVariable:
 
         """
         from pyarts.types import classes as arts_classes
+        from pyarts import classes as native_classes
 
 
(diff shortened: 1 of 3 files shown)


riclarsson (changes requested):
I think that `save` and `load` XML requires changes to make this work.  The former is easy to just wrap the `savexml` method that all `native_classes` should have.  The latter is less clear.  You can wrap `readxml` but that requires introspection of the XML file first (meaning that the loads might slow down significantly).  But I guess the latter logic can be handled in a try-catch block?

The change in `variables.py` I think is slightly wrong.  As far as I understand it, types like `Vector` maps the data directly.  So it should be fine to modify `::value` instead of `::to_arts`.  I don't think a copy is necessary or desired here.  Instead, I would add something that directly uses `from_workspace` if the type is in `native_classes` before the try-catch block that calls `to_arts`.  This would allow use to simply not copy any values at all.  (I assume here that the check for if the variable is initiated is already done.)

olemke:
Good point about the copying, it's corrected now. I'm a bit reluctant to use the ARTS classes over the ones that have currently still Python implementations. For example, the GriddedField classes are widely used and switching to the ARTS class would change the interface significantly and also remove some additional functionality that is present in the Python class.

This PR was solely meant to solve the issue of using `.value` on WSVs which didn't supported it yet. The save/load discussion can be addressed in #311

riclarsson:
@olemke Ok, I can understand about not changing the ones with working classes.  Still, what is then your opinion about removing classes that have half-bad implementations in python?  Most of the classes in catalogues.py are just simple value holders (which the new classes handle quite well) or entire re-implementations of the classes as they were back when the code was implemented.

I like the PR at this stage.  It seems to address my concerns about `::value`.  We leave file IO for another PR then.

olemke:
> @olemke Ok, I can understand about not changing the ones with working classes. Still, what is then your opinion about removing classes that have half-bad implementations in python? Most of the classes in catalogues.py are just simple value holders (which the new classes handle quite well) or entire re-implementations of the classes as they were back when the code was implemented.

For those I'd strongly vote to get rid of them.

riclarsson (approved):

simonpf:
I would like to look at this but won't manager before tonight.

olemke:
No worries, take your time.

simonpf (changes requested):
This looks good, but should definitely covered by a unit test :).

olemke:
@simonpf  You're absolutely right. :-) I've added a `.value` test for the Time class. Not sure if you had something else in mind. Adding a test for all types from `classes` would be a bit much.

simonpf (approved):

PR #312, opened 2021-02-17
Remove value-holder classes from python

riclarsson:
These are better served by the pyarts.classes functionality.

This is in response to: https://github.com/atmtools/arts/pull/310#issuecomment-780540484


riclarsson:
This removes the majority of pyarts's pure python classes.  The criteria for removal was: 1) Has changed in ARTS.  2) Only holds values.  3)  Is no longer in ARTS.  4)  Was a support class.  5)  Things that I cannot imagine working with the python class for.

Basically, the grid-field is kept because it has a few additions lacking in the C API class.  Sparse and CovarianceMatrix I think is better as pure scipy types or via the C API but I am unsure how they are used so I won't touch them.  The scattering data, I don't want to touch it at all so I won't.

Note that this change does not seem to trigger any tests, so it should be fine to remove the functionality since no one cares enough to check if it works.

olemke (approved):
Looks good to me. A lot of old cruft gone.

Since this PR depends on #310 (otherwise the unit test `test_generic_input` involving SpeciesTag fails), it can only be merged after #310.

simonpf (approved):
I don't think I have ever used anything from these files, so OK from me.

PR #314, opened 2021-02-19
Turn off braced-scalar-init clang warning

olemke:
Warning is turned off for parts in autogenerated code where initializer
lists contain single numbers in double braces.


riclarsson (approved):
Looks good.  Will be useful even when/if switching to modules.

PR #316, opened 2021-02-26
Add xml load/save support for ARTS C++ classes

olemke:
Add compatibility to the xml load/save PyARTS functions for ARTS
classes that are only available through the C++ API.

load first tries to read the file with the Python interface.
If that fails, it uses the C++ classes interface.

save checks if the variable contains a savexml function which all the C++
ARTS classes do. If not, it falls back to the Python XML writer.

### DIFF
--- a/python/pyarts/classes/__init__.py
+++ b/python/pyarts/classes/__init__.py
@@ -47,7 +47,7 @@ from pyarts.classes.Timer import Timer
 from pyarts.classes.TransmissionMatrix import TransmissionMatrix, ArrayOfTransmissionMatrix, ArrayOfArrayOfTransmissionMatrix
 from pyarts.classes.Vector import Vector, ArrayOfVector, ArrayOfArrayOfVector
 from pyarts.classes.Verbosity import Verbosity
-from pyarts.classes.XsecRecord import ArrayOfXsecRecord
+from pyarts.classes.XsecRecord import XsecRecord, ArrayOfXsecRecord
 
 
 # Attempt at conversions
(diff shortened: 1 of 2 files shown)


riclarsson:
I think it looks good.  A bit complicated but that's sort of necessary.

The save-function is clear and I don't think it will fail for any of the classes.

I think I added safe-guards for all loads to raise errors when they can.  So that bit should also be fine.  You get a whole lot worse-to-understand error than with python failing though.

Maybe we should print to `std::cerr` for the IO-wrappers?  I'm sure `spyder` is hiding even the `cerr`-stream though, so it might not help everyone.  It's also not very ARTS-y to print the `cerr`.  It is a two-liner in the `BasicInputOutputCAPI`-macro to make this change.

olemke:
I don't fancy outputting directly to `std::err` because it takes away easy control of what is printed on screen from Python.

A better approach to dealing with error messages was implemented by @simonpf in the workspace API. The error output is [caught on the C++ side](https://github.com/atmtools/arts/blob/59f23caf12bcad6f5d6830e283580eb7c7b6afe1/src/interactive_workspace.cc#L157) and then [repackaged into a Python exception](https://github.com/atmtools/arts/blob/59f23caf12bcad6f5d6830e283580eb7c7b6afe1/python/pyarts/workspace/methods.py#L445). This gives the caller on the Python side the option to retain full control on dealing with the error output.

But refactoring the [`xmlsave##TYPE`](https://github.com/atmtools/arts/blob/59f23caf12bcad6f5d6830e283580eb7c7b6afe1/src/arts_api_classes.cc#L132) macro and all classes in `pyarts/classes/` would exceed the scope of this PR.

simonpf (approved):

PR #320, opened 2021-03-08
My work for v3

erikssonpatrick:
1. Some improvement of documentation.
2. Improved radar onion peeling method. Would be nice to parallelize it and Oliver will look at it.


erikssonpatrick:
Renamed the PR.

Tested the parallelization and worked just fine. Thanks.

After learning a bit more about CloudSat's observations, made a small change to the input. Also made the change in TestOnionPeeling.arts that we discussed. 

Don't plan any further changes. OK to PR from me.

PR #321, opened 2021-03-08
Add SurfaceBlackbody method

Smpljack:
Add SurfaceBlackbody method, which calculates surface emission and surface emission jacobian for a blackbody surface. 
I use this method for retrieving the surface temperature in the infrared (IASI).
Implementation is based on SurfaceTessem and SurfaceFastem.


olemke:
Looks good. It would be nice to have a simple test case that actually calls the function and compares results against reference values. It could be either an ARTS controlfile or a Python script. Would be great if you could add one.

erikssonpatrick:
Got 5 min free and took a look as I was curious. Great that you calculate the derivative with dplanck_dt, as it gives an analytical value. But as far as I can see, the code starting on line 2707: 
 const Numeric dd = 0.01;
      Matrix surface_los2;
      surfaceBlackbody(surface_los2,
                       dsurface_rmatrix_dx[irq],
                       dsurface_emission_dx[irq],
                       atmosphere_dim,
                       f_grid,
                       stokes_dim,
                       rtp_pos,
                       rtp_los,
                       skin_t[0] + dd,
                       verbosity);
has no effect and can be removed. 

If you have the energy, before you remove the code, check that
(dsurface_emission_dx[irq]-surface_emission)/dd 
is close to what dplanck_dt gives. If not we have a problem somewhere.

Smpljack:
Thanks for taking a look, Patrick, good catch! The lines you pointed out were indeed of no purpose anymore, but just a leftover of the previous numerical calculation. I did the comparison of analytical and numerical `dsurface_emission_dx` calculations for the testcase of SurfaceBlackbody that is part of this PR (includes 10 frequencies at 183 GHz). The methods are very close, with absolute values of both on order 1e-17 and the absolute differences of `numerical-analytical' on order 1e-26, being positive for all frequencies.

PR #322, opened 2021-03-09
Update hitran online and all hitran IO routines to use different

riclarsson:
default for the pressure shift.  Also update lblrtm to do the same

Doing this I ran the reading routine on a newly downloaded version
of the data and noticed that the old version hid errors in the
reading routine.  I had to add several new species to fix this but
I really don't understand where the name of these species goes into
the weird static array of the old reading routines, so I had to make
a new function from hitran's molparam.txt that adds the new species

Note that the code makes use of several static variables.  If this is undesirable, then the statics can be removed from the "hitran_species.cc" file.  I don't know if it actually makes things faster to just declare these once but the old code used static tables so it should be fine to do the same here.

### DIFF
--- a/python/pyarts/__init__.py
+++ b/python/pyarts/__init__.py
@@ -8,6 +8,7 @@ from pyarts import xml  # noqa
 from pyarts import classes  # noqa
 from pyarts import workspace  # noqa
 from pyarts.common import *  # noqa
+from pyarts import hitran  # noqa
 
 __all__ = [s for s in dir() if not s.startswith('_')]
 __version__ = "@ARTS_VERSION@"
--- a/python/pyarts/hitran.py
+++ b/python/pyarts/hitran.py
@@ -1,14 +1,24 @@
 from urllib.request import urlopen
 
+# Map Hitran to ARTS species names
+_HITRAN_TO_ARTS_NAMES = {
+    "CH3CN-2124": "CH3CN-211124",
+    "CO2-827": "CO2-728",
+    "H2CO-126": "H2CO-1126",
+    "H2CO-128": "H2CO-1128",
+    "H2CO-136": "H2CO-1136",
+    "HC3N-1224": "HC3N-12224",
+    "HCOOH-126": "HCOOH-1261",
+}
+
 
 def gen_latest_molparam_map(molparam_txt_file=None):
     """ Generates a version of latest_molparam_map used in hitran_species.cc
 
     The variable is simply printed to stream with print() as the intent is
-    to use this output in ARTS directly.  Note, to keep this simple, however,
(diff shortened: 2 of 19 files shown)


riclarsson:
The logic of Hitran is independent of Arts.  If they choose to change their logic we obviously shouldn't change ours in the slightest.

It makes way more sense to have a separate version of my `from_mol_iso` for every version of Hitran we wish to support than it does to hide errors like `hiso` is doing or pollute `species_data.cc` with compile-time logic.

In addition, it is much easier to maintain something like this function than to keep introducing compile-time flags to `species_data.cc`.  Hitran can only really do three things here without messing up things entirerly:
1) Add a new molecule.
2) Add a new isotopologue.
3) Change isotopologue order.

The first and second are trivial to fix and you will get a very clear error on how to deal with this from my method.

The last is also trivial to maintain but requires work.  You make versioned copies of the entire function.  You make `from_mol_iso` be accessed when no key is given to the wrapper function.  You then use the `ReadHITRAN` `hitran_type` to choose which wrapped function you call.

The way to maintain the first and second option is exactly the same if all this logic goes into `species_data.cc`.  I am less clear how clean the error will be here but I assume it can be made clear.

The method to maintain the latter is probably to make the compile time option `USE_HITRAN2008` a time stamp so that the `#ifdef USE_HITRAN2008` of CO2 reads `#if HITRAN_VERSION < 20091231` (or whatever date you wish to put here).  You then keep chain-adding these versions to all other species that HITRAN messes with the order of.  You make this user-defined by having a `#ifdef CUSTOM_HITRAN_VERSION` that sets the `HITRAN_VERSION` manually in `species_data.cc` to itself if defined or just be defined to the latest Hitran version manually otherwise.

I simply don't see how the latter is easier to maintain.  The former also solves the user interaction cleanly.  Just give the version of Hitran you wish to use.

riclarsson:
Note: LBLRTM might have to be checked here.  I am not sure at all which Hitran version it bases its molecule numbers on.  Or if it even have these species.

olemke:
Thanks for implementing the proposed changes. I've pushed two commits with minor changes to the `hitran.py` script: Optionally downloading `molparam.txt` directly from the Hitran web page and doing the mapping from AFGL names to ARTS species automatically.

riclarsson:
I like your changes.  Is there anything else missing to merge this?

I mean, I personally would like to have all of species_data.cc as a constexpr lookup table but I guess that isn't going to happen here.  (I have something like that in my playtime-RT code in my radctrl project so I know it is possible, but that solution is not for ARTS just for playtime)

olemke:
I agree, `species_data` seems like a viable candidate to be made `constexpr`.

PR #327, opened 2021-03-20
Introduce new LBL code

riclarsson:
This code does several things.  First, it introduces a new LBL routine.  This new routine basically wraps a few classes that deals with LBL into variant classes that does the computations at the end.  It ends up speeding up compared to old code by about a factor 2 for many Jacobian.  It pessimizes by some 10% for few Jacobian.  And it improves the speed for no-Jacobian by 20%.  (Numbers are from after removing the troubling OMP commands of the old code.)

I don't think using variant is an optimal solution.  A better solution would be to use templates for clear-cut line shapes.  However, current design of ARTS does not allow this.  The main problem is that we have too many options.  For line strength, we have 3 options that matters here.  For line shapes we have 5 options (currently but this can increase).  For Normalization we have 4 options.  So a full set of templates here would end up compiling about 420 versions of the routines (4 norm times 3 strengths times 5 lineshapes times ~6 mirror lineshapes).  Probably 1680 versions is more reasonable since we should want to have LTE code and Jacobian code optimize better if we anyways go with templates.

I think the main problem here is that this is actually not the case.  You don't combine HTP with HTP mirroring.  (Find me a paper that does otherwise!).   Also, you don't smack VVH normalization onto the Doppler line shape.  In fact, when I read about VVH normalization, it does not seem to make much sense using it without a Lorentz line shape mirror.  At least this is what happens in most (all?) of PW Rosenkranz's papers on the subject.  And mirrored line shapes come from the Fourier transform of a dampened wave having a Lorentz factor that is negative, so it makes sense to do it that way.  Anyways, if we combine this down to the number of reasonable line shapes we have today, I would guess we end up with at most some 10-15 specializations at most.  And having 45 templates that would grow to 48 templates if we add one more seems manageable in template form.

So in short:  Would it be better to stop supporting some things here for faster code?

Note:
The new code computes everything in absorption and not x-secs but the comparisons are for just xsec-like calculations so the new stuff is doing some additional work that won't end up hurting users.  Also the new code does the Jacobian stuff a lot more accurate as can  be seen from the added test_lineshapes.cc file.  I changed/added a few methods to deal with this.


riclarsson:
There are quite a few more things in here that I had forgotten about...

This changes the quantum/line matching to be more sane.  This caused quite a few Jacobian that didn't work properly to work (as the line parameters match).

This also changes a few things about how you can interact with the quantum numbers.   Some of this seems to have made clang go crazy.

There's also a very clear slowdown in Zeeman and LBL calculations caused by the SpeciesTag VMR Jacobian now.  The internal variables have to be reset because of this Jacobian.  Without it, a single loop over all absorption is possible.  Would it be acceptable to introduce a dpropmat_clearsky_dxPerturbedSpeciesTagsJacobian method that is used whenever they are added?  They are pure heritage that slow down other code by their very design, so doing something like that seems appropriate.  (The SpeciesTag VMR Jacobian would likely be slower to compute than today but codes like "TestSpectroscopy" improves by some 15% in my tests.)

The last point I want to make is that I would like to change propmat_clearsky to a variable that takes a ComplexMatrix instead of a Tensor4 as its base unit.  The former would drop the Aziumth, Zenith, and Stokes dimensions and instead add a type dimension to the frequency.  There are 5 types of propagation in ARTS today as far as I can see.  These are the normal unpolarized stuff, "Zeeman::Pi", "Zeeman::SigmaPlus", "Zeeman::SigmaMinus", and the Faraday rotation.  The first just adds a single component today to the real part of the final propagation matrix.  The middle three are clearly doing special polarization.  The last is just rotating.  Is there any other type?  Are the Aziumth and Zenith angles used by anyone in a way where they cannot be separated from the Tensor (i.e., not possible to put in Arrays)?  I added these by request but I do not know where they are used.  Anyways, doing this would allow propmat_clearsky to be computed with the same interface throughout, and we would not need to allocate internal memory for at least Zeeman and LBL anymore (if the species tag issue is resolved).

olemke:
The clang error appears to be a compiler bug since it does compile fine on clang-trunk.

Simplified test on Compiler Explorer:
with Clang 11: https://godbolt.org/z/hdhz8o
with Clang trunk: https://godbolt.org/z/zzvfhq

In the meantime, just remove the empty initializer list from [line 143 in `predefined_absorption_models.cc`](https://github.com/atmtools/arts/blob/f25dbd3cae45e84fcfcffc0395fc8d8260dddc29/src/predefined_absorption_models.cc#L143). That avoids the bug and is not needed anyway.

src/lineshape.h
```diff
@@ -91,28 +95,33 @@ class Lorentz {
...
   Complex F;
 
-private:
   Complex dF;
 
-public:
-  Voigt(Numeric F0_noshift, const Output &ls, Numeric DC, Numeric dZ) noexcept;
+  constexpr Voigt(Numeric F0_noshift, const Output &ls, Numeric DC, Numeric dZ) noexcept
+  : mF0(F0_noshift + dZ + ls.D0 + ls.DV),
+  invGD(1.0 / std::abs(DC * mF0)), z(invGD * Complex(-mF0, ls.G0)), F(), dF() {}
```
olemke, line 109: The needs to use `nonstd::abs` as well to compile with libc++. Really annoying that they still haven't made these math functions `constexpr` in the standard. To spot compatibility issues with `libstc++`, I'll make a PR to adapt our clang test build to use `libc++` instead of `libstdc++`.

riclarsson: Fixed now.

riclarsson:
On my machine a slightly modified version of @olemke's test runs at between 0.56-0.6 s for the new LBL method and about 1.1 s for the old LBL method in "Native".  Using "Release" the new method runs at 0.66 s and the old method runs at 0.78-0.87 s.  (The range is from just 2 reruns of the test.)

By mistake, I once did nv=0 in my inner-most "frequency_loop" function.  For both "Native" and "Release" this changes runtime to 0.33 s.  This means that half of the time of LBL our calculations is spent in overhead in the new, faster version.  The support infrastructure calls pretty much the same functions, so I would suspect that the loss is a straight ~0.33 seconds there as well.

olemke:
Is the test you used available in this branch? I'd like to have a look with the profiler.

When using this [Python script](https://gist.github.com/olemke/955cd73bee1275879c2b1c8ca2e10766) to benchmark `propmat_clearskyAddLines` in this branch, calculation time goes down to basically "nothing" with the minimum of 1 frequency (can't do 0 frequencies):

```
➜ python lblbench.py 10000
# Frequencies: 10000
Average run time [s]: 0.456611

➜ python lblbench.py 1
# Frequencies: 1
Average run time [s]: 0.000396
```

riclarsson:
I attach the file.  Modifying the python file is different from what I am talking about.  What I find funny/wrong is that the time it takes to perform these calculations seems to grow with the number of frequency counts regardless of if you perform any calculations or not.

One way to keep growing the number of frequencies while not performing any calculations is effectively done by changing the code:
```cpp
 for (Index iv = 0; iv < nv; iv++) 
```
into:
```cpp
 for (Index iv = 0; iv < 0; iv++) 
```
in the function:
```cpp
void frequency_loop(...
```

I did this by mistake while doing some optimization tests of not using ComplexVector or Vector inside the innermost loop but instead just simple pointer-semantics.  (This optimization works because the code is faster by a little bit.)

This is of course not representative of a real setup but it highlights that more than half of the calculations are happening _before_ the main loop.  Note that commenting out all calls to `frequency_loop` doesn't help here either.  It's still 50% of the work already done.  Commenting out all calls to `cutoff_loop` does however help and the computations is now trivial, i.e., does not grow with frequency count.

[lblbench.zip](https://github.com/atmtools/arts/files/6221752/lblbench.zip)

riclarsson:
There was a very bad search algorithm for finding the upper and lower bounds.  I have switched to using `std::lower_bound` and `std::upper_bound` instead and that pretty much eliminates the search time.  This more than halves computational time in Native.

So now the times one my machine in the test attached above has the following speeds for 5000 frequency counts.

RelWithDebInfo
New: 0.178329 s
Old: 0.475714 s

Release
New: 0.171121 s
Old: 0.425623 s

Native
New: 0.144 s
Old: 0.564 s

riclarsson:
This branch now contains a sparse grid for LBL calculations.  There is nothing obviously wrong in the output when using this sparsification.  There are probably a few bugs here though.  I still need to design more specialized tests to try and identify these.

The algorithm divides the calculation up in 6 separate blocks.  The dense range, the lower sparse range, the upper sparse range, the cutoff frequency, the lower sparse-to-dense neighboring point, and the upper sparse-to-dense neighbor point.  The cutoff is removed from all of these ranges/points.  The sparsification points interpolate to remove some from the dense grid and are then added to the sparse grid.  At the end of computing all lines, a simple linear-interpolation adds the sparse values back onto the dense frequency grid.

Some numbers for now are below for 50'000 frequency grid points where the sparse grid size is 30 GHz and the sparse-to-dense cutoff is 45 GHz.  The general cutoff is at 750 GHz.

RelWithDebInfo:
1.45 s new w/o sparse
0.28 s new w/ sparse
5.26 s old

Native (note that the old method seems to have problems with Native on my machine):
1.24 s new w/o sparse
0.23 s new w/ sparse
5.34 s old

This makes the LBL calculations about 20-25 times faster than before.

olemke:
Really goes to show that we should use STL algorithms wherever possible. I replaced `Linefunctions::find_cutoff_ranges` in the old code with your updated search algorithm for a "fair" comparison. The new version still gives a 2x speedup (5000 frequencies):

RelWithDebInfo
New: 0.114680 s
Old: 0.230885 s

Release
New: 0.112567 s
Old: 0.227431 s

If you're curious where the time is spent in the new implementation, I attach some screenshots for the different loop functions for a profiler run of the `lblbench.py` script. It shows the percentages of the total function time spent on the highlighted lines.

propmat_clearskyAddLines takes 4.63 s for the 10 runs with 5000 frequencies (profiler slows down the calculation a bit).
~2.2 s of that are spent in the Voigt::operator(). Ignore the GCC in the window titles, I had to switch to Clang to get better per-line information. This was done with a8fb80d.

line_loop:
line_loop
cutoff_loop:
cutoff_loop
frequency_loop:
frequency_loop
Voigt::operator():
Voigt_operator

riclarsson:
@olemke Interesting profiles and, yeah, I agree that switching to more <algorithm> and <numeric> would be nice.

Can you please confirm if I am understanding the file-by-file results correctly.

line_loop:
We are not doing any normalization so there's no expensive calls to the setup in its selector.  The calls to the line strength is somewhat expensive since it's doing calculations. The line shape parameter calculations are similarly somewhat expensive, probably as they do call "std::pow" internally.  These 7% are constant regardless of the number of frequency points.  Can I extrapolate this to mean that at about 500 frequencies, these calculations will spend about half their time in just these two functions?

cutoff_loop:
The cutoff-call to `frequency_loop` is only 1.3% if the time here?  It computes only 1 frequency.  At 5000 frequency points and with the 750 GHz cutoff limit we have, there should be about 50 frequency points in the other call to `frequency_loop` from the script.  Yet, it takes 90% of the time to perform the full loop, which is a lot slower than the 2/3 I expect from linear extrapolations of the two calls making exactly the same calls, one just 50 times more often.  Is this the effect of one writing to local data and the other to heap-data?  If so, would we be able to gain back those additional 20% of time if we somehow managed to make the call to just local data?

Still spending 5% of the time in the search function I see.  I don't think this can be done better, and it won't explode or change much by changing the frequency count, so I guess it's fine.  The same with the cut-off removal taking a full 3% of the time.

frequency_loop:
It's spending 10% of its time just adding things up to `F[iv]`?  That's a lot.  I wonder if it would be a good idea to move the line mixing calculations out and into the line strength calculations.  Or perhaps just make it's multiplication happen outside the loop.

Also, 10% on Flm and Sn is not good.  They are doing absolutely nothing in our test scenarios.  Both are literally just returning static constexpr zeroes.  I guess this is a demonstration of the cost of all the "options" we have in ARTS, as we talked about in an email exchange.

Voigt::operator():
This is ideally where we spend most of our time.  The derivatives have a 5% overhead even when unused.  I think this is acceptable, but it could perhaps be also a small improvement to have the class take a poll on if it needs them or not.  I am further thinking that these classes receiving dlsdT on construction would also help quite a lot for this very common derivative.  But I will ignore it for now.

olemke:
line_loop:
Your estimate is pretty much spot on, measured numbers are just a bit lower:
```
500 frequencies:
cutoff_loop: 70.90%
normalizer_selection: 0.18%
linestrength_selection: 8.33%
band.ShapeParameters: 19.41%

5000 frequencies:
cutoff_loop: 92.10%
normalizer_selection: 0.06%
linestrength_selection: 2.22%
band.ShapeParameters: 5.32%

50000 frequencies:
cutoff_loop: 98.91%
normalizer_selection:0.01%
linestrength_selection: 0.26%
band.ShapeParameters: 0.78%
```
cutoff_loop:
For 500 frequencies and the given cutoff, there are not 50 frequencies in the first call, but 72. So the scaling is pretty much  linear:
```
500 frequencies:
frequency_loop 1 (7 frequencies): 47.08%
frequency_loop 2: 8.93%

5000 frequencies:
frequency_loop 1 (72 frequencies): 89.20%
frequency_loop 2: 1.30%

50000 frequencies:
frequency_loop 1 (715 frequencies): 94.87%
frequency_loop 2: 0.18%
```
To address your concern of local-data (stack?) vs. heap access/writing: There is no performance penalty in writing data to the heap vs. stack. Heap memory only has a performance penalty for the allocation, because it has more bookkeeping to do to find and track the memory locations in the fragmented heap space. Since the variables are already allocated outside the loop, it does not have any impact here. I assume you were referring to `F`/`N` and `Fc`/`Nc` in `com` and `cut`?

olemke:
frequency_loop:
Removing the `LM` term from that line reduces the impact of adding to `F[iv]` from 9% to 6.7%. But multiplying it back on would have to be done much further up to not lose what was gained (and more) immediately again.

`Sn` and `Fm` seem to expose the overhead of `std::visit` here for the trivial case.

riclarsson:
Ok, well, I will just mark this down as good-to-know at this point.  I would have to make LM part of IntensityCalculator to use it properly.  This is not trivial as all of the derivatives have to interact well enough.

The main question I have remaining is if it would be better to template the entire thing via some clever usage of `std::get_if<>()` and `nullptr` semantics?  The visit-overhead would be completely gone.  So would probably debug-ability, and I think that the arts-binary would end up a lot larger.  There would probably also be more time for coffee breaks while compiling!

riclarsson:
Last commit adds an interpolation routine for sparse grid that is similar to GRTCode.  I still need to test it a lot more but I think it can live online over the long-weekend.

olemke:
> The main question I have remaining is if it would be better to template the entire thing via some clever usage of `std::get_if<>()` and `nullptr` semantics? The visit-overhead would be completely gone. So would probably debug-ability, and I think that the arts-binary would end up a lot larger. There would probably also be more time for coffee breaks while compiling!

I think with the significant performance improvements you have implemented in this branch and the potential of the multigrid, we should not aim for the template solution. While speed is important, so is maintainability. Your current design makes it easy to understand what's going on in the absorption calculation while neatly hiding away the nitty-gritty details.

riclarsson:
@olemke Then I guess you also agree we don't do anything OpenMP now.  I think then that all that still is required here is to make some tests and doc updates.

olemke:
I think that's fine for now. We still have the OpenMP loops in the higher level functions. Only thing we lost is the OpenMP loop over pressure, right?

riclarsson:
Yup, that's the only thing we lost.  We actually gained better Jacobian, because the old code was broken in parts.  Particularly, the internal VMR derivatives are a lot better now.

riclarsson:
The latest test_lineshape.cc demonstrates the sparsifications.  It really is broken for my linear case so far.  The quadratic test case also has some issues of doubling absorption at the sparse-to-dense boundary.

This code is definitely not ready to merge yet from these results...

riclarsson:
Ok, that was actually easier than I thought.  Turns out the quadratic interpolation doubled one grid point since it was included in both sparse and dense.  Using `std::nextafter(fu, fl)` explicitly excludes `fu` from `upper_bound`, so now it works.

I now think this can be merged.  There may still be problems with the linear sparsity calculator but I don't know how to find those very particular cases so I don't think I should do anything about it.

riclarsson:
There are still some issues with this code but I think it can be merged.

There is one remaining error.  Here's a plot:

dabshist_s45_d30

This is for about 1 cm-1 sparse gridding with 1.5 cm-1 sparse-to-dense cut for the cases we've simulated above.  The vast majority of the quad-interpolation sparse values have less than 1% errors to dense.  Here's the same plot over frequency however:

dabs_s45_d30

As is clear, there's a large additional error for the right-edge of the frequency grid than for the rest.  I cannot figure out where it's from.

Note that to get linear interpolation sparse grid peak down to under a percent, you need to lower the interpolation grid to 1/3 cm-1 or increase the sparse-to-dense grid to 2 cm-1.  This looks like:

dabs_s45_d10
or
dabs_s60_d30

So it can be merged but with an open issue request that the right-edge of the frequency band must get a look-at.  I thereby mark this as ready for review with requests for comments about what meta-information must be improved.  Once merged, I will leave an open issue with a help request to figure out this problem.

olemke:
@riclarsson I did a rebase against master to get rid of some duplicate commits before merging.

olemke:
@riclarsson The description in `methods.cc` for [`propmat_clearskyAddLines`](https://github.com/riclarsson/arts/blob/ac4052cbc1e41817f54a047403c76cc31856c078/src/methods.cc#L12581) seems to be cut off. Is it just a spurious "and" at the end or is there something missing afterwards?

riclarsson:
Hi Oliver, I think this fixes the issues.  There was very little information in the original.  I also added a helper function to compute the sparse frequency grid.

olemke:
Cool, that looks really good now.

PR #328, opened 2021-03-25
Set Release as default build type

olemke:
Set default build type and make the options visible in cmake-gui.
Based on https://blog.kitware.com/cmake-and-the-default-build-type/

Enable ccache by default if available.


CMakeLists.txt
```diff
@@ -26,6 +26,17 @@ if ("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_BINARY_DIR}" OR "${ARTS_SOURCE_DIR}"
...
 
+# Set a default build type if none was specified
+if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
+  set(DEFAULT_BUILD_TYPE "Release")
+  message(STATUS "Setting build type to '${DEFAULT_BUILD_TYPE}' as none was specified.")
+  set(CMAKE_BUILD_TYPE "${DEFAULT_BUILD_TYPE}" CACHE
+      STRING "Choose the type of build." FORCE)
+  # Set the possible values of build type for cmake-gui
+  set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS
+    "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
```
riclarsson, line 37: We still have the Native, which is different from the standard stuff.  It makes just 5-10% faster code (in very specific cases) but nevertheless it should be in our options here.

riclarsson (changes requested):

CMakeLists.txt
olemke: Good point. Thanks for pointing that out.

riclarsson:
As an additional comment:  These days when I start a new CPP project, I tends to use: https://github.com/lefticus/cpp_starter_project as base for CMake (stripping it of most things I don't need).  I like its way of dealing with some of the default options (like ccache being activated by default if available).

olemke:
Yes, it's a good template, I looked at it after he mentioned it in one of the C++ weekly episodes. I'm planning on cleaning up our CMakelists files to adapt them more to modern Cmake standards. Especially not setting global compile options but using the proper target compile options with PRIVATE/INTERFACE etc instead.
For ccache, I'd be fine with enabling it by default if available.

Issue #339, opened 2021-04-16
Adopting lookup table creation to new LBL function propmat_clearskyAddLines 


stefanbuehler:
We want to adopt the absorption lookup table generation to use the new LBL function by @riclarsson . 

Assumption: We change the output variable of abs_xsec_agenda from abs_xsec_per_species	 to propmat_clearsky.

But how to get at the absorption of individual gases? I see two possibilities:

1. Give propmat_clearskyAddLines and all other propmat_clearskyAdd functions the (already existing) variable abs_species_active as input. 
Method abs_lookupCalc would use that variable to select each gas species in turn, and would then call abs_xsec_agenda to compute absorption for that gas.

2. Or can absorption per gas already somehow be contained in dpropmat_clearsky_dx? I guess it is at least closely related to the VMR derivative. In that case, Richard, could you instruct me how to set this up?  How large is the overhead of this option compared to option 1?

Unless I am missing something, I think option 1 is more efficient, since it computes only a single output variable, whereas option 2 computes two, the absorption coefficients and the derivative. (But it would save subsequent division by the number density if the derivative is numerically identical to the cross-section already.)

riclarsson:
Option 2 is better I think.  Perhaps with a small modification.

In simple terms where ns is the number of species tags, nf the dense frequency grid, and nfs the sparse frequency grid the following happens.

Option 1 requires ns allocations of nfs + nf Complex data points and ns allocations of nfs Numerics.  It calls the interpolation routine ns times.  It also requires ns allocations of nf Numerics for the propagation matrix.

Option 2 without modification requires 1 allocation of (ns + 1) * (nfs + nf) Complex data points and 1 allocation of nfs Numerics.  It calls the interpolation routine ns times.  It resets values to zero ns * (nfs + nf) times.  It also requires 1 allocation of (ns + 1) * nf Numerics for the propagation matrix.

Option 2 with modification requires 1 allocation of (ns + 1) * (nfs + nf) Complex data points and 1 allocation of nfs Numerics.  It calls the interpolation routine 1 time.  It also requires 1 allocation of (ns + 1) * nf Numerics for the propagation matrix.

The modification of option 2 is to use VMR derivatives instead of ArrayOfSpeciesTagVMR derivatives.  This makes the output ever-so-slightly non-linear.  Perhaps it must be an option in the setup to switch between them?

[...]

PR #342, opened 2021-04-28
Faster LBL derivatives?

riclarsson:
Here's my output from the script Oliver provided in a currently open issue:

Time per pressure/temperature combination
CASE 1 - All species w/ jacobian ArrayOfSpeciesTagVMR: 1.586376s
CASE 2 - All species w/ jacobian Line::VMR           : 3.782718s
CASE 3 - All species w/o jacobian                    : 1.323047s
CASE 4 - One species at a time                       : 1.330896s

### DIFF
--- a/src/absorption.cc
+++ b/src/absorption.cc
@@ -679,7 +679,7 @@ void xsec_species(Matrix& xsec,
       // absorption cross-section
       MapToEigen(xsec).col(ip).noalias() += sum.F.real();
       for (Index j = 0; j < nj; j++) {
-        if (not propmattype(jacobian_quantities[j])) continue;
+        if (not jacobian_quantities[j].propmattype()) continue;
         MapToEigen(dxsec_dx[j]).col(ip).noalias() += sum.dF.col(j).real();
       }
 
@@ -687,7 +687,7 @@ void xsec_species(Matrix& xsec,
       if (not phase.empty()) {
         MapToEigen(phase).col(ip).noalias() += sum.F.imag();
         for (Index j = 0; j < nj; j++) {
-          if (not propmattype(jacobian_quantities[j])) continue;
+          if (not jacobian_quantities[j].propmattype()) continue;
           MapToEigen(dphase_dx[j]).col(ip).noalias() += sum.dF.col(j).imag();
         }
       }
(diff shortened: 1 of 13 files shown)


olemke:
I reran case 1 and 4 through the profiler. The remaining difference now comes down to the extra time spent in `ComputeValues::operator-=` for the derivates.

Case 1
case1_5000_4

Case 4
case4_5000_4

riclarsson:
Thank you for running the profiler, Oliver.  I will have a look.  It is definitely doing more than it should for case 1.  I think a similar cut-down in time should be possible as for frequency_loop, but I need to check.  (My initial quick-and-dirty stuff just now didn't solve it, so I need to have a more detailed look when there's time.)

riclarsson:
New times are slightly faster.  For the same case as above (5000 frequencies):
```
Time per pressure/temperature combination
CASE 1 - All species w/ jacobian ArrayOfSpeciesTagVMR: 1.486118s
CASE 2 - All species w/ jacobian Line::VMR           : 3.820726s
CASE 3 - All species w/o jacobian                    : 1.339322s
CASE 4 - One species at a time                       : 1.349771s
```

Also for 100,000 frequencies:
```
Time per pressure/temperature combination
CASE 1 - All species w/ jacobian ArrayOfSpeciesTagVMR: 21.045467s
CASE 2 - All species w/ jacobian Line::VMR           : 64.153041s
CASE 3 - All species w/o jacobian                    : 20.579016s
CASE 4 - One species at a time                       : 20.542538s
```

In #339 the case 1 to 4 ratio was 150-200%.  It is 2-10% now.  The case 2 to 4 ratio is pretty much the same as before since it is computing the full derivative.

riclarsson:
I'm marking this as ready for review since the previous tests showed where time was being spent.  I think the small 'regression' in case 2/3/4 is just noise.  They should not be seeing any difference that could cost time now cf the first time comparison of this PR (done at the commit before the latest)

olemke:
I agree that the slight differences in case 2-4 are negligible. I see basically the same times for 3/4 and 2 scores even slightly lower.

```
Before cc84172
CASE 1: 1.273507
CASE 2: 2.751986
CASE 3: 1.070707
CASE 4: 1.074574

After cc84172
CASE 1: 1.129954
CASE 2: 2.709163
CASE 3: 1.071539
CASE 4: 1.075069
```

PR #346, opened 2021-05-05
Fix some plot issue, add IGRF magnetic field to automatically compute…

riclarsson:
… it,

fix some time issues, move grids out, add a Zeeman test

The latter fails for unknown reason.  As with the controlfiles,
it should be possible to compare agains saved data.  For some
reason, it does not find the data it needs to compare against...

Note that there are two things to fix here before merge but I need input:

1) The zeeman-test is working but it doesn't know it.  There must be a way to compare against saved data...
2) There's a reimplementation in part of the legendre functions.  There seems to be some implementation available but I did not understand it and it seems to actually compute factorials, making it difficult to use.


olemke:
Regarding 1.: Like with the ARTS controlfile tests, Python tests are executed from the build directory. The current directory when running a Python test is `build/controlfiles-python/`. This avoids that files that are written to the current directory clutter the source tree. I've adjusted `ycalc.py` to read the reference from the source test directory.

olemke:
For the comparison against saved data you could just use the Compare WSM:

```python
    arts.VectorCreate("yref")
    arts.ReadXML(arts.yref, os.path.join(testdir, "refdata.xml"))
    arts.Compare(arts.yref, arts.y, 1e-3, "y reference validation failed")
```

olemke:
Regarding 2.: The "old" legendre functions are a straight adaption from the [GSL library](http://www.gnu.org/software/gsl/). But only `gsl_integration_glfixed_table_alloc` and `gauss_legendre_tbl` are actively used. All the others lie dormant since years, so I'm fine with removing them. I have a commit ready to push to do that.

riclarsson:
Thank you Oliver, I'll change it to use CompareRelative I guess.  It might cause issues since there are clearly a zero-crossings in the polarization, and if those match the frequency grid native probably won't act nice.

I'll move my legendre2 stuff and remove the old stuff at the same time.

riclarsson:
With that commit I think there's one major design issue remaining.  My `igrf13.cc` files contains static data.  This really helps getting the simulations up-and-running.  However, it is truly model data.

Should it be input from arts-xml-data instead?  The changes are:

1) `MagFieldsCalcIGRF` is renamed `MagFieldsCalcSchmidtQuasiNormLegendrePolynomial`
2) `MagFieldsCalcSchmidtQuasiNormLegendrePolynomial` takes `h` and `g` as time-grid data.
3) `igrf13.cc/h` are removed and replaced by some function that does just its `compute_impl`.
4) `GriddedField` has to take `ArrayOfTime` as a grid, or a special type has to be created and face the user.

1-3 are easy.  4) is cumbersome and the main reason I went with the current implementation.

I therefore will just leave this as it is and remove the draft-mark.

NOTE: The overarching pyarts test fails on my machine with the error:

```
=================================== FAILURES ===================================
____________________________ TestARTS.test_run_arts ____________________________

self = <test.test_arts.TestARTS object at 0x7f1b2d7c2610>

    @pytest.mark.skipif(not shutil.which('arts'), reason='arts not in PATH')
    def test_run_arts(self):
        """Test ARTS system call.
    
        Note: This test is only run, if ARTS is found in PATH.
        """
>       arts_out = arts.run_arts(help=True)
E       NameError: name 'arts' is not defined

test/test_arts.py:17: NameError
======================== 1 failed, 141 passed in 11.29s ========================
```

This seems unrelated to this commit, but if it isn't, it should be fixed.

riclarsson:
> Regarding 1.: Like with the ARTS controlfile tests, Python tests are executed from the build directory. The current directory when running a Python test is `build/controlfiles-python/`. This avoids that files that are written to the current directory clutter the source tree. I've adjusted `ycalc.py` to read the reference from the source test directory.

Just as a short comment:  The normal controlfiles also does this.  They print their files not in source, but they can still read files from the source automatically.  Is it not possible to have your `__file__` expression included in the command that calls the controlfiles-python reading environment?  At the very least in the search for `ReadXML`, if not in the search path of `pyarts.xml`?

riclarsson:
I just realized that I need to fix the documentation.  Please don't merge before the next documentation update.

olemke:
> Just as a short comment: The normal controlfiles also does this. They print their files not in source, but they can still read files from the source automatically. Is it not possible to have your `__file__` expression included in the command that calls the controlfiles-python reading environment? At the very least in the search for `ReadXML`, if not in the search path of `pyarts.xml`?

The correct source path was already added to the environment variable `ARTS_INCLUDE_DIR`. Therefore adding the path manually for ReadXML is not necessary. I removed it from there again.

olemke:
> NOTE: The overarching pyarts test fails on my machine with the error:

How do you run the tests to trigger this error? Running `pytest` in `build/python/test` automatically skips the test if `arts` is not in the `PATH`.

riclarsson:
> > NOTE: The overarching pyarts test fails on my machine with the error:
> 
> How do you run the tests to trigger this error? Running `pytest` in `build/python/test` automatically skips the test if `arts` is not in the `PATH`.

`make check-pyarts-only`

olemke:
Ah, I see it now, it fails if arts is in the PATH. Fixed. :-)

riclarsson:
> > Just as a short comment: The normal controlfiles also does this. They print their files not in source, but they can still read files from the source automatically. Is it not possible to have your `__file__` expression included in the command that calls the controlfiles-python reading environment? At the very least in the search for `ReadXML`, if not in the search path of `pyarts.xml`?
> 
> The correct source path was already added to the environment variable `ARTS_INCLUDE_DIR`. Therefore adding the path manually for ReadXML is not necessary. I removed it from there again.

I see.  I still think it would be useful to allow `pyarts.xml` in controlfiles-python since it helps a lot writing more python-friendly code.  Anyways, it is not that important and for another set of issues.

riclarsson:
I updated the documentation with more references.  There was also some other errors in the tex-documentation that I changed/fixed.

I think that this is ready for merge now.

olemke:
> 1-3 are easy. 4) is cumbersome and the main reason I went with the current implementation.

It is cumbersome, but if we decide to support `ArrayOfTime` as valid grid dimension, it needs to be done. I open an issue for that. Let's keep your current implementation for this PR.

olemke:
> I see. I still think it would be useful to allow `pyarts.xml` in controlfiles-python since it helps a lot writing more python-friendly code. Anyways, it is not that important and for another set of issues.

I mean it is allowed, you just have to pass the full path as shown. ;-) But I see your point. I just think that a general reading routine should not by default search for files in locations not explicitly specified by the user and possible find a file the user didn't expect to. `ReadXML` is different because it's used in the specific context of a controlfile/workspace. `xml.load` has a broader use. Maybe adding an additional bool option `search_arts_path` to `xml.load` to explicitly turn this on would be a good compromise?

riclarsson:
It's not for this PR, but something like that would be very useful.

PR #347, opened 2021-05-07
Optionally search ARTS path for files in xml.load

olemke:
Add option `search_arts_path` to `xml.load`. When set to `True`, `ARTS_INCLUDE_PATH`
and `ARTS_DATA_PATH` are also searched for the given file. The current
directory is always searched first. This replicates the behaviour of the
`ReadXML` WSM.


riclarsson:
Is there no way to get the default parameter from a class or global variable?  That way you could have something like this in the xml-file:

```python
def load(filename, search_arts_path=False):
    ...

def search_arts_environment(val = True):
    load.__defaults__ = (None, bool(val))
```

olemke:
Instead of monkey patching the default keyword value, I think I would rather reconsider my decision of having `False` as a default and change it to `True` instead, making `xml.load` behave the same as `ReadXML`.

PR #349, opened 2021-05-10
Select species

riclarsson:
Select a single `ArrayOfSpeciesTag` in `propmat_clearskyAddLines`.  This is in preparation for updating the lookup table.

Note limitations: If the default value of the species is not `""` or `NODEF`, the `autoarts.h` will fail to compile the test.  I think this should be quite OK and even desirable.  There's no use of having a non-empty default here (either by true empty, or by user having to choose themselves).

Note opinion: This adds `ArrayOfSpeciesTag` to the things that the parser deals with.  It expects the input to be the same as for `abs_speciesSet`.  This goes against how `Vector` and other array-types are created.  They have the form `[...]`.  I think it is better to treat `ArrayOfSpeciesTag` as its own type rather than an array, because that's how it is most often used in deep code, and how the user sees it in other places of the controlfile.

Note for future: I think `TokVal` is a bit over-designed for current generation C++.  It obviously works, but it is a bit difficult to add new types.  To do so, `TokVal` itself has to grow.  We should possibly change `TokVal` so that all classes with constructors in either the form of `class T { T(const String& s); };` or of the form `class T { T(const std::string_view sv); };` can be parsed.  It is easier to wait for C++20 for such changes as it should allow for `requires`-clauses to deal with most of the logic.  (`ArrayOfSpeciesTag` would have been redesigned using Simon's trick, and have a constructor that takes a `String` and wraps it around `array_species_tag_from_string`.)  I am not intending to do any of this in this PR but just venting an idea after having added `ArrayOfSpeciesTag`.

### DIFF
--- a/controlfiles-python/classes/TestSpeciesTag.py
+++ b/controlfiles-python/classes/TestSpeciesTag.py
@@ -42,3 +42,6 @@ assert str(ws.ast.value) == '""'
 
 ws.ArrayOfSpeciesTagSet(ws.ast, ArrayOfSpeciesTag([SpeciesTag("O2-66")]))
 assert str(ws.ast.value) == '"O2-66-*-*"'
+
+ws.ArrayOfSpeciesTagSet(ws.ast, [SpeciesTag("O2-66"), SpeciesTag("O2-68")])
+assert str(ws.ast.value) == '"O2-66-*-*, O2-68-*-*"'
--- a/controlfiles-python/classes/TestSpeciesTag.py
+++ b/controlfiles-python/classes/TestSpeciesTag.py
@@ -1,5 +1,6 @@
 from pyarts.workspace import Workspace
-from pyarts.classes.SpeciesTag import SpeciesTag, ArrayOfArrayOfSpeciesTag
+from pyarts.classes.SpeciesTag import (SpeciesTag, ArrayOfSpeciesTag,
+                                       ArrayOfArrayOfSpeciesTag)
 from pyarts.classes import from_workspace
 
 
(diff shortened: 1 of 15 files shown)


riclarsson:
@olemke Is the `TokVal` update acceptable?

riclarsson:
I've no idea why the python parser makes these mistakes.  I also cannot figure out how to make it a single-line command in python.  I need to do:

```python
ws.ArrayOfSpeciesTagCreate("tmp")
pyarts.classes.from_workspace(ws.tmp).append("O2-66")
ws.propmat_clearskyAddLines(select_speciestags=ws.tmp)
```

to make it work.

None of these work:

```python
ws.propmat_clearskyAddLines(select_speciestags="O2-66")
```

```python
ws.propmat_clearskyAddLines(select_speciestags=["O2-66"])
```

```python
ws.propmat_clearskyAddLines(select_speciestags=pyarts.classes.ArrayOfSpeciesTag([pyarts.classes.SpeciesTag.SpeciesTag("O2-66")]))
```

Especially the last, which contains a valid ARTS type but pyarts cannot handle it.  (I will also update the array-wrapper so `pyarts.classes.ArrayOfSpeciesTag(["O2-66"])` does what's expected.)

olemke:
I think I'm close to a solution, but have to do some further testing before committing and resolve one remaining issue in the conversion of controlfiles to Python. So far, these cases are all working now:

```python
ws.ArrayOfSpeciesTagCreate("ast")
ws.ArrayOfSpeciesTagSet(ws.ast, "O2-66, O2-68")
ws.Print(ws.ast)
ws.ArrayOfSpeciesTagSet(ws.ast, "O2-68")
ws.Print(ws.ast)
ws.ArrayOfSpeciesTagSet(ws.ast, "")
ws.Print(ws.ast)
ws.ArrayOfSpeciesTagSet(ws.ast, pyarts.classes.ArrayOfSpeciesTag([pyarts.classes.SpeciesTag("O2-66")]))
ws.Print(ws.ast)
```

```
- ReadXML
- Print
O2-66-*-* O2-68-*-*
- ReadXML
- Print
O2-68-*-*
- ReadXML
- Print

- ReadXML
- Print
O2-66-*-*
```

I agree that the special behaviour of `ArrayOfSpeciesTag` compared to "real" arrays is not good and causes some pain here as well.

olemke (approved):
Changes, including those to TokVal, look good. Anything you wanna add before merging?

riclarsson:
It's OK to merge now.

PR #351, opened 2021-05-11
Reproduce arts.2.3.1277 cutoff behavior

riclarsson:
It seems I removed a feature when I removed the very old xsec calculations from v.2.3.  I removed the possibility to have symmetric cutoff.  This is now back as a cutoff option (in only propmat_clearskyAddLines, the other code should be removed at some time so I don't intend to support it).

Anyways, if the cutoff type is set to SpeedIndependentSymmetricByLine then the cutoff frequency is now f0 + d0 + fc.  In the case of the cutoff type being ByLine, the cutoff frequency is just f0 + fc.  The latter is not symmetric.  Note that the first is also not symmetric when line mixing is active or when some beyond-Voigt line shape is used.


riclarsson:
Note that this is purely a "feature-add", so it can be merged without any influence on running code.  However, if there are any opinions on the name I choose for the new cutoff-option, then this must be addressed now since it potentially goes into the line catalog and will cause a lot of problem for people to change later.

stefanbuehler:
I don't understand the explanation, sorry. What does symmetric / asymmetric refer to?

stefanbuehler:
I still don't understand. 

To me it seems that there are only 3 reasonable options for cutoff:

1. No cutoff
2. Cutoff at line center +- the cutoff frequency
3. Like 2, plus subtract the lineshape value at the cutoff frequency (plinth removal)

1 and 2 are straightforward. For 3, there are two implementation options:

a) Check that lineshape is symmetric and throw runtime error in case it isn't.
b) Remove a trepezoid-shaped plinth (a linear line that goes from LS(f0-cutoff) to LS(f0+cutoff))

Option b has the advantage of being backward-compatible with how we have handled things so far.

In any case, from a user point of view (or from the point of view of what is encoded in the catalogue) there are only these 3 sensible options, I think.

Perhaps best to talk at the next vid-con? - Easier than in writing.

stuartfox:
Perhaps I can comment here, since I'm the person who first raised the question? I should point out that the example I sent was rather to illustrate something that I thought was a bug (i.e. something that produced non-physical data), and not a configuration I actually wanted to run. I came across the feature when I was trying to understand the source of some differences in simulations between an older v2.3 ARTS set-up and what I thought should do the same thing when translated into ARTS v2.4 syntax.

Clearly this is a bit more of a complicated issue than I originally thought though My actual use case is that I would like to be able to fairly closely reproduce the absorption properties used e.g. by the AER family of models. Here, a continuum has been defined based on subtracting lines with a frequency cut-off from a total absorption measurement, and AER provide a continuum model (implemented in ARTS) and a line catalogue that we should be able to use to reproduce the same total absorption. For this we need the option to use a line cut-off and subtract a "plinth" or else we will be double-counting the contribution to the total absorption. I thought this is what my old ARTS configuration was doing, although I must confess I never actually checked in detail. I haven't looked in detail about what AER actually do in their code to deal with the fact that the "plinth" value is not identical either side of the line for a non-symmetric line shape (as presumably everything is if you include a VVH or VVW pre-factor), but @stefanbuehler suggestion of a trapezoidal plinth sounds sensible. In any event, something that can lead to negative absorption values should probably be avoided without at least warning the user of the possible issue.

riclarsson:
@stuartfox I've no idea if the new code is the correct behavior or the behavior you are after as I've never worked with the AER continuum models directly.  I would also just assume the old setup did the right thing, and this PR simply adds that behavior back.  For some reason its accidental removal has not caused problems before now, which means it has perhaps not been important to people that this is doing the right thing.

I would suspect it isn't, since the AER types of continua only works if you have exactly the same line absorption calculations as was being used in the original fits anyways.  This makes these models completely unreliable outside of their particular ecosystem as far as I can see.

Anyways, the new code subtracts a symmetric plinth if the line shape is pure Doppler, Lorentz or Voigt.  This is exactly the same behavior as 2.3.1277.  If you were happy with the results to begin with, you should be OK.

If you use any pre-factors (VVH or VVW), or if you have any line mixing, or if you have any other type of line shape, the subtraction is skewed.  The pre-factors are not symmetric around the line center in ARTS, so they cannot be used if your goal is a flat plinth.  In fact, they tend to grow as the frequency gets larger so I would expect them to be able to produce negative absorption.  Again, this is exactly the same behavior as before so if you were happy...

A linearly changing subtraction should be able to fix at least the pre-factor problem.  It is not a terribly big change, but I would like it confirmed that this is indeed how it is intended to be used before I make any such additions to the code.  I still don't think it is worth for us to spend much time on these ecosystem locked continuum models without being able to use exactly the line absorptions that were used to generate them to adapt for our ecosystem, but that's much beyond this PR.

stefanbuehler:
@riclarson: Not clear to me what is implemented in this branch now. Your answer to Stuart could be interpreted in different ways.

a) No change required, this branch can just go away.
b) Code fixed so that it behaves as before.
c) The new SpeedIndependentSymmetricByLine option was added.

I think with c I would still have an issue, since I think that behaviour should be the default, not a separate option that people might miss.

stefanbuehler:
@stuartfox: The old behaviour was that the plinth was subtracted from the (symmetric) Voigt line (after pressure shift and Doppler shift are applied, so around the actual line center). The pre-factor was applied afterwards. Since it is multiplicative, absorption value zero at +- the cutoff frequency from line center still holds. In the case of VVW, the mirror line at negative frequency was treated as an individual line. So it had its plinth removed in the same way. Note that except at quite low frequency, the mirror line often did not affect absorption at all, since it was outside its cutoff.

stuartfox:
Ok - the old behaviour sounds sensible to me @stefanbuehler . However, I'm not sure that the new option @riclarsson has added does let you reproduce it as it doesn't work with a non-symmetric lineshape (producing large negative brightness temperatures). I can give an example if required (basically, change the line cutoff to 750GHz and add a VVW pre-factor, plus the controlfile changes @riclarsson indicated to use the SpeedIndependentSymmetricByLine option).

stuartfox:
Actually - it seems that v2.5 introduces a new problem not present at v2.4 when using ByLine cutoff with a non-symmetric lineshape - I've put an example here https://github.com/stuartfox/arts-cutoff-problem/tree/realistic-test 

There are small (up to 0.08K) differences between v2.3 and v2.4 which are presumably caused by the frequency-shift issue with the ByLine cutoff that @riclarsson has already identified, but then the v2.5 results have large negative brightness temperatures across a lot of the spectrum.

riclarsson:
For some reasons, there seems to be old commits added to this...

Anyways, the 'guaranteed' symmetric plinth-cutoff is now in place.  I think.  This only applies to code using `propmat_clearskyAddLines`.

I decided for now to remove all other options of cutoff as it is no longer relevant for speeding up LBL calculations.  Keeping the bare minimum makes the checks easier to verify.  This means you can only use "None" and "ByLine" as options, the latter is modified to work as the old ARTS-way did it.  So some simulation results will change as a result of this PR.

I would strongly recommend adding some tests to "artscomponents" in either python or pure ARTS to ensure that the cutoff behavior is consistent in the future with the use case they are intended for.

About checks, the LBL checks are a lot stricter now.  If they fail, you can as always say they pass by parsing `lbl_checked = 1`, but then you are on your own in terms of reliability of the code.

olemke:
> For some reasons, there seems to be old commits added to this...

The old commits go away if you rebase against master `git rebase upstream/master`.

riclarsson:
@olemke The `git rebase upstream/master` command removed my latest changes for some reason...

olemke:
I don't know. I ran it before on your branch and it left exactly the 4 commits not already present in master as expected.

```
commit f6af0ffbf43371cf9739a0ac677f40e33726d288 (HEAD -> riclarsson-master)
Author: Richard Larsson <larsson@mps.mpg.de>
Date:   Mon May 31 07:12:16 2021 +0200

    Remove likely unused options of cutoff calculations and only
    retain no-cutoff and symmetric cutoff options.  Also make strict
    checks in lbl_checkedCalc that any line using cutoff is actually
    following these rules.  No Zeeman, no line mixing, no beyond-Voigt
    is allowed when using cutoff

commit 5c0cbb680410da9d4c74d464c84c963f69789cde
Author: Richard Larsson <larsson@mps.mpg.de>
Date:   Mon May 31 06:34:28 2021 +0200

    Make cutoff calculations inline with the line shape(s)

commit 3dbba10a7989a642e3116e7af98f98ab0cec949f
Author: Richard Larsson <ric.larsson@gmail.com>
Date:   Tue May 11 11:31:45 2021 +0200

    Revert selection change and instead add a symmetric option for cutoff
    in relation to the speed independent shift.  All other line center
    shifts results in non-symmetric line shapes so using cutoff is anyways
    against sanity

commit 3f06fa2461f879245c9bb102307e4fac4503c856
Author: Richard Larsson <ric.larsson@gmail.com>
Date:   Tue May 11 11:01:04 2021 +0200

    Select the cutoff based on closest proximity to f0

commit 2a347b9ca49d3bbb626f88734e532893b7673b04 (upstream/master, master)
Merge: 9568bebf7 8482ba849
Author: Oliver Lemke <olemke@gmail.com>
Date:   Thu May 27 16:35:17 2021 +0200

    Merge pull request #353 from olemke/fix-debug-build

    Fix debug build
```

riclarsson:
Hmm, I think that I have to rewrite the code then...

olemke:
Should I force push my branch with the 4 correct commits? I still have it here.

riclarsson:
Yes, please!

olemke:
Done. Hope this helps.

riclarsson:
It helps a lot, thanks!

This brings the current issue to a close if someone can verify that this code works as intended.

riclarsson:
There's now even a test.  This should be merged now.  @stefanbuehler or @stuartfox, is there anything else you would need?

The changes in control-files should be zero compared to before this PR.  This branch effectively removes all other options than "ByLine".

So the following things happens in this branch:

1) The symmetric plinth is restored from my understanding of 2.3.1277 behavior.

2) lbl_checkedCalc Properly goes over all the lines and checks that if cut-off is active, then there's no line mixing, there's no Zeeman effect, and there's no beyond-Voigt line shapes.

This should be enough to guarantee that the plinth is symmetric.

Issue #350, opened 2021-05-10
CO2 absorption calculation fails because partition function coeffs are 0 for CO2-737


olemke:
**Describe the bug**
Since the new isotopologue CO2-737 was added, absorption calculations for CO2 contain NaNs because the [partition function coefficients are set to 0](https://github.com/atmtools/arts/blob/7ef4fcdcded9b988df8e07ee727583ac4bddc37f/src/partition_function_data.cc#L412).

**To Reproduce**
1. Set `abs_species` to "CO2"
2. Make an absorption calculation

Red areas mark NaNs for a calculation with a 750e9 Hz cutoff. Without cutoff, y contains only NaNs.
![co2nans](https://user-images.githubusercontent.com/665724/117667741-7eccc000-b1a5-11eb-85d9-4e66bec3365b.png)

The calculation succeeds if CO2-737 is excluded.

If applicable, error output:
```
Run-time error in controlfile: co2nans.arts
Run-time error in method: yCalc
User Error: failed
Error is found at:
        /Users/olemke/Hacking/uhh/arts/src/m_rte.cc:1556
Please follow these instructions to correct your error:
User Error: std::isnan(y[ii])
Error is found at:
        /Users/olemke/Hacking/uhh/arts/src/rte.cc:2356
Please follow these instructions to correct your error:
One or several NaNs found in *y*.
Stopping ARTS execution.
Goodbye.
```

**Expected behavior**
Absorption calculation should succeed. The current solution for a full CO2 calculation is excluding CO2-737 explicitly by setting `abs_species` to "CO2-626, CO2-627, CO2-628, CO2-636, CO2-637, CO2-638, CO2-727, CO2-728, CO2-828, CO2-837, CO2-838". Which is very cumbersome. We should either provide the partition function coefficients for all species or exclude isotopologues with 0 coeffs automatically from the calculation or at least throw an error.

**Additional context**
Same applies for other recently added isotopologues: NO2-656, OCS-634, CS2-\*, C2N2-\*, COCl2-\*

riclarsson:
Is this with the flag for bad partition functions active?  If so, then this is acceptable behavior.  Otherwise, it must clearly throw in the relevant check function.

olemke:
> Is this with the flag for bad partition functions active? If so, then this is acceptable behavior. Otherwise, it must clearly throw in the relevant check function.

The flag for bad partition functions was not active.

[...]

Issue #326, opened 2021-03-19
Planetary tests fail due to missing iso ratio for new CO2-737


olemke:
**Describe the bug**
We recently added the new isotopologue CO2-737 to species data. The planetary tests now fail, because this isotopologue is not available in the input files provided in `arts-xml-data`:

```
arts-xml-data/planets/Jupiter/isotopratio_Jupiter.xml
arts-xml-data/planets/Mars/isotopratio_Mars.xml
arts-xml-data/planets/Venus/isotopratio_Venus.xml
```

**To Reproduce**
Run the planetary surface tests in your build directory:
```
ctest -E python -R arts.ctlfile.xmldata.artscomponents.arts-xml-data.TestSurf
```
In the current master, this triggers an assertion, but with the fix in PR #325, you'll get:
```
Run-time error in controlfile: /Users/olemke/Hacking/uhh/arts/controlfiles/artscomponents/arts-xml-data/TestSurf_Venus.arts
Run-time error in method: lbl_checkedCalc
User Error: isoratios.getParam(sp, iso).nelem() == 0
Error is found at:
        /Users/olemke/Hacking/uhh/arts/src/absorption.cc:333
Please follow these instructions to correct your error:
No isotopologue ratio found for CO2-737
```

The question is what is the correct way to approach a fix here. Even if we add the appropriate isotopologue ratios to those files, the planetary tests are using the Perrin catalogue, which doesn't provide any CO2-737 lines anyway.

riclarsson:
I think your solution is good enough.  Combine this with limiting the planetary test to things it can have known about on creation.

The planetary tests are clearly poorly designed to not work in case a new molecule is added.  To support all molecules is a goal not a demand.  We should not stop adding things for Earth just because it doesn't work yet on other planets.

That said, it does present a bit of a greater problem with the general design of the isotopologues:  In cases where the ratio is unknown, you have to hope that there's no other molecule partly named CO2 that influences the spectra around your target CO2 molecule.  If there is, then you cannot perform reliable forward calculations.

riclarsson:
#354 and updates to arts-xml-data closes this

PR #354, opened 2021-05-27
Partfun

riclarsson:
This is a suggested solution to issues #350 and partly #326.  And a general redesign of how species are dealt with in ARTS.

All `make check-pyarts-only` works as is.  All but the documentation tests of `make check` works with `-DENABLE_BUILTIN_PARTFUN=1`.  Several tests fail without this flag.  I made a commit to arts-xml-data so that the thing should build, which by defaults means `-DENABLE_BUILTIN_PARTFUN=0`.  All failing test but the documentation is caused by the fact that we have bad builtin partition functions.

This PR removes the two cc-files with species and partition functions data.  These are instead replaced by builtin Earth data similar to today for the ratios, and compile-time partition functions.

The isotopologue ratios are moved into either HITRAN data or into the default-builtin function.  The new user-facing type is called `SpeciesIsotopologueRatios`.  Several things about it is incomplete as of now with this class, but it works for the builtin ratios.  Much of the data in arts-xml-data should be adopted to use this type instead of what is currently in place.

The partition functions are now automatically compiled from data files.  Error handling will work properly if you are missing a species in the partition functions (you will get a user-error thrown at you).  The main thing is that the compiled partition functions are independent from the data we have in our defined isotopes.  To add a new partition function is as simple as adding another file to the partition functions data folder and recompiling ARTS.

There are several technical issues still in this PR that needs to be sorted before any merge (e.g., File IO).  There are also a few design choices to discuss (e.g., should the isotopes be properly pointed at instead of referred to by index).  For now, I will leave those aside for general opinions on the PR.

It changes major and quite fundamental parts of ARTS so there's bound to be a few mistakes in the code and some growth pain, but those parts have been considered cumbersome for quite some time and actively causes issues for people when touched today.  The PR also makes it easier to use proper partition functions instead of bad partition functions, which is clearly a bad practice that most of ARTS community is happy with regardless...


riclarsson:
Current technical issues to be solved before any merger is even on the table:

1) File IO for generating the partition functions requires new 3rdparty XML-parser.  This is not desired.  It would be better to use one of the matpack XML-parsers.

2) Some of the tests are significantly slower.  I think there's some edge-case I am missing here where using the new data structures are weird.  These slower tests include "TestAbs.arts" and "TestClearSky.arts" and "TestRegridFields.arts".  They are slower by a factor 3-5, so it is clear there's something amiss.  (Other tests are faster, such as "TestSpectroscopy.arts" taking about half-as-long, so it's clearly some edge cases.)

3) Any other wrong technical things?

Current design issues:

1) Should SpeciesTag and QuantumIdentifier carry a pointer/reference to their values inside the isotopologues list?  Currently they have an index to it instead.  The former is more elegant and demands that things work or fail instead of being able to carry bad data around.

2) Should the IsotopeRecord carry the partition functions with them (plural, because two functions are required - one for the temperature derivative)?

3) Any other design questions?

riclarsson:
As of 22f593c "make check" works again.  We decided on meeting to be opinionated, so the next commit breaks many controlfiles again.  NOTE: These breaks are now only because ARTS was broken by default anyways.  The errors seen in these controlfiles are physical misrepresentations by the current master-branch's parameterization of the partition functions

riclarsson:
As of 22f593c "make check" works again.  We decided on meeting to be opinionated, so the next commit breaks many things again.

riclarsson:
Current open technical issues:

1) Update all the test-results that use partition functions.  How sure are we the results are still good, or do we even care because we understand why they are different now so the test results are still valid?

2) Switch to TIPS 2021 if data can be found.  Where is it?  I cannot see it on hitran.org

3) File IO still needs 3rdparty XML parser.  Is it OK to just extract matpack IO to its own build-step to deal with this?

4) Isotopologue ratios lacks file IO currently.  We decided to go with a single-file solution at meeting.

@olemke Thanks for dealing with all the other tests that wants to use the bad partition functions!

olemke:
> 3\. File IO still needs 3rdparty XML parser.  Is it OK to just extract matpack IO to its own build-step to deal with this?

I'll separate the Matpack IO and also move GriddedField into matpack. The parser also needs some refactoring because ArtsXMLTag currently depends on artscore types such as `ArrayOfSpecies` and others. Will keep you posted.

riclarsson:
@olemke `GriddedField1` is the current implementation only.  I think it would be perfectly acceptable to use a special struct that looks sort of like:

```cpp
struct PartfunData1 {
PartitionFunctions:: Type type;
Vector Q;  // Also served as Coeff
Vector T;
};

// or

struct PartfunData2 {
PartitionFunctions:: Type type;
Matrix data;  // Size depends on type
};
```

And for writing the file IO of this type, I would just need the IO of `Vector` or `Matrix` exposed in `matpack` instead of in `artscore`.  This seems like less of a change.  Perhaps it is cleaner to get all of the `matpack` types IO into `matpack` the library, just for some consistency, but moving gridded-field seems like more of a change.

(My main concern is that gridded-fields are not part of `matpack` currently, so you would need some intermediary library just to combine the IO).

olemke:
Even when just extracting Matrix and Vector, the parser still has to be refactored due to the mentioned dependency on artscore. And since that has to be done anyway, the cleaner solution is to provide independent IO for all of matpack and not just a subset.
If you don't require gridded field, I will hold off on moving it to matpack for now. It might restrict us in the future if we want to add a more complex grid types that are only available in artscore (e.g. Time).

riclarsson:
@olemke It seems I have mostly figures out why git hub doesn't want to build some of this stuff.  However, both gcc-8 and clang are still fighting me quite a lot.  Why?  I tried linking against the filesystem-library that gcc needed in the way-back-when past, but that didn't help.  Clang is even weirder than gcc-8, it seems to not understand that there's a way to generate a long from a ulong.  Instead it insists it has to be generated from some class...

olemke:
I'll have a look at the errors.

olemke:
The preliminary filesystem support in GCC 8 and especially on Ubuntu/Debian seems very broken. I've managed to successfully compile this branch on macOS and Manjaro with GCC 8 by linking against std++fs. But Ubuntu and also Debian on our server just plain out refuse no matter what.
I think the time has come to consider finally dropping GCC 8. But before making any decision in that regard, I have to look into options to get a newer GCC onto our cluster. Not being able to compile ARTS on our own cluster is not an option.

For clang, the compilation on Ubuntu works as you've seen from my `2af7974` commit if we link against libstdc++ instead of libc++. Again a Ubuntu problem unfortunately, since on macOS und Manjaro the compilation works fine even with libc++.

> Clang is even weirder than gcc-8, it seems to not understand that there's a way to generate a long from a ulong. Instead it insists it has to be generated from some class...

Was that an earlier error that you fixed in the meantime? Because after fixing the filesystem issue, clang compilation seems fine.

riclarsson:
@olemke I didn't fix the long conversion issue.  It was perhaps why I couldn't understand the errors.  (As in, the error as I saw in the logs wasn't really there, but just output because the build crashed when reaching that point.)

About old Ubuntu and `filesystem`.  It is of course possible to not use `filesystem`.  I don't like it, since it is such a convenience, but we are not using it elsewhere.  Did you also try if the `filesystem` headers lives in `<experimental/filesystem>` on `gcc8`?  I remember having to use those headers for some small project a while ago, so perhaps this was `gcc8`?

I don't mind dropping `gcc8` otherwise.  My main issue seems to still be with clang.  Their standard library randomly being out-of-spec is a horrible issue.  (And they still don't support quite a lot of C++17 library features, so this will continue to bite us...)

olemke:
> @olemke I didn't fix the long conversion issue. It was perhaps why I couldn't understand the errors. (As in, the error as I saw in the logs wasn't really there, but just output because the build crashed when reaching that point.)
> 
> About old Ubuntu and `filesystem`. It is of course possible to not use `filesystem`. I don't like it, since it is such a convenience, but we are not using it elsewhere. Did you also try if the `filesystem` headers lives in `<experimental/filesystem>` on `gcc8`? I remember having to use those headers for some small project a while ago, so perhaps this was `gcc8`?

The problem is not the header files. On our server, I get unreferenced symbols from the linker even though I'm linking against `-lstdc++fs`.

In the github action, the compilation of `gen_autopar_tfun` is actually successful (with and without linking `stdc++fs`), but the execution of that executable fails with a segfault:

[compilation succeeds](https://github.com/atmtools/arts/runs/2716211002#step:6:322)
```
322 [ 21%] Built target gen_auto_partfun
```

[execution segfaults](https://github.com/atmtools/arts/runs/2716211002#step:6:390):
```
390 [ 32%] Generating auto_part_fun.h
391 Segmentation fault (core dumped)
```

> I don't mind dropping `gcc8` otherwise. My main issue seems to still be with clang. Their standard library randomly being out-of-spec is a horrible issue. (And they still don't support quite a lot of C++17 library features, so this will continue to bite us...)

GCC and Clang are both state of the art compilers, both with their flaws. But they're both widely used and here to stay. And clang offers a lot of benefits with its modern design and tools built on top of its architecture such as clangd. Sure it's not perfect (yet). If clang doesn't compile something, it does not necessarily mean it's out-of-spec. GCC is often more lenient whereas Clang follows the C++ standard more strictly. It can be annoying but in the end standard compliance is the best way to portability in the long run.
In the case of `filesystem` not working with `libc++`, it appears to be a standard violation on the first glance. But since it compiles fine on macOS und Manjaro Linux (with the same clang version), it seems more like an issue with the particular `libc++` version in the Ubuntu VM used by the Github action.

riclarsson:
The reading of the partition functions is separated out of the generation function and into its own xml-code-file.

I noticed some problems.  The following species are defined very differently in ARTS cf AFGL/HITRAN:

1) CH3CN (we have 6 indices, AFGL/HITRAN uses 4)
2) H2CO (we have 4 indices, AFGL/HITRAN uses 3)
2) HCOOH (we have 4 indices, AFGL/HITRAN uses 3)

I made the decision to remove built in partition functions for those species I understand the conversion based on old species_data.cc descriptions.  The README.txt is updated with all of these changes.

I am not sure if I should also rename the species in TIPS but not in ARTS?  It seems a waste, but perhaps it is necessary?

riclarsson:
@olemke OK, I don't know what else could be the issue with gcc8 then.  Let me know if it can be removed as a dependency or if I must change to use strings instead of `filesystem`.  I don't really know how it could easily be done, but that's another issue...

olemke:
The checks are not running in this PR at the moment because of the merge failure. You can still see them in your actions tab on your branch. https://github.com/riclarsson/arts/actions

But I would like to do a clean rebase against master. But only if you don't have any local changes/commits on your end that haven't been pushed yet. Let me know if it's okay, then I'll do the rebase.

I've also just found the solution for GCC 8 compilation on our cluster. With `link_libraries` it works. Let see if it also works on Github now. https://github.com/riclarsson/arts/runs/2718408076

riclarsson:
I've no uncommitted code at this point. I will commit new partition
functions data as soon as I have time out of the lab

olemke:
Finally figured out the issue with the missing header files for libc++. Turns out that things got messed up because the build action installed the OS version of `libc++` instead of the one matching the compiler. Also that removed the `c++abi-dev` package with screwed things up further.

olemke:
> The reading of the partition functions is separated out of the generation function and into its own xml-code-file.
> 
> I noticed some problems. The following species are defined very differently in ARTS cf AFGL/HITRAN:
> 
> 1. CH3CN (we have 6 indices, AFGL/HITRAN uses 4)
> 2. H2CO (we have 4 indices, AFGL/HITRAN uses 3)
> 3. HCOOH (we have 4 indices, AFGL/HITRAN uses 3)
> 
> I made the decision to remove built in partition functions for those species I understand the conversion based on old species_data.cc descriptions. The README.txt is updated with all of these changes.
> 
> I am not sure if I should also rename the species in TIPS but not in ARTS? It seems a waste, but perhaps it is necessary?

Since we base the partition functions on TIPS, it seems more consistent to use the TIPS names now? But that would of course break backwards compatibility and require renaming the species in our catalogs. So maybe not worth it.

@stefanbuehler What's your take on this? For reference: [Partfun README](https://github.com/riclarsson/arts/blob/partfun/src/partfun/README.txt)

stefanbuehler:
I think I am also in favour of using the TIPS names for the species 
where there is a clash. Hopefully less maintenance work in the future. 
But if you do this work now, you should really base it on the current 
(2021 paper I sent) TIPS version, not on the 2017 version!

Also, this shows that we need a good version number scheme for the ARTS 
catalog. I think I have pointed this out before. Probably it should be 
an additional variable (not just a comment) in the catalog file, so that 
at runtime we can check that the version number is not too old.

stefanbuehler:
Darn, I did not realize TIPS uses HDO. I don’t like it, what do you 
think? If that is used only in few places by TIPS, I would suggest to 
rename it there. Are there other places, where our convention is more 
stringent/logical than TIPS?

riclarsson:
Yes, TIPS documents using HDO, as does HITRAN in fact.  TIPS consistently uses D for deuterium in all of their documentation.  You see that in their papers.

But we need not adhere to it too stringently.  Neither HITRAN nor TIPS follow their own convention in their software.  The main problems are the three species above and that we switched the order on CO2-827 to CO2-728 in an inconsistent move (which should be reverted regardless).

I think we should follow the TIPS software until we have a good reason to go against it.  I don't think following their documentation is good since it is inconsistent with their software.  Their software is almost by all definitions how they intend their results to be used.

stefanbuehler:
That sounds reasonable.

riclarsson:
The only thing remaining for this PR is to switch to use matpack-derived IO.

@olemke What is your status on using built-in XML routines?  I could attempt to extract the relevant parts myself tomorrow if you don't have time.  It seems to me that extracting

```cpp
  void read_from_stream(istream& is);

  void write_to_stream(ostream& os);

  void add_attribute(const String& aname, const String& value);

  void add_attribute(const String& aname, const Index& value);

  void get_attribute_value(const String& aname, String& value);
  
  void get_attribute_value(const String& aname, Index& value)
```

to their own base-class is enough to support an xml_io_matpack.cc/h combo.  Some additional methods might be required.

I would like this merged as soon as possible.

olemke:
I started working on it and will give an update tomorrow.

olemke:
@riclarsson It's done. To read matpack types in `xml_io_partfun_h`, you need to include `xml_io_base.h` and link `gen_auto_partfun` against the `xmliobase` library. See the `test_xml` target as an example.

riclarsson:
@olemke Thank you!  This made it very easy to remove the 3rdparty XML-library and just use the internal stuff.  I still need to clean up a few things before I'll mark this ready for review.

Do you understand why mac cannot run the CPP-api, though?  There are no calls to any negative frequency values.  f_grid is always above 21 GHz.

olemke:
The failure is completely random. Happens about every other time. Error messages also vary. Sometimes negative frequency, sometimes NaN in y. I suspect that proper initialization is missing somewhere. Unfortunately, when I run it in the debugger, the error doesn't happen. Just tried like 20 runs. :-(

riclarsson:
Right, this is a bit cumbersome.  I cannot debug it since it is on a mac.  I don't think mac and linux gcc does that much differently when it comes to default initialization?  The setting of the ARTS variables does go through quite some hoops to be manipulated, perhaps I've introduced some undefined behavior there that mac's cannot have?

My only remaining goal to work on now is to add some simple python interface to PartitionFunctionsData.  It's not really a user-space variable but it helps if it is possible to manipulate them.

olemke:
Found it. It's really dangerous to use Touch as a replacement for initialization. Its premise is that it doesn't alter the values of existing variables. So for Numeric and Index, they will not be assigned 0.
I don't want to change the behaviour of Touch, because it can otherwise lead to bugs if controlfiles call Touch on existing variables.
We could introduce a Reset() WSM that also reinitialises a variable.

riclarsson:
@olemke Thank you! Excellent find.  It's a physical parameter so ignoring it is definitely not good, it was my mistake to not set it.

I think with that commit and the latest python-interface additions, this is ready for some review before merging.

riclarsson:
The main thing to fix here is if everything is clear?

I have no idea, since it is all written by me and I can understand my own code and intent.  @stefanbuehler , do you follow the code and/or intent?

(I will update absorption_theory.tex and absorption.tex, which are the only two documentation files talking about partition functions that I can see.)

stefanbuehler:
I have not followed the code changes. Yes please update the latex and 
builtin documentation to match the new implementation.

olemke:
@riclarsson I will take a look through the code tomorrow.
If the rebase from master gives you trouble, just let me know and I'll do it tomorrow morning.

riclarsson:
@stefanbuehler Documentation is now up-to-date, I think.

@olemke I managed to rebase it, I think.  Just had to rerun some simulation because apparently both branches did stuff to NLTE calculations.

riclarsson:
I added a check that all partition functions exist.  It is runtime-only because I didn't like the compile-time errors and didn't know how to display them better.  Perhaps this is better for `test_isotope.cc` as well, to have it fail instead of printing when it does fail and then make it part of the `make check` routine?  It would be trivial to make it a checked test.

@stefanbuehler and @olemke, to test that my code is understandable, I would suggest that you try and add either a new species or a new isotope.  It's the only way I can think of that would make you experience the code enough to know if it should be changed.

Otherwise, your review of the code itself tomorrow will be very appreciated, Oliver.  If you find nothing you think needs to be added or changed, and if there are no additional comments by Stefan, then I think this is merge-able.

There's some additional stuff in ARTS-XML-DATA that needs to be synchronized with the merge though.  Mainly, all references to `SpeciesAuxData` should be removed.  The ones for partition functions should simply be deleted.  The ones for isotopes should be updated so the type is `SpeciesIsotopologueRatios` and the `@`-character is removed.  Also their `version` and `nparam` parameters are no longer used, they could be removed or they will simply be ignored.

olemke:
@riclarsson I've converted the `isotopratio_Venus.xml` to the new format to do some testing with externally provided isotopologue ratios. Here is what I found:

1. Species that are present in the isotopratio file but not known to ARTS produce a clear error message. That's good.
2. Species that exist in ARTS, but are not listed in the isotopratio file do not produce an error about the missing isotopologue ratio. This would be ok if that isotopologue has no lines in the catalog (e.g. CO2-737). But there is also no error if the calculation contains lines of that isotopologue. As a test, I removed CO2-628 from the isotopratio file. The only indication that "something" is wrong is a runtime error in `yCalc` about NaN values in y.

If think 2. should produce a more helpful error. [Here is my updated`isotopratio_Venus.xml` file](https://gist.githubusercontent.com/olemke/1f9f81dadc020e963fbd25f873419873/raw/193be0c197b9bbd6f4a9c0985ed3391c74d649e5/isotopratio_Venus.xml) with CO2-628 removed if you want to drop it into `arts-xml-data/planets/Venus/` and run the Venus test case yourself:

```
ctest -V -R ^arts.ctlfile.xmldata.artscomponents.arts-xml-data.TestSurf_Venus
```

olemke:
Can we avoid the redundancy of having all isotopologue ratios defined in `hitran_species.cc` and `isotopologues.h`? Aren't the builtins identical to the latest molparam?

riclarsson:
The built-in values are for computation, the Hitran species are for reading Hitran.  If Hitran ever change one of those values, we cannot update to use it without losing again the ability to read old Hitran databases.  So they must be treated separately.

I must have failed in my update of the isotopologue check.  It's supposed to be caught there if you may access a species which you cannot compute because of the lack of ratios.

riclarsson:
The error of the venus-test with your file now reads:

```
168: Chosen joker species: CO2-*-*-* have missing isotopologue ratio.
168: The following species have isotopologue ratios:
168: [CO2-626 CO2-627 CO2-636 CO2-637 CO2-638 CO2-727 CO2-827 CO2-828 CO2-837 CO2-838],
168: and the following species do not:
168: [CO2-628 CO2-737]
168: One way to continue your simulations is to exclude the species missing isotoplogue ratios,
168: another is to add the missing ratio(s) to your data file
```

riclarsson:
Ok, I went back to this and realized it didn't actually solve the problem right.  Even when CO2-628 was added back, the error checking above would fail for CO2-737.  This wasn't the original intent.  So now the checks are based on the absorption lines.  The failing error is now:

```
168: Please follow these instructions to correct your error:
168: Species: CO2-628 has no isotopologue ratios
168: You must add its isotopologue ratios to your included data or
168: exclude the species from your computation setup
```

If the isotopologue ratio for CO2-628 is added back now, then the `ctest` you reference above passes.

olemke:
Thanks @riclarsson . It think that's handled in the best possible way now.

riclarsson:
Excellent, thanks!

The remaining thing was about combining HITRAN species isotopologue ratios and the builtin isotopologue ratios.  I think they must be kept separate for our future selves' sanity.  To expand on that:

Imagine if HITRAN adds CH2DCN.  We have in the builtin the isotopologue ratio of 0.441185e-03.  If HITRAN chooses to use a different value, we cannot read HITRAN and preserve old calculation results.

Making reading HITRAN and computing the absorption spectra depend on each other means we cannot add species outside of HITRAN and have our educated guess of the isotopologue ratios be different from whatever HITRAN would use in the future.  CH2DCN is not in HITRAN, but in JPL and via it, ARTS.

It's not easy to maintain a single list in this way for all species.  It's better to just accept we need a bunch of static data to be able to understand the data we get from HITRAN.

Note also that the two lists are not exactly the same today.  The HITRAN numbers are not from ARTS but from `molparam.txt` that is provided with HITRAN.  So some things that looks like bugs (but might be design decisions, so I have not made any changes here) like us defining COCl2 as having both species with exactly the same ratio, is now preserved in the calculations but not in reading HITRAN.

stefanbuehler:
> The remaining thing was about combining HITRAN species isotopologue 
> ratios and the builtin isotopologue ratios.  I think they must be kept 
> separate for our future selves' sanity.  To expand on that:

Your initial reply to Oliver was completely right. The HITRAN table ist 
necessary for reading the HITRAN catalog and nothing else. It has to be 
a separate table.

olemke:
Yes, I see why we need to have those separate. Makes perfect sense.

I'm currently testing the procedure of adding a new (imaginary) species and will provide some feedback on that soon.

olemke:
I succeeded adding a new isotopologue and species. So that's good. :-) I'll summarize what I did to also give @stefanbuehler an overview what changed compared to the old implementation:

Adding an isotopologue to an existing species is straightforward like before. Instead of adding it in `species_data.cc` and `partition_function_data.cc`, the name, mass and gi[1] are now added in `isotopologues.h` to the `Isotopologues` array and the ratio in the `isotopologue_ratiosInitFromBuiltin` function.

[1] The gi parameter would benefit from some documentation or a more telling name. It's not obvious (to me ;-) ) what it means.

Adding a new species is a bit more involved than before. In the old implementation, similar to just adding a new isotopologue, the new species had to be added to `species_data.cc` and `partition_function_data.cc`. In the new implementation, in addition to adding the new species in `isotopologues.h`, it requires touching a couple more places. The species is added in `species.h` to the `Species` enum class, to the `fromShortName`+`toShortName` functions which map the string names to enum values and vice versa, in `partfun.h` to the `partfun_impl` function responsible for the partition function calculation (+derivative) and in `isotopologues.cc` to the `isotopologues(Species spec)` function to retrieve the isotopologues of the given species.

Despite adding a species in the new implementation requires a few more steps, I think getting rid of the global `species_data` and `SpeciesMap` variables and their initialization at runtime is beneficial. Errors in the species and isotopologues already being caught at compile time is safer.

One question regarding the long names used in the `Species` enum class: I'm not requesting any changes, just want to understand the reason for introducing the set of longer names (e.g. HypochlorousAcid for HOCl) instead of using the same species names from the isotopologues also for the enum values.

riclarsson:
I've added some Doxygen-styled comments to each of the internals of the IsotopeRecord.  gi is a degeneracy term.  I think in general it means that there are by default many ways to reach the same energy levels of a molecule if it is above 1.  In turn, if you need to estimate the partition function yourself (e.g., within the framework of NLTE), then you may need to take it into account.  I just left it in because it's in HITRAN, and I think it's fine to leave the kind of data that could help us down-the-line around.  (I remember I have had use of it when dealing with HITRAN line mixing, but I don't know what line or how...)

The long names for `Species` is basically there because I didn't want to have to deal with "NO+" in any weird way.  It shouldn't slow anything down, and I sort of like having the readable names around when looking at the code.

riclarsson:
Also, adding a new Isotopologue is more involved than you say.  You need to also add its partition function or "make check" will fail.

olemke:
> Also, adding a new Isotopologue is more involved than you say. You need to also add its partition function or "make check" will fail.

True, was a bit too concentrated on the code side of things and taking it for granted that the partition file has to be added as well. :-)

The script for converting the TIPS data would also need to be included in ARTS. Probably not the data itself, better a README with a link to the Zenodo page. `3rdparty/TIPS/` would be a good place I guess. Could you please add it? We can afterwards remove the old `3rdparty/partFct`.

Thanks for the detailed explanation on the degeneracy term and adding code documentation for it.

Regarding the long names, I'm fine with keeping them.

riclarsson:
Ok, this removes `3rdparty/partFct`, adds `3rdparty/TIPS/`, and comments how to use the latter.

(I hope this is everything now, because my messing up a save has now messed up my branch completely since `stash`+`stash pop` did something weird and required a commit to work again...)

3rdparty/TIPS/download_tips.py
```diff
@@ -0,0 +1,12 @@
...
+#
+# Download TIPS 2021 data from Zenodo.
+#
+# Sometimes the download fails for no apparent reason. If that happens, wait
+# a few seconds and rerun the download script. It will continue from the point
+# where the error occurred.
+#
+# Needs the zenodo_get module: pip install --user zenodo_get
+
+python -m zenodo_get -R 5 -m -o TIPS2021 10.5281/zenodo.4708098
```
riclarsson, line 11: @olemke This seems to create a TIPS2021 folder locally?  Is it possible to add also a local `.gitignore` to ignore this folder?

riclarsson (commented):
It seems you have added a .sh file with a .py extension.

3rdparty/TIPS/download_tips.py
olemke: Oops, it started as python script then ended up as a bash script. Forgot to change the extensions.
I added the download folder to `.gitignore`

olemke:
@riclarsson Thanks for adding the python script. It's important that we have that included to be able to regenerate the partfun data if needed.

I will take care of the necessary modifications to the planet isotopologue ratios files in arts-xml-data and will update them after this PR is merged. Just to confirm: The Perrin catalog files don't have to be modified for the renamed species because that is handled automatically when the catalog is read, correct?

@stefanbuehler Are there any open issues/questions to resolve before merging this?

riclarsson:
@olemke The Perrin catalog should be OK.  It would probably be beneficial to update it.  But I think all of the species are in the `update_isot_name`-function that's supposed to take care of this.

Thanks for dealing with the iso-ratios.  Also remember to remove "spectroscopy/PartitionSums", it's not valid anymore.

Also thanks for the download script.  I was sitting there manually downloading the files before, thinking people must be crazy for publishing on that platform.

(That said, I think it is time to cut our losses and remove the ARTSCAT-4 format from arts-xml-data.  It will lose some quantum numbers but that's all if converted to the new format.  I don't think anyone is using these catalogs for their missing quantum numbers anyways, because then ARTS would be updated for it.  I would like to be able to somehow formally deprecate `ReadARTSCAT`, though that would require changes to MdRecord and the auto-generated wrapper functions to actually output deprecation warnings...)

olemke:
> @olemke The Perrin catalog should be OK. It would probably be beneficial to update it. But I think all of the species are in the `update_isot_name`-function that's supposed to take care of this.

We can then postpone that until we convert them to the newer format as you proposed below and get the updated names for free in the conversion process.

> Thanks for dealing with the iso-ratios. Also remember to remove "spectroscopy/PartitionSums", it's not valid anymore.

I'll remove that directory as well.

> Also thanks for the download script. I was sitting there manually downloading the files before, thinking people must be crazy for publishing on that platform.

You had more patience than me then. :-) Don't know why Zenodo doesn't supply a download all link. But since they provide many very large datasets, they probably want to prevent mass downloads. It would've been better by the authors of the TIPS data to upload the files as a zip archive instead of single files which would've avoided that problem. In general, Zenodo is a great platform to share data and acquire a DOI very easily.

> (That said, I think it is time to cut our losses and remove the ARTSCAT-4 format from arts-xml-data. It will lose some quantum numbers but that's all if converted to the new format. I don't think anyone is using these catalogs for their missing quantum numbers anyways, because then ARTS would be updated for it. I would like to be able to somehow formally deprecate `ReadARTSCAT`, though that would require changes to MdRecord and the auto-generated wrapper functions to actually output deprecation warnings...)

Definitely something we should consider.

olemke:
Stefan is very busy these days so I talked to him briefly in the group meeting yesterday whether he has anything to consider before merging this.

He proposed to have a look at the builtin isotopologue ratios and see if any differ from the latest Hitran version. Any differences are probably only for historical reasons and it would be better to update the builtin ones. I did a diff between the iso ratios in `isotopologues.h` and `hitran_species.cc` but couldn't spot any other differences than the typo I corrected earlier. Are you aware of any that I might have missed?

The other thing was updating the user guide, which you did.

I see nothing else preventing the merge.

riclarsson:
Hi Oliver,

I attach a patch that adds `isotopologue_ratiosInitFromHitran`.  With this patch applied, you can run:

```python
import pyarts
ws = pyarts.workspace.Workspace()
ws.isotopologue_ratiosInitFromBuiltin()
ws.isotopologue_ratios.value.savexml("builtin.xml")
ws.isotopologue_ratiosInitFromHitran()
ws.isotopologue_ratios.value.savexml("hitran.xml")
```

I see that there's at least two more species: 

CO2-737
COCl2-2657

that are different between Hitran and ARTS (old and new).

olemke:
Thanks @riclarsson. That is much more convenient than diffing the source files. Unfortunately it appears the patch got lost. Maybe just mail it directly to me.

olemke:
While running the extended test set (check-all-controlfiles), I found some tests that fail in this branch, but not in master:

These three tests fail the comparison to the reference values:
```
ctest -V -R ^arts.ctlfile.slow.instruments.odinsmr.TestOdinSMR_1D

138: Run-time error in method: Compare
138: y-yREFERENCE FAILED!
138: Max allowed deviation set to: 0.01
138: but the vectors deviate with: -0.0605066
```
```
ctest -V -R ^arts.ctlfile.xmldata.artscomponents.cia.TestRTwithCIA

172: Run-time error in method: Compare
172: iy-iy_reference FAILED!
172: Brightness temperature with CIA N2 continuum differs from reference calculation
172: Max allowed deviation set to : 1e-08
172: but the matrices deviate with: 0.0170381
```
```
ctest -V -R ^arts.ctlfile.xmldata.artscomponents.arts-xml-data.TestPlanetIsoRatios

Run-time error in method: Compare
yREFERENCE-y FAILED!
Max allowed deviation set to: 2e-09
but the vectors deviate with: 0.00941998
```

Which brings up the usual question, are our comparisons too strict or are these significant enough to indicate a problem?

Another problem showed for controlfiles that use non line-based species such as "free_electrons" or "liquidcloud". `ReadSplitARTSCAT` tries to open files for them:
```
ctest -V -R ^arts.ctlfile.planettoolbox.planetary_toolbox.demos.passive.DemoVenus_fullRT_1D_clearsky

Run-time error in method: ReadSplitARTSCAT
User Error: not ignore_missing
Error is found at:
        /Users/olemke/Hacking/uhh/arts/src/m_absorptionlines.cc:386
Please follow these instructions to correct your error:
Errors in calls by *propmat_clearskyAddZeeman*:
Cannot find input file: spectroscopy/Perrin/free_electrons.xml
```

This can of course be worked around by passing `ignore_missing=1` to the reading routine, but don't think that's how it's intended to work. The reading routine is supposed to automatically ignore those species, right?

olemke:
Thanks for the patch @riclarsson . Works perfectly.
I agree that tests should be more specific to what they're targeting. I'll have a look at the two in question.

PR #357, opened 2021-06-10
Make manual mirroring mirror ARTS v2.3.1277 behavior

riclarsson:
Manual mirroring now behaves the same as in older versions before the line shape shifts.  I managed to mess up the direction of the shifts because those weren't required to be mirrored in old ARTS.

After creating `abs_lines`, running the new `abs_linesMakeManualMirroring` once should be enough to copy the old behavior of running `abs_lines_per_speciesAddMirrorLines`.  Perhaps some names should be updated.

@stuartfox , can you check that this works for you?  I ran your two tests `test_v2.3.1277.arts `and `test_v2.5.arts` with your default settings and with the two function calls above to check the difference and the versions seem to be close enough now.

### DIFF
--- a/src/absorptionlines.cc
+++ b/src/absorptionlines.cc
@@ -3379,7 +3379,7 @@ Numeric Lines::F_mean(const ConstVectorView wgts) const noexcept {
 Numeric Lines::CutoffFreq(size_t k, Numeric shift) const noexcept {
   switch(mcutoff) {
     case CutoffType::ByLine:
-      return F0(k) + shift + mcutofffreq;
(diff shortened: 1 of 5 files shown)


stuartfox:
@riclarsson I can confirm that this change lets me reproduce the mirroring behaviour from v2.3 both with and without a line cutoff and in the presence of pressure shifts. The only question I have is how it should be used if one only wants to add mirror lines for certain species (e.g. H2O, but not O3)? Do you need to read the H2O first, add the manual mirroring, and then add the other species? Would it be possible to pass a species tag(s) to abs_linesMakeManualMirroring or abs_lines_per_speciesMakeManualMirroring?

PR #358, opened 2021-06-10
Add abs_lines_per_speciesMakeManualMirroringSpecies

riclarsson:
@stuartfox After this merge, you will be able to do

```
abs_lines_per_speciesMakeManualMirroringSpecies(species="H2O")
```

The string input as a species must be the full string you use to define a species tag group.  So if your full definition is something like:

```
abs_speciesSet(species=["H2O,H2O-SelfContStandardType", "O3", "Some", "Other,Species"])
```

then you have to write something like:

```
abs_lines_per_speciesMakeManualMirroringSpecies(species="H2O,H2O-SelfContStandardType")
```

(Spaces are still optional)


stuartfox:
Thanks for doing this @riclarsson. I'm always impressed how quickly the ARTS developers respond to my requests!

PR #361, opened 2021-06-29
Linemixing

riclarsson:
This adds quite a few things.  I think my reimplementation of the oxygen line mixing works now.  There's also a lot of work on CO2 line mixing in here, and the basics of what would be needed to make use of it from scratch is in place.

From a user perspective, the input for the line mixing model now comes directly from the controlfile in the form of the `ecs_data` map [Error-Corrected-Sudden data].  I say map because I needed to have this per-band/isotopologue and then per-species, so it's a new type of type for ARTS users.  To help generate it, two fixed methods, one dynamic method, and an initializer are provided.

From a scientific point of view, a novel way to get at some of the line mixing parameters that people tend to want have been added.  I had a lot of problems with this.  It turns out the normalizers in ARTS didn't really fulfill what I needed.  So I have also added a simple frequency normalizer to the absorption lines that follows more closely how I imagine the line strength should behave away from the line center.  Lastly, I added a polynominal temperature scheme for the line shape parameters.  Mostly because it was easier to fit towards than the existing exponential versions.

For debug builds it is now possible to set axis-names and line names and stuff like that in the plotting-tool.  I had at one time too many lines and got confused what I was debugging so it got required.


src/linemixing.cc
```diff
@@ -385,10 +710,15 @@ ComplexMatrix single_species_ecs_relaxation_matrix(const AbsorptionLines& band,
...
-      Makarov2020etal::relaxation_matrix_offdiagonal<param>(W.imag(), T, band, sorting, band.SpeciesMass(), species_mass);
+      Makarov2020etal::relaxation_matrix_offdiagonal(W.imag(), band, sorting, species_ecs_data, T);
       break;
+    case PopulationType::ByRovibLinearDipoleLineMixing: {
+      std::cerr << "THIS CODE IS NOT TESTED!\n";
+      LinearRovibErrorCorrectedSudden::relaxation_matrix_offdiagonal(W.imag(), band, sorting, species_ecs_data, T);
+    } break;
     default:
-      ARTS_ASSERT (false, "Bad type");
+      ARTS_ASSERT ("Bad type, we don't support band population type: ", band.Population(),
```
olemke, line 720: Assertion is missing a proper condition. Using the string as a condition will always evaluate to true and never trigger the assertion.

riclarsson: Ok, I fixed it.  I am still working on the CO2 internal solver so there's a few more things changing still.  But the O2 code is good now.

olemke:
> I am still working on the CO2 internal solver so there's a few more things changing still. But the O2 code is good now.

I marked this PR as a draft, assuming the CO2 updates should also go into this branch before merging.

riclarsson:
If this passes the tests, it should now be merged.

The internal CO2 calculations don't work as they should.  I think it is a very small error in the implemented theory.  The theory is as referenced in the code, but the O2 calculations now work because they don't follow theory but instead they follow the implemented code by the authors of the papers I reference.  So I think it is too optimistic to think that this would work for CO2 without hiccups without actually looking at all the details one extra time.

I want this merged so that I can reference the CO2 problem easily when emailing people doing CO2 line mixing.  Since the O2-stuff works, it is easier to pin point where the problems could be located.

PR #365, opened 2021-07-13
Minor updates before holidays

riclarsson:
This can be merged as is, I think.

The main thing here is deprecating add-lines.

I need this update merged before October since my old email-address will be invalid shortly.


olemke:
@riclarsson Updating the lookup table calculation is next on my todo list. Have a great vacation!

riclarsson:
@olemke If you need any input from me let me know.  I guess it should be straightforward but I am not sure.

I'm technically on vacation from Tuesday next week until I start in Hamburg in October, but I will check my gmail somewhat regularly.  I had too many days saved up...

olemke:
@riclarsson Will do if I hit any roadblocks.

Wow, you saved up a lot. But it's also nice to have a longer break before starting something "new". :-)

Issue #368, opened 2021-09-03
[PyARTS] Exception in Python agenda does not stop execution


olemke:
**Describe the bug**
If an exception occurs in an agenda defined with the `@arts_agenda` decorator, an error is printed but execution continues.

**To Reproduce**
Example:
```python
from pyarts.workspace import Workspace, arts_agenda

ws = Workspace(verbosity=2)

@arts_agenda
def my_agenda(ws):
    ws.UnknownMethod()

ws.test_agenda = my_agenda
ws.AgendaExecute(ws.test_agenda)
# Execution should not continue beyond here
print("This should not be reached")
```

Output:
```
- AgendaExecute
--- Logging error ---
Traceback (most recent call last):
  File "/Users/olemke/Hacking/uhh/arts/build/python/pyarts/workspace/workspace.py", line 140, in callback
    eval(compile(m , "<unknown>", 'exec'), context)
  File "<unknown>", line 3, in <module>
  File "/Users/olemke/Hacking/uhh/arts/build/python/pyarts/workspace/workspace.py", line 486, in __getattr__
    raise AttributeError("No workspace variable " + str(name) + " found.")
AttributeError: No workspace variable UnknownMethod found.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/olemke/miniforge3-arm/lib/python3.9/logging/__init__.py", line 1083, in emit
    msg = self.format(record)
  File "/Users/olemke/miniforge3-arm/lib/python3.9/logging/__init__.py", line 927, in format
    return fmt.format(record)
  File "/Users/olemke/miniforge3-arm/lib/python3.9/logging/__init__.py", line 663, in format
    record.message = record.getMessage()
  File "/Users/olemke/miniforge3-arm/lib/python3.9/logging/__init__.py", line 367, in getMessage
    msg = msg % self.args
TypeError: not all arguments converted during string formatting
Call stack:
  File "/Users/olemke/Hacking/uhh/testing/pyarts_agenda_nonexistent/test_agenda_invalid_method.py", line 10, in <module>
    ws.AgendaExecute(ws.test_agenda)
  File "/Users/olemke/Hacking/uhh/arts/build/python/pyarts/workspace/workspace.py", line 276, in __call__
    self.m.call(self.ws, *args, **kwargs)
  File "/Users/olemke/Hacking/uhh/arts/build/python/pyarts/workspace/methods.py", line 445, in call
[...]

PR #369, opened 2021-09-03
Add 'enable_callbacks' keyword to arts_agenda decorator.

simonpf:
This addresses #368 by adding a keyword ``allow_callbacks`` to
the decorator.

````
# Fails at parse time.
@arts_agenda
def my_agenda(ws):
    ws.UnknownMethod()
````

The above now fails when the code is parsed. 

To allow Python callbacks the user  needs to set the ``allow_callbacks`` keyword to ``True``. This has the disadvantage that errors like  below are encountered first at runtime.  Ideally, there should be some error handling on the ARTS side for the callbacks, but I don't want to dive into the pointer hell of the callback mechanism at this point.

````
# Silently fails at runtime.
@arts_agenda(allow_callbacks=True)
def my_agenda(ws):
    ws.UnknownMethod()
````


olemke:
Nice solution using a control keyword.

PR #375, opened 2021-09-22
Bug fix to MT-CKD v3.2 H2O foreign continuum

stuartfox:
There is currently an error in the indexing for the correction factor XFAC_RHU in the H2O Foreign continuum MT-CKD v3.2 implementation in ARTS. This is because the C++ code closely follows the FORTRAN code from AER, but there is a difference between how the arrays are dimensioned between the two codes. The AER FORTRAN code declares XFAC_RHU to have dimension -1, 61 whereas the C++ code uses the default dimensions 0 to 62. Since the two codes extract the same index J from the array there is an unwanted offset in the C++ version.

This PR corrects this error. It also changes some of the indexing in the MTCKD v3.2 H2O self and foreign continua (sections SL260/296 and FR296) to match the conventions used by the AER code more closely, although this has no impact on the results.

The revised code gives a closer match to the foreign continuum absorption between ARTS and the official AER code.


olemke:
Thanks @stuartfox !

PR #380, opened 2021-11-02
corrected the documentation and adapted to the new angle convention

j-petersen:


riclarsson:
I like the update but want to make a small comment about style here.  I'm not saying you should change the code because this is more of a generic style comment.

If you want the code to be 'easier' to read, you can put `using Conversion::deg2rad, Conversion::cosd, Conversion::sind;` at the top of the `scat_angle` function (just after the opening `{`), or even `using namespace Conversion;`.  This way, you can remove all occurrences of `Conversion::` in the code.

We've actually never discussed how this should be done and what style to use here, so @olemke  might have other opinions on when and where we should make use of `using ...`.

The change would be that
```
acos(Conversion::cosd(za_sca) * Conversion::cosd(za_inc) +
     Conversion::sind(za_sca) * Conversion::sind(za_inc) * 
     Conversion::cosd(aa_sca - aa_inc));
```
looks like
```
acos(cosd(za_sca) * cosd(za_inc) +
     sind(za_sca) * sind(za_inc) * 
     cosd(aa_sca - aa_inc));
```

Which is less clutter but it's perhaps not as clear that you are doing implicit conversions.

olemke:
I'm very much in favour of making the symbols you actually use explicitly available with `using Conversion::deg2rad` etc.
Importing whole namespaces with `using namespace ...` should be avoided (I'm aware we're still doing that in some places in ARTS), for the same reasons `import * from ...` in python is a no-no because it's messy.

riclarsson:
As a piece of review that I think you should update since you are looking at this code.

I think you should add to the documentation of the function at line 2072:
```cpp
\return Scattering angle [rad]
```
and remove
```cpp
\param[out] theta_rad    Scattering angle [rad].
```

In doxygen documentation, all the `\param` statements refers to input, output by reference, or input and output by reference.  This function uses pure value semantics (which is actually better) so it should be saying this by writing `\return`.

Also, I would move the function itself to the top of the file and remove it from the header.  As far as I can see it's only ever used in the same file.  There's no reason to expose it to the public namespace, and also there are some optimisations by the compilers that _might_ happen if you keep functions as local as possible.  It probably won't make any difference here, but exposing the function also doesn't help.

j-petersen:
I see your point. I thought of making it clear that this is a conversion and not some sort of cast as the d for double in `cosd` might suggest. 

Should we discuss which style to use in the next arts dev meeting and leave the pull request open until then?

riclarsson:
> I'm very much in favour of making the symbols you actually use explicitly available with `using Conversion::deg2rad` etc. Importing whole namespaces with `using namespace ...` should be avoided (I'm aware we're still doing that in some places in ARTS), for the same reasons `import * from ...` in python is a no-no because it's messy.

It should be fine inside the function to include all.  But the more verbose "import what you need" style is probably better.

Have you tried removing the using namespace std from Arts?  It's quite funny the amount of errors you get.  A very nice rabbit hole.  I wonder if there's at tool to fix it.

olemke:
@j-petersen I don't think we need a discussion on this before moving forward. We should mention it at the next meeting as a recommendation.

olemke:
> > I'm very much in favour of making the symbols you actually use explicitly available with `using Conversion::deg2rad` etc. Importing whole namespaces with `using namespace ...` should be avoided (I'm aware we're still doing that in some places in ARTS), for the same reasons `import * from ...` in python is a no-no because it's messy.
> 
> It should be fine inside the function to include all. But the more verbose "import what you need" style is probably better.

Yes, good point. And it keeps the function code more concise. 

> Have you tried removing the using namespace std from Arts? It's quite funny the amount of errors you get. A very nice rabbit hole. I wonder if there's at tool to fix it.

Never dared to. ;-)

j-petersen:
> In doxygen documentation, all the `\param` statements refers to input, output by reference, or input and output by reference. This function uses pure value semantics (which is actually better) so it should be saying this by writing `\return`.
> 
> Also, I would move the function itself to the top of the file and remove it from the header. As far as I can see it's only ever used in the same file. There's no reason to expose it to the public namespace, and also there are some optimisations by the compilers that _might_ happen if you keep functions as local as possible. It probably won't make any difference here, but exposing the function also doesn't help.

Okay, I'll change the return statement, but I'll keep the function where it is since i want to use it at a different place. That's the reason I looked into it in the first place.

riclarsson:
I am actually wondering how similar this is to my Zeeman effect function at line 227 in zeemandata.cc.  It seems we have two different solutions here.  My function is a bit more general since it doesn't think the incoming vector is normalized, and I need several derivatives, but I cannot help but wonder if we are reproducing the same code over-and-over again.

PR #381, opened 2021-11-04
Apply VMR scaling in propmat_clearsky_fieldCalc and extend test to actually test things

riclarsson:
This should address the issue you found @stuartfox .  I think.  I don't actually understand how it could be working with complete models.

Additionally, I am modifying a test to at least make relative value checks against the target.


stuartfox:
@riclarsson  this has mostly fixed things with the lines, but broken the complete models. Also, it looks like the lines in the propmat_clearsky_fieldCalc aren't getting their forefactor (I'm using VVW) applied. Let me know if you need me to produce an example to demonstrate the issues I'm seeing.

riclarsson:
@stuartfox A test would be very useful.  I will try to set one up myself after lunch but your test would be a good way to confirm if I failed or not to fix the issue.

stuartfox:
@riclarsson I'm assuming you don't need me to put together a test any more as you seem to have sorted this out now?

riclarsson:
@stuartfox I don't need one.  Could you confirm it works on your end now?

stuartfox:
@riclarsson Yes - the results from propmat_clearsky_fieldCalc now look like what I expected.

PR #384, opened 2021-11-08
Mtckd350 h2o

stuartfox:
This PR adds v3.5 of the MT-CKD water vapour continuum from AER to the available options in ARTS


riclarsson:
Hi Stuart,

This is very nice to have.  Thank you for taking the time!

It will take us some time still to merge this.  Sorry!  We had an ARTS meeting last month where we discussed the status of the continua-family of functions.  We had a decision that the "legacy_continua.cc" files should be frozen and new functionality added to their own, separate functions.  The entire infrastructure of this "legacy_continua.cc", as you might have noticed, is not really good as it stands.

You don't have to do anything.  I will separate this new function to the new format and commit it to your branch.  I just wanted to let you know the reason for this delay.

stuartfox:
HI @riclarsson ,

No problem - legacy_continua.cc is a monstrous piece of ugly code so I don't blame you for wanting to reform it. There's a lot of commonality between the different models that are just copy-and-pastes and repeated code, although there are also some more subtle differences which make it a bit more difficult to handle everything in a common way.

I'd be interested to understand your new approach for defining continua and "complete absorption models" as I'm likely to need to configure ARTS to follow the latest iteration of Phil Rosenkranz's absorption routines for a current project. I seem to remember that I might have seen an "MPM2020" in a recent bit of code so maybe this is something that you've done already?

riclarsson:
@stuartfox This should now work within the more modern interface of the continua models.  I will walk through what's done here and how to access things.  Please let me know if anything is unclear.

The only two changes you should see from a "user" point of view is that you should no longer add the description of the models in continua.arts-style, AND you need to add `propmat_clearskyAddPredefined` to your propagation matrix agenda.  If you see anything more than that changing, then I've made some sort of mistake.

Basically, the modern interface separates the implementation of these 'predefined' models from Arts core functionality and builds them as their own library (atop only matpack, the Vector/Matrix package).  You can find the current implementations in src/predefined/.  Your CKDMT350 implementations are both in the CKDMT350.cc file.  Please have a look that you can understand them with the changes I had to make for this separation to fit with future ARTS development.

What's very important here is that I changed this from operating on the cross-sections to operate on the absorption.  Absorption in ARTS is done via the `PropagationMatrix` class.  Your model and all the models in the legacy continua only access the unpolarized part of this matrix.  Other than that, the absorption in ARTS is also computed fully on a single level.  So no more vectors of pressures and so on but a single scalar.  I made this change because we will soon scrap the xsec-agenda anyways, so there would be no point to write a new interface using it.

I also left the documentation with only your name as an author there.  I am not convinced that my changes warrants attribution, but it is therefore very important to me that you accept this and can understand the code.

(I am still going to push one more commit to deal with Jacobian better, so expect some small changes still before this is merged.)

riclarsson:
Hi Stuart, it took more time than I wanted to fix this as I got stuck in a loop until I realized that ARTS never had a physically good definition of derivatives in the old implementation of them, and so that the error I saw was justified since it mimicked this poor definition... anyways this is not the place to go more detailed in this rant.

As long as this runs through the checks (which I hope it does) then this is the final commit from me here about moving it into a modern interface.  This can then be merged as soon as you give your approval, @stuartfox.

stuartfox:
Hi @riclarsson, it looks like your changes lead to the same results in my tests so I'm happy for this to be merged.

PR #386, opened 2021-11-16
Artscat

riclarsson:
Makes ARTS compatible with latest Hitran version and adds and modifies several parts related to species manipulation.

This is necessary to merge for us to be able to rebase Artscat to Hitran 2020.

Two notes:

1) Hitran and Tips no longer agree on isotopologue naming schemes.  This can potentially create issues in the future.  I hope my interpretation of the names are correct (I believe they are, but cannot be sure).

2) Hitran has changed the numerical value of their assumed isotopologue ratios for several species.  This means it is technically no longer valid to read old versions of Hitran.  I am unsure if this is important enough to keep around?  This can be seen in src/hitran_species.cc.  Should we keep backups of all the old versions of Hitran somewhere?  I no longer know how the reproduce the data from before  the 2016 version (which is the numbers we've got here).  So it is possible that several versions here are not Ok.  Note, of course, that as long as Artscat is being used, the user will not encounter these issues.


stefanbuehler:
Hi Richard,

you are right. But I think it is enough to point this out in the documentation. From my perspective, it is not necessary that we support much older versions of HITRAN. Are the HITRAN ratios stored in a file now? Then we should give a version number to that file. We could then quite easily generate a correct file if we really need to read older data.

Or, what do you think?

Stefan

riclarsson:
The Hitran ratios are compiled into Arts.  They are part of the `hitran_species.cc` file.

We could separate them.  This would take quite some effort.  I think it is possible but we would need to keep a lot more data around and then process these to generate the code, or force the user to read in the data manually.  The former means more steps of compilation, while the latter means passing user-space variables down to file-manipulating functions AND that we need some specific `HitranMolecularParamaters` class that requires all of the user-facing class specific things.  I don't like either of these approaches.

Another approach is to just keep multiplying the number of data-maps we compile in the above `.cc` file.  For now this is OK but I don't know how this will be in the future (i.e., I believe it will be difficult to maintain).

I actually like the idea that we only ever support one version of Hitran the most.  We can then compile this in without worrying about anything else.  The user-facing approach should be to simply put the version of Arts that supports what version of Hitran into the `ReadHITRAN` method.  Then the user could just get that version of Arts, read Hitran and save it to xml-files, and then reload the latest version of Arts and use all the fancy new features as they please.  The maintainability of this is quite easy so-long-as people update the method description.

stefanbuehler:
Yes, I think this is an okay solution.

riclarsson:
@olemke I re-did the order of the Hitran catalogs after the `test_*` files changes so people won't be confused about "arts" or "pyarts" not compiling.  (They work but I am sure many people don't actually specify a target.)

Can you have a quick look through the other changes?  Some of them felt hacky but I am not sure.  After you are done, this branch can be merged.  It is then possible to use Hitran2020 in Arts.

olemke (approved):

PR #391, opened 2022-01-18
Fix reading python routine for Array<X> types in Arts

riclarsson:
Add update() and update_directory() to pyarts.xml


riclarsson:
Note: I am not sure how to write to a temporary file in update.  If there is a failure in saving but not loading, then update() may now clobber and destroy the file

riclarsson:
@olemke Please review the python that it is acceptable.  I don't trust my python being pythonic.  If it is acceptable, please merge it :)

olemke (approved):

PR #397, opened 2022-02-14
Make the loop parallel

riclarsson:
There is some discussion on Arts currently being slow for some loopup table generations.  This is an attempt to fix that.

@olemke, could you review and see if it is OK?


olemke (approved):
Looks good and gives a decent speedup.

PR #399, opened 2022-02-18
Move things away from the controlfiles*/

riclarsson:
We want to provide "tests" and "examples". rather than the current combined setup.  This is to clearly separate tests that are run to stop degradation of code from tests that demonstrates Arts usability.

There are several important things to consider here (and I cannot do them all myself).  I will list these below and ask questions that I will need input on as I go along.

0) This is still an incomplete/draft PR.  I am putting it as a PR rather than as a issue, as the latter does not contain code.
1) Both tests and examples should run as part of the continuous integration of Arts.  This means the tests cannot be slow.  It also means it is somewhat problematic if the tests depends on arts-{xml,cat}-data, though as long as the tests are fast I think the latter should be OK?  Are we OK with not allowing any slow tests anymore (and should the final state of this PR be that we remove that as an option)?  All the current slow tests would have to either be dropped or adapted to a simpler state.
2) We will not allow the existence of any .arts or .py file that is not run as part of these sub-folders.  How do we ensure that ALL .py and .arts files that are part of this are run?  I think a separate test that scans the directories and the CMakeLists.txt for files is required. test_tests.py and test_examples.py, and then also run these files as part of the test-suite.
3) What folder structure should we use internally?  I think that free-form structure without any central compare-data or load-data repository should be used.  On the top-level of both tests/ and examples/ we can split it into "arts" and "pyarts" (or "python"?).  I don't like the centralized testdata/ folder of today as I have no idea if all of the files are being used.
4) I don't know what's an example and what's a test today.  I need clear input from everyone involved in Arts controlfiles (pinging @stefanbuehler , @olemke , @simonpf , @erikssonpatrick , @m-brath to have the currently active key people aware of this).  What controlfile goes where?  I would like an initial list of the controlfiles that everyone feels strongly about and have them defined as "tests" or "examples".  As a major problem for this, I don't even think our clearsky stuff are good examples, since they don't perform normal clearsky calculations but rather demonstrates how you shift between 1D to 3D.
5)  If you have new examples that you think should replace something old, either email me the file or make a PR to my branch and explain why the old one goes away and the new one is better.  @m-brath , you said the lectures contains something that maybe can be adapted to serve as good examples?
6) This should be allowed to take some time but I think it is important it is done before Arts-3.
7) Anything important that I missed?


riclarsson:
@olemke I forgot: I only moved the general files for now.  How do I set the default include path of Arts so that these general files are pointed towards properly?  (also to point out the severity of the rot of the controlfiles/, one of the general/ includes did not even run...)

olemke:
@riclarsson `ARTS_DEFAULT_INCLUDE_DIR` in the toplevel `CMakeLists.txt` is used to set that. However, there were some issues after changing that. I've committed a fix with better path handling in [9f6b00a](https://github.com/atmtools/arts/pull/399/commits/9f6b00a89c9fc02848aed970f3797973e5c4453e).

Issue #311, opened 2021-02-17
pyarts and unsupported classes


riclarsson:
**Is your enhancement request related to a problem? Please describe.**
When running `pyarts`, there are several classes that are not supported by value return from the workspace variable.  Specifically:

```python
import pyarts
ws = pyarts.workspace.Workspace()
pyarts.xml.save(ws.sensor_time, "test.xml") # or pyarts.xml.save(ws.sensor_time.value, "test.xml")
```

Does not work.  Instead you have to do:
```python
import pyarts
ws = pyarts.workspace.Workspace()
pyarts.classes.from_workspace(ws.sensor_time).savexml("test.xml")
```

**Describe the solution you'd like**
I would like the first option to work.  Both the active and out-commented versions.  The out-commented version is the only one that works today, and then only for some types.

**Describe alternatives you've considered**
There are two steps needed to make this happen that I can see:
1) Make `pyarts.xml.save` accept `WorkspaceVariable` and make it wrap the behavior above for this type.
2) Make `WorkspaceVariable::value` return `pyarts.classes.from_workspace(ws.VAR)` when `VAR` type is not explicitly supported by other `WorkspaceVariable::value` methods.  Now, `pyarts.xml.save` needs to know if the method called has its own complete `type(ws.VAR.value)::savexml` method.  If it does, prefer it to the previous mapped version of XML writing.

The `savexml` methods in `classes` all wraps the ARTS versions.  This makes them preferred for writing XML files to the re-written python versions.  I would prefer if `WorkspaceVariable::value` always returns `pyarts.classes.from_workspace(ws.VAR)`, but this might break a few scripts?  I don't really know what would break.  `ArrayOfIndex`?

There are probably better versions here but I don't know what they would be.

**Are you willing to help adding this enhancement?**
No.  I don't understand all the attributes and all that of python.  This makes me a bad candidate to implement these changes.

**Additional context**
The fact that some `WorkspaceVariable::value` methods doesn't work has led to user questions and I think it will lead to more.  The pyarts.classes methods will support all this because the github build process will fail if ARTS developers don't add

simonpf:
Totally see your point and this should not to be too difficult to achieve. However, I am currently working in a different direction and planning to refactor the Python interface prior to the 3.0 release.

[...]

PR #400, opened 2022-02-23
Pybind11

riclarsson:
This draft would replace the current C-API for python bindings with pybind11 generated ones.

pybind11 is an open source project to automatically generate python bindings.  I've used it to generate bindings that are currently in python in C++ code that is exposed to python via the pybind interface.

I am making this part of the Arts distribution for the same reason we have Eigen in there.  I am not sure these headers are 100% stable across multiple distributions, so it must be part of our distribution.  I am not sure we need their entire test-suite in here, though, so picking and choosing what to remove is important.  All I know is that their CMakeLists.txt somehow generates a CMake command we need.  Other than that I think it is a full 'headers only' library.

The main goal to have this code is to significantly simplify the generation of the class-bindings that are required to various Arts types.  This ought to work mostly as expected for all Arts types, though several features of the old classes are still missing.

The advantages are that everything should now be type safe.  Also, the full Arts documentation is exposed for workspace variables, methods, and agendas.  Also, you should be able to reliable set more Arts classes from something that is equivalent to their types in python (e.g., lists should be able to set most Array<T>s).  There are also a few more minor changes like the ability to set workspace variables from their correct type without first creating them independently, making it easier to use.  Lastly, having this interface makes it extremely easy to expose internal Arts functions if need be (e.g., if you have a working python prototype).

The disadvantages are that some features are missing in some classes.  I am not expecting all the python tests to work in this initial DPR.  Another disadvantage is that we lose control over some of the error handling (e.g., if you cannot generate an Arts type from a python object).  A third disadvantage is that we take a lot longer to compile.

I have not measured runtime performance.  I don't think it will be much worse or better than before.  I have also not updated the documentation.  Lastly, I have not tested if the interface is leaky.  It's a well-used package, but it is definitely possible I have messed up in how I use it somewhere...

This requires deep review before it is possible to merge.  Mainly, @simonpf and @olemke , I need your input on if it works and/or some key components should be fixed as they are bad, or if the loss of some features is acceptable for now.


riclarsson:
@simonpf I am aware that you are extremely busy at the moment, so I am not expecting you to take too much time to look at this.  I went with pybind11 here because you suggested it in #311 .  I think it is a good enough solution I have here, but there might be several holes I've stepped in once again.  I've tried to follow your API as much as I can here and also to reuse (though sometimes just partly) as much code and ideas as I could.

simonpf:
Thanks @riclarsson, I guess this could take ARTS Python interoperability to the next level. Will have a proper look on Friday.

python/pyarts/workspace/workspace.py
```diff
@@ -143,11 +109,12 @@ def parse_function(func, allow_callbacks):
         arg_name = func_ast.args.args[0].arg
     except:
         raise Exception("Agenda definition needs workspace arguments.")
-
-    ws = Workspace(0)
-
+    
     context = copy(func.__globals__)
-    context.update({arg_name : ws})
+    assert isinstance(context.get(arg_name), Workspace), \
```
simonpf, line 114: Why is this required? Requiring the function argument to correspond to an existing variable is a pretty ugly clash with the Python syntax.

simonpf:
Let me start out by saying that I think that this is a great effort and that I see great  potential to make ARTS even more versatile. While the old interface was essentially just a way to use ARTS through Python this one could really unlock a lot of the functionality that is in ARTS but currently not efficiently accessible through any of the available interfaces.

I have compiled the code and tried to get my old tests to run, which covered some of the functionality that I consider essential. However, I didn't get very far. I think an issue that we are facing with lowering all the interface own to C++ is that it isn't as easy to work around built-in limitations of ARTS, which I used to do on the Python side of the interface. So I think to making this interface fully functional will require some changes to the behavior of the ARTS runtime.

Some issues that I encountered so far:
- You can't parse an agenda twice without encountering an error. Note that this is pretty annoying for interactive use.
- Same thing for creating a variable twice.
- See comment above regarding definition of agendas.
- WSV names are currently global. I haven't yet worked my way through how you handle creation of temporary values for GINs but that tended to cause trouble. See also my comment on #387. So this may be something to consider here.

So I don't really know what to say here. This is a pretty big commitment and there's not really a point in half-assing it. I am happy to help as much as I can if the ARTS developers commit to it, but I am not willing to put work into an effort that dies up after three months.

python/pyarts/workspace/workspace.py
riclarsson: I am also not really happy about this.

It's necessary to have a `Workspace` to create workspace variables and callback functions onto it.

You created a separate `InteractiveWorkspace` instance to deal with this problem.  I also need such an interface to the `Workspace` to get the workspace variables that are defined (and define new ones).

We have a mid-to-long term goal to make the `Workspace` own its entire context, so keeping your solution will not really be possible long-term.  I am therefore not really happy about your solution going forward.

Would you find it acceptable to have the arts_agenda take another input that contains the current workspace?  The check there would then rather read something like

```python
if global_workspace is None: 
    global_workspace = context.get(arg_name)
assert isinstance(global_workspace, Workspace), / ...
```

Then I would simply use `global_workspace` in the rest of the function.

riclarsson:
Are your tests part of the pyarts tests?  I would like to see if I can fix the issues directly.

I'll assume you mean parsing+executing controlfiles that creates workspace variables.  I'll have a look to see how difficult the change is to allow double 'create' in Arts's parser.

I don't do anything special with GIN values.  If a function that takes GIN is called, the GIN will be allocated on the workspace as `::FUNCTIONNAME::VARIABLENAME`.  Callback functions are also allocated on the `Workspace` as `::callback::STATIC_COUNTER` and user-defined input are allocated to the `Workspace` as `::anon::ANOTHER_STATIC_COUNTER`.  I don't de-allocate any of these yet (I am still working on that).

There is no point in merging this before it works as expected.  There are a few things that might have to change.  I don't see this dying in three months.

riclarsson:
@olemke If you add `.def(py::init([](Numeric x) {return Vector{x};}))` and `py::implicitly_convertible<Numeric, Vector>();` at appropriate lines in `py_matpack.cpp` then the `ws.f_grid = 1.` behavior is reproduced.  However, if you also add `py::implicitly_convertible<Index, Vector>();`, you are on your own if `ws.f_grid = 1` is understood as [1] or [0] (where the zero might be whatever the default initialization of `Numeric` happens to be at the time).

As a fun example, here are some Print statements with output in comments

```python
ws.Print("H2O-161 K 2 2 J 1 1", 0)  # H2O-161 J 1 1 K 2 2
ws.Print("H2O-161 K 2 2 J 1", 0)    # H2O-161 K 2 2 J 1
ws.Print("H2O-161", 0)              # H2O-161-*-*
```

The first one is understood as a `QuantumIdentifier`, the second as a `String`, and the last as a `SpeciesTag`.  Note the difference between J-and-K in order between the first two.

simonpf:
> `arts_agenda`cannot be called anymore as a stand-alone decorator. The first positional argument of it has to be a `Workspace` instance. This breaks all current python scripts. I talked to @olemke and we concluded it is better to do it this way than to try and find a global `Workspace` somehow. I have no idea how pythonic it is to demand inputs to decorators.

I don't think this makes sense from a user perspective. One of the main advantages of the Python interface is that it allows you to setup a modular simulation framework and an important part of this is defining agendas and using them across different applications and workspaces. So what will happen is that people will just create the dummy workspace themselves and pass it in to the ``arts_agenda`` macro. 

But in principle you don't even need a workspace instance to define an agenda, do you? All the necessary information is global anyways. So maybe the best way would be to just expose this directly. I can have a look at that tomorrow.

olemke:
@simonpf You're correct that at the moment all necessary information for agendas is available globally. But this is also something that is creating confusion for users. Because WSV definitions are global, but WSV values are stored per workspace, it violates encapsulation principles which can lead to problems. If a user creates two workspaces, one would expect them to be completely distinct. But creating a custom variable creates them in all workspaces. Therefore Richard and I discussed to eventually relocate the currently global WSV definitions to be part of the workspace.
At that point, the agenda would need to know which workspace it belongs to. That being said, I see that this will create issues with predefined agendas, e.g. being imported from modules. I think this needs further consideration.

simonpf:
@olemke I disagree. Note that WSVs are global in the same sense that Python class properties (which we use to represent them in the Python interface) side are global. Every instance has the same attributes but with different values. Therefore, I think that having different WSVs across different workspaces would be the actually confusing approach.

As far as I can see, most usecases for workspace-specific variables disappear with the new Python interface because they can be passed directly from Python to WSMs. The only remaining usecase I see would be to define agendas that use special WSVs. But in that case I would find it much cleaner if you would have to declare that new WSVs in the corresponding Python module:

````Python
my_variable = WSV("name", "type")
@arts_agenda
def dummy_agenda:
    Print(my_variable)
````

In this way you could define generic agendas and definitions of new workspace variables is made explicit.

Regarding the global character of the variable definitions, I think the main issue is actually that we need anonymous workspace variables to pass function parameters in agendas. So I would suggest to change the getaway functions to take pointers instead of variable indices. Agendas could then have WSM arguments stored as pointers.

**Comment by simonpf** [comment]
I had a quick go at allowing pointers being passed to the getaway functions yesterday. I didn't quite manage to get it to work all the way through to the Python interface but I pushed a proof of concept to [https://github.com/simonpf/arts/tree/pybind11](https://github.com/simonpf/arts/tree/pybind11). The key idea is to replace the Index values in the MethodRecords by a [VaraibleReference](https://github.com/simonpf/arts/tree/pybind11) struct, which can hold either an Index or a pointer directly.

In this way Agendas don't need to create or set WSVs anymore but can pass in pointers to the objects they hold directly. So there's no need for the agendas to interact with the global workspace during execution. I think this can solve all the issues we had with the global definitions of workspace variables. It also makes agendas completely independent from workspaces, so they can be defined in a generic manner.

**Comment by riclarsson** [comment]
@simonpf , I would very much like it if pointers could be passed to the methods.  This would certainly allow local classes to be used as pure input much easier and without additional copies.  However, I am not convinced that it's a good idea to make the rest of the changes you suggest. 

I really think that having workspace local names and data should be the target.  That is, `Workspace` static data is not static.

I was thinking about the `@arts_agenda` decorator a bit more.  We could just delay its parsing if there's no workspace passed to it.  That is, it returns what is effectively a C++ object of type `std::function<py::object(Workspace&)>>`.  Now, all we have to do is to not `py::cast<Agenda>(...)` but rather `py::cast<std::variant<Agenda, std::function<py::object(Workspace&)>>>(...)`.  We then simply check if we are sending a true Agenda or a delayed parser, and if the latter simply do `py::cast<Agenda>(std::get<std::function<py::object(Workspace&)>>(...)(ws))`.  Anywhere we are accessing a property we anyways have the workspace.

**Comment by riclarsson** [comment]
@olemke I cannot figure out the two remaining failing tests.  Could you help?  Especially, how can anything get stuck in the pytest?

I can now build this with debug info and it passes the tests on GCC9.  I think there might be an issue with sizes on the debug build.  I can only reliably build the python stuff with 3 cores using 32 Gb RAM and the build directory ends up being 37 Gb large at the end....

**Comment by simonpf** [comment]
@riclarsson  I guess then we could just scrap the agenda decorator altogether and allow conversion from Python functions to agendas. The only thing that is more complicated then is appending agendas, which I found quite useful.

**Comment by riclarsson** [comment]
@simonpf Skipping the decorator entirely would be quite a lot more difficult than what I have done.

Basically the decorator call gets put on hold and is returned as an `DelayedAgenda` instance with a call operator.  The call operator expects a workspace and returns a true `Agenda` by resuming the parsing once a workspace is available.

This would solve your problem and allow appending full agendas, I think:
``` python
class DelayedAgenda:
    """ Helper class to delay the parsing of an Agenda until a workspace exist
    """
    
    def __init__(self, func, allow_callbacks, set_agenda):
        self.func = [func]
        self.allow_callbacks = [allow_callbacks]
        self.set_agenda = [set_agenda]
    def append(self, other):
        self.func.extend(other.func)
        self.allow_callbacks.extend(other.allow_callbacks)
        self.set_agenda.extend(other.set_agenda)
    def __call__(self, ws):
        a = parse_function(self.func[0], ws, self.allow_callbacks[0], self.set_agenda[0])
        for i in range(1, len(self.func)):
            a.append_agenda_methods(parse_function(self.func[i], ws, self.allow_callbacks[i], self.set_agenda[i]))
        return a
```

I'm not in a position to test this before Monday though.

riclarsson:
@simonpf I tested it.  It works as I had intended it.  There's still a one major problem.

It doesn't really lend itself to interactive use.  Defining a delayed agenda currently with `@arts_agenda` and storing it to some variable, you cannot then parse it into an agenda interactively later.  At least not in the spyder-setup I am using; I am not sure if other setups store the source still.  There will be an `OSError('source code not available')` thrown in spyder, anyways.

The function itself is still defined but the source code seems forgotten.  We can thus still achieve the same main-agenda output by wrapping this function in a `CallbackFunction`.  Though that would really mess with the stacking for deeper agenda calls, I think.  I don't see a workaround for this.  I think we can also live with this problem.  I don't really see a strong use-case for appending methods to delayed agendas interactively.

PR #401, opened 2022-03-10
Some new WSMs and fixes

erikssonpatrick:
Sorry, but this is a mix of things again. Stuff that has had to be fixed during the last month for two of my projects.

New WSMs: HydrotableCalc, sensor_losReverse and IntersectionGeometricalWithAltitude. 
Fixes in: iy_surface_agenda, surfaceMapToLinearPolarisation, particle_bulkpropRadarOnionPeeling, atmgeom_checkedCalc, cloudbox_checkedCalc and more.

HydrotableCalc not fully finished but asks for a pull request now to make the other changes available for Inderpreet and Bengt, so we can start some batch simulations.

Sorry for the bad spelling in the commit messages!

### DIFF
--- a/controlfiles/artscomponents/surface/TestSurfaceSinglePol.arts
+++ b/controlfiles/artscomponents/surface/TestSurfaceSinglePol.arts
@@ -99,7 +99,8 @@ VectorSet( trv, [1] )
 AgendaSet( surface_rtprop_agenda ){
   specular_losCalc
   Copy( surface_skin_t, sst )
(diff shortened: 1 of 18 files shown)


erikssonpatrick:
@olemke Yes, various stuff in this commit as well. I had hoped to finish HydrotableCalc before the pull request, but we need to get some simulations started that depend on some of my extensions here. So would be good if you could look at this relatively soon.

PR #423, opened 2022-05-11
Replace " with ” when printing and reading from Arts XML-tag attributes

riclarsson:


riclarsson:
This fixes a problem with reading new quantum number values.  Note that if ” is used anywhere in current attribute tags, or if it is part of some data, then this will create problems.

PR #427, opened 2022-05-16
Auto agenda

riclarsson:
This gives automatic agenda settings for propmat clear sky.  It requires lines and species to be set.  There are a few weirdities in here but I think they are ok.

I'm not using the cpp API. Should this be removed? I don't think that anyone is using it and the pybind stuff is probably easier to link to anyways


riclarsson:
For the record.  Some check-all tests will fail with this adaptation.  @olemke and I discussed this and the new workspace variable `select_abs_species` will be made part of more functions than `*AddLines` in an upcoming lookup-table adaptation.  These tests should by then be restored to working order so we decided to not adapt these failing tests with expectations that the problem will fix itself by then.

PR #433, opened 2022-05-23
Improve istream double parsing performance

olemke:


src/double_imanip.h
```diff
@@ -21,52 +21,59 @@
...
-   \author Oliver Lemke
-   \date 2021-06-03
+   \author Oliver Lemke, Richard Larsson
+   \date 2022-05-23
 */
 
 #ifndef double_imanip_h
 #define double_imanip_h
 
 #include <fstream>
```
riclarsson, line 34: I wonder why `<limits>` is included in clang but not GCC, nevertheless it looks good bar GCC's lack of inclusions

riclarsson (changes requested):
I wonder why `<limits>` is included in clang but not GCC, nevertheless it looks good bar GCC's lack of inclusions

olemke:
My impression was that clang is the one with cleaner header structures requiring more explicit system includes. 🤷‍♂️

Issue #37, opened 2019-09-04
Add description for workspace groups


olemke:
Add the possibility to add description to workspace groups for the online documentation. Should follow the same implementation as WSVs.

riclarsson:
Request:  Once this is done, adding the description to make_autoarts_h.cc under both the namespace Group and Var for the relevant groups (perhaps with additional information regarding the latter being a refactored-style class) would be very much appreciated.

PR #436, opened 2022-05-25
Add group descriptions to Arts infrastructure

riclarsson:
All groups have some small descriptions added to them.  These should be updated by anyone wanting to use this.

There are two macros provided for helping generate python documentation; see code for details.  All workspace types must be documented, this is checked by the `check-python` command/


riclarsson:
Closes #37 when merged

PR #441, opened 2022-06-07
Classroom example

riclarsson:
By request, this adds a startup script for rotational spectra to an examplex folder in Arts.  The example should allow people to explore spectra related to https://github.com/atmtools/arts-lectures/tree/master/exercises/01-rotational_spectra 

By default no absorption line beyond 3 THz is read.  It takes about 3 seconds to update a layer.  ~90% of this is for O3, ~10% is for ClO, and the rest of the molecules are fast enough to not notice.

@m-brath You said I should limit this to only rotational spectra so I have deleted the other code.

@stefanbuehler and @m-brath It would be good to have your review here.  It should just work from a non-mamba environment that has the right packages available, though as we tend to work from mamba/conda environments and it does not work there, it is a bit of a problem.

@olemke How would you suggest we test this so that it continues to work?  I need a way to not start the GUI *and* to manipulate the `ctrl` struct of `propmat_clearsky_agendaGUI` (and to update it when `res` is changed by the `async`).


olemke:
> @olemke How would you suggest we test this so that it continues to work? I need a way to not start the GUI _and_ to manipulate the `ctrl` struct of `propmat_clearsky_agendaGUI` (and to update it when `res` is changed by the `async`).

How about using an environment variable to toggle a headless mode and putting something like this just before the `ARTS_GUI::propmat` call:

```cpp
  if (std::getenv("ARTS_HEADLESS")) {
    CREATE_OUT1;
    out1 << "Omitting GUI because ARTS_HEADLESS is set.\n";
    return;
  }
```

riclarsson:
@m-brath To avoid the error you found, the latest version now simply puts all ARTS errors into the GUI instead.  You are allowed to exit or continue at that time.  Still, we need a positive sorted `f_grid` in a lot of our calculations so we are not getting around that issue.

@stefanbuehler I've added longwave calculations.  It takes about 15 seconds on my machine to run one update.  If we add a cutoff of 750 GHz, then this is down to about 0.3 seconds.  There is no good reason here to use the sparse grid to speed things up, sadly, as we only have about 1000 frequency grid points.  It is the large number of bands that do most of the work here.  If we limit ourselves to some isotopologue or to just stronger absorption features, then this changes the story, but I am not sure I want to do that.

PR #448, opened 2022-06-15
Remove old xsec

riclarsson:
This removes most of the old XSEC code.  It might be too wide.  It also removes two test functions that were not run


riclarsson:
So I cannot fix the remaining issues of these tests:

	107 - arts.ctlfile.fast.artscomponents.montecarlo.TestRteCalcMC (Failed)
	108 - python.arts.ctlfile.fast.artscomponents.montecarlo.TestRteCalcMC (Failed)
	185 - arts.ctlfile.xmldata.artscomponents.arts-xml-data.TestPlanetIsoRatios (Failed)
	186 - python.arts.ctlfile.xmldata.artscomponents.arts-xml-data.TestPlanetIsoRatios (Failed)
	201 - arts.ctlfile.fast.artscomponents.wfuns.TestSpectroscopy (Failed)
	202 - python.arts.ctlfile.fast.artscomponents.wfuns.TestSpectroscopy (Failed)
	203 - arts.ctlfile.planettoolbox.planetary_toolbox.demos.passive.DemoVenus_fullRT_1D_clearsky (Failed)
	204 - python.arts.ctlfile.planettoolbox.planetary_toolbox.demos.passive.DemoVenus_fullRT_1D_clearsky (Failed)
	305 - arts.pyarts.slow.artscomponents.surfaceblackbody.TestSurfaceBlackbody (Failed)
	306 - arts.pyarts.fast.artscomponents.linemixing.TestMolOxyAdaptation (Failed)

Also, I really thing we should not be running all translated tests (or perhaps not translate all tests?).  It is very annoying to have to run some of the slow tests taking almost 10 minutes TWICE.

For tests 107 and 108,  the error is:
`The MC path sampling has failed five times. A few failures should be OK, but this number is suspiciously high and the reason to these failures should be tracked down.`

The errors seem to be related to non-positive temperatures.  This is super weird as I changed nothing about how temperature should be treated; it has to be above 0 K.  Perhaps previous test required 0 K???

For tests 185 and 186, the error is:
```
yREFERENCE-y FAILED!
Max allowed deviation set to: 2e-09
but the vectors deviate with: -0.00255508
```

This might be a compute package error or whatever.  I have no idea what should change here...

The spectroscopy failure is known for some BLASes.

Tests 203 and 204, the failure is that the stokes_dim must be above 2 as it uses Faraday.  Somehow, this was not tested before even though it is in the species list.  It is not something I can fix.

The last two also fails because of moved files.  They are OK to leave failing for this PR.

PR #455, opened 2022-07-04
Def groups

riclarsson:


riclarsson:
@olemke Is this clean enough as an implementation?  It works on the initial Workspace in both arts and python files.

@stefanbuehler  The question about default agendas is still open.  I strongly think we should have no default agendas.

My reasoning is two-fold that agendas deal with the physics applied by Arts.  You don't want to hide that.  Our target audience is scientifically literate.  They will want to know who to cite and what assumptions go into their calculations.  At least that's what I would like to imagine they would want.

The only reasonable way to default an agenda therefore seems to be if all its input is ignored and all its output simply touched.  We don't have any such cases in Arts, though perhaps for a few cases this would make sense.

Leaving the philosophy of what agendas are, the agendas that have defaults today are two: the geo positioning agenda (1) and the water equilibrium agenda (2).

1) The positioning agenda has seemingly four ways to be set.  It therefore fits as an agenda.  However, not a single method cares about the position that is computed.  It seems to me therefore that this should not even exist as an agenda, and instead the user can call one of the three methods that exist to compute it themselves if they want this as output.  And since no method cares about the output of this agenda, it seems to me that it can at most be used in internal logic, so the functions that do make use of this should perhaps rather just use another mechanism to compute the position.  However, regardless of all that logic, if we somehow care about geo positioning, they it is better for the user to know that.
2) The water equilibrium sounds like it is being set to a reasonable value.  However, again there is no method that requires the output.  Again it therefore seems like it should just be used as an internal function.  Since there exist exactly and only one way to set this agenda, it seems even more to me that this should not ever have been an agenda to begin with.

riclarsson:
Stefan, this comment is just left as a log.  So no need to respond.

We talked about some of the less used variables in Arts having defaults rather than being GIN.  You were worried that things like [abs_p_interp_order](https://atmtools.github.io/arts-docs-master/docserver/variables/abs_p_interp_order.html) might not be used enough and that it would be better if it was just a GIN.

It is not, as the link [2022-07-12] shows.  It is used also for the ease of setting up the lookup table.  I think it is definitely possible to make this variable and the other interpolation variables GIN, but it is not for this PR.  A general clean-up of `workspace.cc`, removing several of the variables that are used very seldom or in very particular ways (e.g., the one before, stuff like [nelem](https://atmtools.github.io/arts-docs-master/docserver/variables/nelem.html), and several others) would perhaps be a good idea.

I think that it is not for this PR, and as you were otherwise OK with this PR, I will merge it right after leaving this comment.

PR #474, opened 2022-07-21
Try to be consistent in using double_imanip for Numeric extraction from string/stream

riclarsson:
@olemke There were quite a few places where the locale mattered for double manipulation apparently.  I hope I caught most of them here.  Notably, `mystring.h` has an extract template that was called quite a lot so it could give some speedup this way


olemke:
Nice cleanup, thanks. No noticeable change in `make check` times on Linux with GCC though. 😞 Slowdown when running multiple test in parallel is probably caused by something else then.

riclarsson:
Would be nice to know what.  AFAIK we should be doing local calculations all the time so single core should just work OK.

PR #484, opened 2022-08-09
arts user error for aa_grid len > 1 when star is off

j-petersen:
When there is no star there is no azimuth dependency and therefore the `aa_grid` should have length 1. 
The Integration in `spectral_irradiance_fieldFromSpectralRadianceField` should run over the pen of the array itself and not other variables.

### DIFF
--- a/src/m_disort.cc
+++ b/src/m_disort.cc
@@ -362,12 +362,9 @@ void DisortCalcWithARTSSurface(Workspace& ws,
 
 
   } else {
-    
-    ARTS_USER_ERROR_IF(aa_grid.nelem() != 1,
-                       "There is no star and therefore no azimuth dependency.\n",
-                       "The length of th aa_grid should be 1, but is",
-                       aa_grid.nelem(),
-                       "instead.")
+    CREATE_OUT3;
+    out3 << "There is no star and therefore no azimuth dependency.\n";
+    out3 << "The length of th aa_grid should be 1, but is" << aa_grid.nelem() << "instead.\n";
 
     init_ifield(cloudbox_field,
                 f_grid,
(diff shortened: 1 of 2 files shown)


src/m_disort.cc
```diff
@@ -362,6 +362,13 @@ void DisortCalcWithARTSSurface(Workspace& ws,
 
 
   } else {
+    
+    ARTS_USER_ERROR_IF(aa_grid.nelem() != 1,
+                       "There is no star and therefore no azimuth dependency.\n",
+                       "The length of th aa_grid should be 1, but is",
+                       aa_grid.nelem(),
+                       "instead.")
+
```
olemke, line 371: This should not be an error, but an informational output to out3.

src/m_disort.cc
```diff
@@ -362,12 +362,9 @@ void DisortCalcWithARTSSurface(Workspace& ws,
...
   } else {
-    
-    ARTS_USER_ERROR_IF(aa_grid.nelem() != 1,
-                       "There is no star and therefore no azimuth dependency.\n",
-                       "The length of th aa_grid should be 1, but is",
-                       aa_grid.nelem(),
-                       "instead.")
+    CREATE_OUT3;
+    out3 << "There is no star and therefore no azimuth dependency.\n";
+    out3 << "The length of th aa_grid should be 1, but is" << aa_grid.nelem() << "instead.\n";
```
riclarsson, line 367: I really think this is formulated poorly.  You might want aa_grid to have a size elsewhere.  Simply state in `out3` that 

```cpp
"Disort calculation encountered aa_grid size larger than 1 in a case when it does not use aa_grid.  Calculations are performed as if there is no aa_grid."
```

j-petersen: I agree this is better.

PR #491, opened 2022-08-15
Enable C++20

olemke:
Build requires GCC 11 now.

The manylinux image comes only with GCC 10 as the system compiler. GCC 11 is installed inside the Python environment. There might be an issue with the Python 3.8 if it only provides GCC 10. Deploy logs will tell.

TODO: Enforce minimum compiler version in CMakeLists.txt.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -12,9 +12,6 @@ jobs:
       matrix:
         name: [
           ubuntu-doc,
-          ubuntu-gcc-9,
-          ubuntu-gcc-10,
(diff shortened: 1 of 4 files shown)


riclarsson:
I suggest we merge this and put a pin in the CI problem until such a time that the conda-builds can replace the pypi builds.  This would mean manually uploading versions to rttools if and when we need a new static version of pyarts (i.e., when we publish a paper).  You would have to be responsible for such static versions being uploaded, Oliver, but I suspect you are very close to get the automatic conda things working (and that those fit better in a separate branch anyways)

PR #492, opened 2022-08-15
Add no negative flag to propmat_clearskyAddFromLookup

m-brath:
Due to the high order interpolation within *propmat_clearskyAddFromLookup* negative values can occur from the cross sections from the XFIT-method though the cross sections of the lookup table are >= 0.  Therefore, I add an check and replace negative values with zero in *propmat_clearskyAddFromLookup*. With the "no_negative_from_lut" flag the check can be switched off. The default is set to True (1).

### DIFF
--- a/src/methods.cc
+++ b/src/methods.cc
@@ -13961,10 +13961,12 @@ Possible models:
          "jacobian_quantities",
          "abs_species",
          "select_abs_species"),
-      GIN("extpolfac"),
-      GIN_TYPE("Numeric"),
-      GIN_DEFAULT("0.5"),
-      GIN_DESC("Extrapolation factor (for temperature and VMR grid edges).")));
+      GIN("extpolfac","no_negative_from_lut"),
+      GIN_TYPE("Numeric","Index"),
+      GIN_DEFAULT("0.5","1"),
+      GIN_DESC("Extrapolation factor (for temperature and VMR grid edges).",
+               "Boolean. If it is true negative values due to interpolation\n"
+               "are set to zero.")));
 
   md_data_raw.push_back(create_mdrecord(
       NAME("propmat_clearskyAddHitranLineMixingLines"),
--- a/src/methods.cc
+++ b/src/methods.cc
@@ -13961,7 +13961,7 @@ Possible models:
          "jacobian_quantities",
          "abs_species",
          "select_abs_species"),
-      GIN("extpolfac","no_negative_from_lut"),
+      GIN("extpolfac","no_negatives"),
       GIN_TYPE("Numeric","Index"),
       GIN_DEFAULT("0.5","1"),
       GIN_DESC("Extrapolation factor (for temperature and VMR grid edges).",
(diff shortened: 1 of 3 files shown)


src/methods.cc
```diff
@@ -13961,10 +13961,12 @@ Possible models:
          "jacobian_quantities",
          "abs_species",
          "select_abs_species"),
-      GIN("extpolfac"),
-      GIN_TYPE("Numeric"),
-      GIN_DEFAULT("0.5"),
-      GIN_DESC("Extrapolation factor (for temperature and VMR grid edges).")));
+      GIN("extpolfac","no_negative_from_lut"),
```
riclarsson, line 13964: I think we need to have a rethink of the GIN that goes into these `propmat_clearsky*` methods.  Two things are problematic with this new name

1. We call the other variable `extpolfac` despite having similarly named GIN for other methods.
2. We already have a simpler named `no_negative` as part of the LBL GIN.

Clearly there is an overlap of the functionality we want for controlling the output, input, and/or general functionality of these methods.  We should not keep adding similarly named parameters but solve this in a better way.

riclarsson (changes requested):
We should not have so many similar names for controlling the exact same feature.

src/methods.cc
olemke: Me and Freddy discussed the parameter yesterday and came to the conclusion that it might be desirable to control the negative values generated by the lookup table separately from the negative values from LBL. So we have two ways to go forward:
1. Rename `no_negative_from_lut` to `no_negatives` to keep things simple and compromise on the separate control.
2. Keep both but name them better to make them easier to distinguish. E.g.,  rename the `no_negatives` parameter to `no_negative_from_lbl`. (Final names can be discussed of course)
Since in practice either `AddLines` or `AddFromLookup` are in the `propmat_clearsky_agenda`, but not both at the same time (a point I missed in our discussion yesterday), option 1 is maybe the better one here?

riclarsson: The longer names make very little sense in my opinion.  We have different ways of using these methods.  The long names make sense only when using the automatic setup (ignoring that the suffix `lut` should read `abs_lookup` for consistency with the `use_abs_lookup` name).

When running these methods standalone, the long name just reads like insanity `propmat_clearskyAddFromLookup(no_negative_from_lut=0)` vs `propmat_clearskyAddFromLookup(no_negative=0)`.  There is no additional information in the long name, the length just adds noise.  (A third way is when you manually setup an `Agenda`, in which case the individual methods would need to set this no-negatives manually all by themselves, so you have fine-grained control.)

Of your two options, the only one I find acceptable is to have short names.  If you need fine-grained control, you don't call the automatic function.  It's only there to cover the most common use-cases.

A third option is to have a single control struct that goes into each of these methods and contains all the GIN control we need for all of the methods.  Of course, some methods would get some controls they don't need but that doesn't really matter as they have to document what they do anyways.  The disadvantage is mainly that it would be difficult to use in `.arts` files.  The main advantage is that we won't have to introduce new GIN in the interface, we just expand on the control.  A second advantage is name consistency.  A third advantage is perhaps subtler, but it would guard us from logic inconsistency in the future as well; we won't have a `iy_unit` that changes meaning depending on function.

riclarsson (approved):

PR #493, opened 2022-08-19
Rename .val to .value for Index+Numeric

olemke:
Makes property for Index and Numeric types consistent with the other ARTS types.

### DIFF
--- a/src/python_interface/py_basic.cpp
+++ b/src/python_interface/py_basic.cpp
@@ -102,6 +102,8 @@ be accessed without copy using element-wise access operators)--");
       .PythonInterfaceWorkspaceVariableConversion(ArrayOfIndex)
       .PythonInterfaceBasicRepresentation(ArrayOfIndex)
       .PythonInterfaceArrayDefault(Index)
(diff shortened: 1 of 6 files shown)


src/python_interface/py_basic.cpp
```diff
@@ -147,7 +147,7 @@ object's instances (i.e., no element-wise comparisions))--");
       .PythonInterfaceComparisonOperators(Numeric_, Numeric_)
       .PythonInterfaceMathOperators(Numeric_, Numeric_)
       .def_property(
-          "val",
+          "value",
```
riclarsson, line 150: For both this and for the `Index`, I think you should consider also integrating the `PythonInterfaceValueOperators` instead of all the manual work that we are doing to add operators here.  Or at least a subset thereof (separating the `numpy` stuff out for `matpack`).

We are needlessly declaring a lot of variables that we should be able to rely on python for just because we want to allow `Index` and `Numeric` references (the only reason these two classes exist).

riclarsson (changes requested):
If we do this rename, it should probably also support more values operators

PR #502, opened 2022-08-25
Remove eigen from common matpack*.h;  Allow init vector and matrix ty…

riclarsson:
…pes via conceptual templates


riclarsson:
@olemke I don't know if this speeds up any compile times, but at least it separates concerns

olemke:
Doesn't seem to effect compile times, but yes, separating concerns is good.

On a side node, while testing the compile times, I noticed that it seems to be worth to enable PCH also with GCC 12. It improved compile times from 7:50 to 6:30 on my linux machine.

PR #520, opened 2022-09-09
Beginning to deepcopy workspace

riclarsson:


riclarsson:
@erikssonpatrick This is what you asked for: deep copies of the workspace.  I only tested it on some very simple stuff though, as I as of now have no use for this so no test case set up.

olemke (changes requested):
At least a minimal test cases should be included. Nothing fancy, but at least something that verifies the intended behaviour. Something like:
- Create a workspace
- Set one variable to a known value
- Make a deep-copy
- Change the variable value in the original workspace
- Verify that the variable in the copied workspace still has the correct, original value

erikssonpatrick:
Just a clarification. I asked about this, but that was under the assumption that the feature existed. I got that impression from some discussion.

riclarsson:
@olemke This now has the tests you requested.

olemke (approved):

PR #534, opened 2022-09-28
Port mpm89

riclarsson:
@stefanbuehler Here is a port of MPM89 to the modern format so we can keep Liebe's work around even when we do terminate the legacy support.


stefanbuehler:
Perfect, thanks! Stefan

PR #536, opened 2022-09-30
Rem legacy cont

riclarsson:
This removes all legacy continua models.  I ported the latest MT CKD models, the liquid water droplet ELL07, and the "standard" Rosenkranz models.  Also ported PWR98 for this PR.

I updated all controlfiles to use one of the newer models in cases where the model they were using no longer exist.

At the time of opening the PR, four tests are failing with a numerical error.  These and their errors are:

```bash
16/87 Test  #25: python.arts.ctlfile.fast.artscomponents.absorption.TestAbsDoppler .......................***Failed    0.58 sec
Traceback (most recent call last):
  File "controlfiles/artscomponents/absorption/TestAbsDoppler.py", line 51, in <module>
    ws.CompareRelative(ws.propmat_clearsky_field, ws.test_of_propmat_clearsky_field, 1e-05, "propmat_clearsky_fieldCalc output changed significantly without Doppler!")
RuntimeError: propmat_clearsky_field-test_of_propmat_clearsky_field FAILED!
propmat_clearsky_fieldCalc output changed significantly without Doppler!
Max allowed deviation set to: 0.001%
but the input deviate with: 0.00294364%
If you compare non-scalar variables, the reported deviation is
the first one found violating the criterion. The maximum
difference can be higher.
```

```bash
30/87 Test  #61: python.arts.ctlfile.fast.artscomponents.transmission.TestTransmissionWithScat ...........***Failed    0.67 sec
Traceback (most recent call last):
  File "controlfiles/artscomponents/transmission/TestTransmissionWithScat.py", line 111, in <module>
    ws.Compare(ws.iy, ws.iyREFERENCE, 1e-06)
RuntimeError: iy-iyREFERENCE FAILED!
Max allowed deviation set to : 1e-06
but the matrices deviate with: -4.81353e-05
```

```bash
60/87 Test #139: python.arts.ctlfile.fast.instruments.odinsmr.TestOdinSMR ................................***Failed    1.15 sec
Traceback (most recent call last):
  File "controlfiles/instruments/odinsmr/TestOdinSMR.py", line 91, in <module>
    ws.Compare(ws.y, ws.yREFERENCE, 0.01)
RuntimeError: y-yREFERENCE FAILED!
Max allowed deviation set to: 0.01
but the vectors deviate with: 3.7305
```

```bash
70/87 Test #191: arts.pyarts.fast.artscomponents.disort.TestDisortCalcIrradiance .........................***Failed   61.87 sec
Traceback (most recent call last):
  File "controlfiles-python/artscomponents/disort/TestDisortCalcIrradiance.py", line 172, in <module>
    ws.Compare(ws.irradiance_field, ws.irradiance_field0, 1e-6)
RuntimeError: irradiance_field-irradiance_field0 FAILED!
Max allowed deviation set to : 1e-06
but the tensors deviate with: -0.00209112
```

Clearly, I could just update three of these and use the new model values.  Some differences are not that important to care about.  However, the Odin test is failing spectacularly.  The main reason for this is that it actually wasn't using a standard continua model at all but set its own parameters.

@erikssonpatrick , is the Odin continua necessary?  Some of the continua says they are rescaled to account for completely unrelated species.  If it is necessary I will add a way to set the data (so that there is no standard way of setting it and you will have to load data from arts-cat-data or via a setup mechanic to use it)


erikssonpatrick:
@riclarsson The Odin cases are all very old. No proper testing is done and too complex to serve as demos. And I have in fact considered to remove them. So my suggestion is that you just remove all in the odinsmr folder.

riclarsson:
@erikssonpatrick Ok, I have removed the Odin tests from the latest version.  It would be nice to know if I should make the "standard" case take input parameters to be able to deal with the setup Odin had, but I guess we can leave that to a future update.

Now I just need to try and debug mac os GCC12.  (My mac cannot build with GCC12 at all because of a weird bug, so it is a bit difficult to know what's going on... anyways it will be solved before this is merged)

olemke (approved):

PR #538, opened 2022-10-07
Telsem atlas reading and agenda bug fixes

olemke:


riclarsson (changes requested):
I recognise the code code change in src/m_agenda.cc :)

This means I should probably do a DRY statement about this code.  I request you refactor this so that there exist a standalone function `std::pair<ArrayOfIndex, ArrayOfIndex> arts_agenda_inout(std::string_view)` that can be called when this is necessary.  It has the added advantage that you just write `const auto [ain, aout] = arts_agenda_inout(this_agenda.name());` in the current code (and in the part we changed yesterday).  Constness when possible is good.

olemke:
@riclarsson Good point. Refactored accordingly.

olemke:
> edit: the name of the function is perhaps bad. It should reflect that it is the global agenda's inout that are set*

I'll adjust it.

riclarsson (approved):

PR #548, opened 2022-10-21
Make empty propmat option

riclarsson:


riclarsson:
This adds an empty option to `propmat_clearsky_agendaSet` so that debugging of e.g. Rayleigh scattering can happen

PR #549, opened 2022-10-21
Fix some clangd; also make QuantumIdentifier mappable

riclarsson:


riclarsson:
I need to be able to `std::map <QuantumStuff, X>` in the near future.  This makes the PR separately so that it does not break current functionality.

PR #552, opened 2022-10-24
Altitude as main height

riclarsson:
Introduces two types `AtmPoint` and `AtmField` that encapsulate all the clearsky atmospheric data.

The `AtmField` contains  fields of data that allows `AtmPoint` data to be extracted/interpolated/computed.  Its basetype of data is any type you want that allows a clean extraction.  I only allow `GriddedField4`, `Tensor4`, `Numeric`, and functional data so far, the latter is how you get to the any type.  The `GriddedField4` should have dimensions Time, Pressure, Latitude, and Longitude.  To fix for old data of `GriddedField3`, a `Atm::fix` function is provided that does the conversion to a correct `GriddedField4` even if you only have 1D data for these grids.  The `Tensor4` data type is only allowed if the field has been regularized to some fixed grid with the `regularize(...)` function.  After this regularization, any new data to the field has to have the right sizes and be a `Tensor4`.  The `Numeric` data is self-explanatory.  The functional data is just a function that takes the 4D position of a single atmospheric point and computes the point value (e.g., for Legendre calculations).  If more gridded extraction is wanted for speed, this can be added when needs be; the data types can be expanded but I do want to keep it relatively small, e.g., no `GriddedField1` and other types that duplicate intent.

`AtmPoint` only contains `Numeric` data but wind and magnetic fields can be returned as 3-sized arrays of `Numeric`.  The rest of the data is accessed through keys such as `abs_species[0]` or `abs_lines_per_species[0][0].quantumidentity`.  Each key-type has it's own underlying data map so no copies are made.

I don't think `AtmField` needs a public way to access its data.  The same is true for `AtmPoint`, but it has public interfaces to copy its data (it's `Numeric`, there's no need to not copy).  The question of how to manipulate `AtmField` data for OEM is therefore still open.

The interface these types allows is mostly:

```cpp
// Construct
AtmField atmf{...};

// Generate a AtmPoint at a Time, Pressure, Latitutde, Longitude
AtmPoint atmp = atmf.at(Time{}, 1, 0 ,0);

// Regularize to a grid of Times, Pressures, Latitudes, Longitudes
atmf.regularize(ArrayOfTime(1, Time{}), Vector(1, 0), Vector(1, 0), Vector(1, 0));

// Access a value
Numeric o2_vmr = atmp[ArrayOfSpeciesTag("O2-66")];
Numeric nlte = atmp[QuantumIdentifier{"O2-66 J 1 0"}];
Numeric pressure = atmp.P();  // or atmp[Atm::Key::pressure];

// Set a value to either a Numeric, a GriddedField4, a Tensor4, or a functional interface
atmf.set(ArrayOfSpeciesTag("O2-66"), 0.21);
atmf.set(Atm::Key::temperature, Tensor4(1,1,1,1, 250));
atmf.set(QuantumIdentifier{...}, GriddedField4{...});
atmf.set(Atm::Key::mag_u, Atm::FunctionData{...});

atmf.has_data(Atm::Key::temperature, QuantumIdenfier{...}, ArrayOfSpeciesTag("O2-66"), ...)
```

Each step has a few more details to them.

I originally implemented this with altitude as the vertical coordinate.  If and when `Ppath` updates, this should be changed back.  It is fairly trivial to do so if the interface does not expand too much.

Open questions that remains before I go ahead and integrate this more:
1. Will OEM always require strictly gridded data?  Or will you sometimes want ungridded data?
2. How do we best allow vector extraction of data?  Today `GridPos` is used, we were told.
3. Are defaults wanted?  VMR, wind, magnetic fields all have 0s as a default now.  Of course, if you don't want a default you can ensure a value exist.
4. Anything else?
5. edit:  Extra question: Do we want strong types for the coordinates?  `Numeric(Time, Numeric, Numeric, Numeric)` is not as clean as `Numeric(Time, Pressure, Latitude, Longitude)` even though they can functionally do the same thing.


riclarsson:
@m-brath , @olemke , @stefanbuehler , @erikssonpatrick 

This is the interface I was talking about being under implementation for a unifying atmospheric data structure.

You can use `ArrayOfSpeciesTag` and `QuantumIdentifier` as keys now that a simple `<=>` operator allows for automatic comparison operations.  Thus the atmospheric fields can be accessed so that `atmosphere[abs_species[ispecies]]` means give me the data in some `std::map<ArrayOfSpeciesTag, FieldData>`.  Thus we can access values using access operators.  I think this is perhaps the easiest way to deal with growing data.  The concept of "VMR" is thus just understood; if you access this with `ArrayOfSpeciesTag{"O2-66"}` you get VMR but if you access this with `ArrayOfSpeciesTag{"liquidcloud-ELL07"}` you don't have to get VMR.

Anyways the rest is described in the original message.  I want to know if this is an acceptable type to go forward with integrating for the rest of ARTS.

stefanbuehler:
I'm confused by the time dimension. We do not have that today, do we?

riclarsson:
> I'm confused by the time dimension. We do not have that today, do we?

We do not have that today.  If we want it, this is when we should add it.  It will be difficult to add back.  It was discussed during the ARTS meeting that we might want time as a dimension.

src/variousZZZ.cc
```diff
@@ -0,0 +1,397 @@
...
+  
+  // Get grid positions
+  ArrayOfGridPos gp_row(1), gp_col(1);
+  gridpos_local(gp_row, G.get_numeric_grid(0), Vector(1, pos2D[0]));
+  gridpos_local(gp_col, G.get_numeric_grid(1), Vector(1, pos2D[1]));
+
+  // Interpolate
+  Vector itw(4);
+  interpweights(itw, gp_row[0], gp_col[0]);
+  return interp(itw, G.data, gp_row[0], gp_col[0]);
```
olemke, line 261: @erikssonpatrick There is bug here that is uncovered in debug mode by new asserts in the latest matpack. G.data must have at least 2 elements in each dimension for the red interpolation to work. Same applies to `interp_gfield3`.

erikssonpatrick: Great that new matpack is better on finding bugs! This could be avoided with some local if-statements. But that would be an ugly fix. The real solution would be to update the interpolation code to handle this. And other types of possible extrapolations, we will introduce.
It is said that we will be able to set O2 VMR by a scalar value. I see this as a nearest extrapolation for length-1 grid.  And my temporary solution here tried to allow the same for GF2. As I have said, we should be as consistent with inter- and extra-polation as possible. 
Another issue here is that the second dimension often will be longitude, and then the circular interpolation started by Richard should be applied. How to flag when that interpolation should be done?
@riclarsson @olemke A meeting to discuss how to handle this? To handle atmospheric and surface data consistently.

riclarsson: @erikssonpatrick Sure, we can have quick meeting.  I agree we should be consistent here.

The atmospheric fields keeps values like O2-VMR being constant as a simple `Numeric` so it is not applicable here, where you have fields to extrapolate from.  Perhaps you want a similar solution for the surface?

olemke: @erikssonpatrick When does it fit for you Patrick? How about today at 14:00? We're also both available tomorrow if you prefer that.

erikssonpatrick: This is not very urgent from my side. But no free time Tue or Wed. Today 14.00 works, so maybe best to take it despite I will have no time to prepare. Otherwise any time Thu is fine.

olemke:
All further development of the features started in this PR should be done by opening PRs against the [arts3-dev branch](https://github.com/atmtools/arts/tree/arts3-dev).

PR #554, opened 2022-10-26
solar ref spectra in irradiance

j-petersen:
The solar reference spectra are now given in irradiance, and therefore the conversion is removed. 
The method docu got corrected and improved.


riclarsson:
I am not sure what this actually changes and that makes me a bit queasy about merging.

I mean, it removes a scaling by pi, I see that.

I think that if you had an `examples/stars/get_toa.py` that reads the spectra you have in `arts-xml-data` to compute the TOA back again, would that still work after this or would you also need to also change the `arts-xml-data`?  Because it feels a bit arbitrary to just remove a pi without a test that you can reproduce the original data correctly.

PR #573, opened 2022-11-18
Define star spectrum at given location

j-petersen:


j-petersen:
This idea should be further followed so that the star's position is also given in local coordinates (zenith, azimuth) and not in global coordinates (lat, lon).

src/m_star.cc
```diff
@@ -123,6 +129,70 @@ void starsAddSingleFromGrid(ArrayOfStar &stars,
...
+  Numeric scale_factor = Math::pow2(radius)/
+                        (Math::pow2(radius) + Math::pow2(distance_star_loc));
+
+  // interpolate field
+  Matrix int_data = regrid_star_spectrum(star_spectrum_raw, f_grid, stokes_dim, temperature, verbosity);
+
+  // create star
+  Star& new_star = stars.emplace_back();
+
+  new_star.spectrum = int_data; // set spectrum
```
riclarsson, line 182: You really should write this directly as

```cpp
new_star.spectrum = regrid_star_spectrum(star_spectrum_raw, f_grid, stokes_dim, temperature, verbosity);
```

What you have done here makes 2 copies of the `int_data` as you copy it into your `new_star` rather than move it into you `new_star`.

riclarsson (changes requested):

src/m_sun.cc
```diff
@@ -104,26 +107,110 @@ void sunsAddSingleFromGrid(ArrayOfSun &suns,
...
+  if (zenith < ANGTOL){
+    sun_altitude = distance + toa_altitude;
+    sun_latitude = location_latitude;
+    sun_longitude = location_longitude;
+  } else if (zenith > 180 - ANGTOL) {
+    sun_altitude = distance - toa_altitude;
+    sun_latitude = -location_latitude;
+    sun_longitude = location_longitude + 180 - 360.0 * Numeric(round((location_longitude - 0.0) / 360.0));
+  } else {
+  Numeric x, y, z, dx, dy, dz;
```
riclarsson, line 165: The tab is wrong, move it in.

src/m_sun.cc
```diff
@@ -104,26 +107,110 @@ void sunsAddSingleFromGrid(ArrayOfSun &suns,
...
+                dx,
+                dy,
+                dz,
+                toa_altitude,
+                location_latitude,
+                location_longitude,
+                zenith,
+                azimuth);
+
+    Vector sun_pos_cart = {x+distance*dx, y+distance*dy, z+distance*dz};
```
riclarsson, line 178: It is not necessary to create this Vector

riclarsson (commented):
I want some small changes but once fixed I think you are good.  And the changes are very small.  Good job!

src/m_sun.cc
j-petersen: I think it is not the best for readability to remove the Vector, however, you are right, performance wise it is unnecessary.

riclarsson: Then call it `std::array<Numeric, 3>` instead.  That is about as good as it gets.

j-petersen: I'm fine with it, because it can still be figured out quickly. Anyhow it is good to know how it would be done the best way.

riclarsson (approved):

PR #578, opened 2022-12-05
Add description for adding groups; you must also do python

riclarsson:


doc/uguide/development.tex
```diff
@@ -498,6 +498,12 @@ \subsection{How to add a workspace variable group}
   (\shortcode{operator<<}), you have to add an explicit implementation
   of the \builtindoc{Print} function in \fileindex{m\_general.h} and
   \fileindex{m\_general.cc}.
+\item Add the group to the python interface's main module.  See the python
+interface description for adding new groups.  Failing to add the group here
+will likely lead to the inability to import the pyarts target to python.  Note
```
simonpf, line 503: I think 'pyarts module' is a better formuation than 'pyarts target'.

riclarsson: Thanks for the catch!

PR #588, opened 2023-01-17
Use source_location when available for better error msg

riclarsson:


riclarsson:
Please merge if you find it OK, @olemke 

Moving the implementation to some debug.cc file is also possible, and might be better, but I am not what package that should go into, and debug.h is described as macro land, so I think it works OK in there for now

PR #593, opened 2023-01-25
Add LBL example routine

riclarsson:
This adds an example for how to use the arts-cat-data line data.

It is focused only on describing how you set up you LBL species tags and how these then influence how the data is read from file.

As a policy, all things that are added to "examples/" should have two reviewers so I will wait for two of @olemke , @stefanbuehler , and @erikssonpatrick have time to review this before it is merged.  Please take time to read this as we will be applying the same policy to your code that touches "examples/" in the future :)


stefanbuehler:
Great! I really like the spirit of this. :-) OK to merge from my point of view.

olemke:
Looks really good. One suggestions I have is to maybe add a simple plot as an output to give the user something to look at when running the example? Nothing fancy, just something like:

```python
import matplotlib.pyplot as plt
plt.plot(ws.f_grid.value / 1e9, ws.y.value)
plt.xlabel("Frequency / GHz")
plt.ylabel("Brightness Temperature / K")
plt.savefig("lines.pdf")
```

examples/arts-cat-data/lines/lines.py
```diff
@@ -0,0 +1,169 @@
...
+ARTS setup for you.  Note that the flag use_abs_lookup=1 can be passed to this
+method call to set up the agenda for USING the the lookup-table.  Without the
+flag, ARTS should be configured correctly to either COMPUTE the lookup-table
+or to compute the absorption on-the-fly
+
+"""
+ws.propmat_clearsky_agendaAuto()
+
+# That's it! You now have working absorption calculations for your loaded lines.
+# We still need to setup an atmosphere to perform our calculations in.
```
olemke, line 116: to setup -> to set up

examples/arts-cat-data/lines/lines.py
```diff
@@ -0,0 +1,169 @@
...
+# This part is not covered in great details in this example
+
+# setup some agendas and core settings
```
olemke, line 119: setup -> set up

olemke (changes requested):
Really like the example and how well documented it is. Just some minor style comments and typos. Good to merge after those are resolved.

erikssonpatrick:
Great to see the first example on our new example files. Yes, this is much nicer than our old "demos". So a general thumbs up. But I also agree with comments from Stefan and Oliver.
1. Should we really generate spectra in all examples? A lot of code will be repeated over and over. Now it is just three lines of PyARTS that are specific to this example. I think it is better to focus on the absorption calculation, because it is more to say there. And I don't think we shall include testing by comparing to reference values in examples. That is for tests.
2. Would be nice to end with a figure.

My suggestion is to generate a lookup table (with multiple species) and plot the cross-sections.

olemke:
> .... And I don't think we shall include testing by comparing to reference values in examples. That is for tests.

But we want to make sure that the examples work correctly and don't break unnoticed. The easiest way to do this is to run them during the check stage and verify the result.

examples/arts-cat-data/lines/lines.py
```diff
@@ -0,0 +1,169 @@
...
+ws.cloudbox_checkedCalc()
+ws.sensor_checkedCalc()
+
+# perform the forward calculations using standard methods
+ws.yCalc()
+
+# Save test results
+# ws.y.value.savexml("lines_test_result.xml", type="zascii")
+
+# test that we are still
```
olemke, line 166: end of sentence missing

erikssonpatrick:
Some more detailed comments:
* I think it is better if the practical example includes multiple species, e.g. ws.abs_speciesSet(species=["O2-66", "H2O"])
* Despite all the nice descriptions, I think there are still aspects that are not clear. For example, what will end up in second species if doing: ws.abs_speciesSet(species=["O2-66", "O2"])
* Don't you want to be clearer about on-the-fly vs lookup table in this example? That is, to exemplify this. As it is now, I don't think a newbie will get it.  A detail here, I don't follow this explanation: "Without the flag, ARTS should be configured correctly to either COMPUTE the lookup-table or to compute the absorption on-the-fly" 
* Or should it be two examples, one focusing on abs_species and a second dealing with on-the-fly/lookup-table?

An extension to my first comment. I suggest to set up a minimal example on a full calculation, pretty much as this example but with most comments removed. And adding new comments, e.g. shortly explaining that the definition of an absorption species is flexible, but referring to this example for details.

olemke (approved):

PR #594, opened 2023-01-25
Add CIA example routine

riclarsson:
Same as for LBL but using CIA routines.

I still requires two reviewers to look at this before I will merge.

This code also found and fixed an error in the reading-routine that is now fixed.  Since no one ran into this problem before I think it is not urgent so I will not split this PR.

@stefanbuehler , the example shows O2-CIA-O2-0, which is far into the THz range, so I think that it represents well the idea you mentioned before --- I did extend ARTS CIA to higher frequencies by porting newer HITRAN code.

To @erikssonpatrick , @olemke and @stefanbuehler , I think CIA is poorly implemented in ARTS.  I do not see the need for the final index, separating the internal data structure.  The example I give for how to deal with adding all O2-CIA-O2 absorption is the only way to do this today.

I also still think that the example for adding additional `abs_species` just to get at a VMR that is required for a CIA tag is a little bit ridiculous, but that will go away with the #552 branch as that is merged, since it redefines the entire atmosphere.  So I don't want to suggest any solutions for now.


stefanbuehler:
Overall: Great! This is really nice, the right level of detail. 

My comments:

Just wording: are surrounded by --> are separated by (otherwise wrong meaning)

"O2-CIA-O2-0,O2-CIA-O2-1,O2-CIA-O2-2,O2-CIA-O2-3,O2-CIA-O2-4,O2-CIA-O2-5,O2-CIA-O2-6,O2-CIA-O2-7": Isn't this rather cumbersome? Especially since most users can not know what the band indices mean exactly. How about introducing "O2-CIA-O2" to mean all bands, consistent with how the LBL species work?

Finally, quite a large part of the control file is the RT calculation. In case this is repeated in several of the examples, perhaps kick it out here? Perhaps instead just calculate absorption?

examples/arts-cat-data/cia/cia.py
```diff
@@ -0,0 +1,131 @@
+"""
+
+This file will showcase how you can load collision-induced absorption (CIA)
+data from arts-cat-data into the workspace and do the required setups to
+perform a simple forward calculations using this data
+
+Note that this example presumes that you have set ARTS_DATA_PATH to contain
```
olemke, line 7: Add "you have set _the environment variable_ ARTS_DATA_PATH" for clarity

examples/arts-cat-data/cia/cia.py
```diff
@@ -0,0 +1,131 @@
...
+
+"""
+
+Set ws.abs_species to the species tags that you wish to use.  CIA tags are
+structured so that the main and secondary species are surrounded by the word
+"CIA".  These tags are followed by an extraction index, to select which CIA
+data is used.
+
+This sets up self CIA by molecular oxygen for the first "O2-CIA-O2-0" CIA
+band.  To include all CIA band by O2, you have to write:
```
olemke, line 34: all CIA band -> all CIA bands

examples/arts-cat-data/cia/cia.py
```diff
@@ -0,0 +1,131 @@
...
+    "O2-CIA-O2-0,O2-CIA-O2-1,O2-CIA-O2-2,O2-CIA-O2-3,O2-CIA-O2-4,O2-CIA-O2-5,O2-CIA-O2-6,O2-CIA-O2-7"
+    
+CIA is not just self-induced but can occur between two different moleules.  In
```
olemke, line 37: moleules -> molecules

examples/arts-cat-data/cia/cia.py
```diff
@@ -0,0 +1,131 @@
...
+this case you can change the tag to read something like "O2-CIA-N2-0", though
+you must ensure that there also exist another N2 species in your ws.abs_species
```
olemke, line 39: exist -> exists

examples/arts-cat-data/cia/cia.py
```diff
@@ -0,0 +1,131 @@
...
+
+"""
+ws.propmat_clearsky_agendaAuto(T_extrapolfac=1)
+
+# That's it! You now have working absorption calculations for your loaded lines.
+# We still need to setup an atmosphere to perform our calculations in.
+# This part is not covered in great details in this example
+
+# setup some agendas and core settings
+ws.iy_unit="PlanckBT"
```
olemke, line 82: Minor style issue: missing spaces around the `=` sing

examples/arts-cat-data/cia/cia.py
```diff
@@ -0,0 +1,131 @@
...
+ws.cloudbox_checkedCalc()
+ws.sensor_checkedCalc()
+
+# perform the forward calculations using standard methods
+ws.yCalc()
+
+# Save test results
+# ws.y.value.savexml("cia_test_result.xml", type="zascii")
+
+# test that we are still
```
olemke, line 128: sentence cut off

olemke (changes requested):
Just some typo and style corrections.

riclarsson (commented):
Fix textual stuff from Oliver

riclarsson (commented):
Fix text comment by Stefan

riclarsson:
@stefanbuehler I will leave introducing "O2-CIA-O2" as a separate branch but also leave this PR open until that issue is resolved.  I agree, there is no point in having "O2-CIA-O2-N", as this just confuses things.  If you want less that or just a specific band, you prepare the data yourself.

Also, I think showing just the absorption is a good idea.  I will change this after checking with Patrick and Oliver at tomorrow's Zoom.

olemke (approved):

PR #597, opened 2023-02-02
CIA now does all species

riclarsson:


riclarsson:
@stefanbuehler This addresses your comment on #594 .  The only accepted behavior of the CIA tag system now is "O2-CIA-O2", that is no bands at all.  I think it is better to just do this this way because choosing a band is only important if it is difficult to modify the input data.  Python makes modifying the input trivially easy, so I think we should go for the more sensible option here.

stefanbuehler:
I agree.

riclarsson:
Excellent.  Then I merge this!

PR #598, opened 2023-02-20
Interpolation routine in a uniform matter

riclarsson:
This turns the interpolation routine I added in the Interpolation namespace generic for any matpack type.  That is the same interface as for interpweights and all that we have had before, so 1-dimensional interpolation just as ARTS does it today.

I also added a non-interpweights version of reinterp and interp because that of course works and just allows you to allocate one less structure.  This is of course a lot faster unless you want to reuse the interpolation weights, in which case it gets slower after a few interpolations.

The branch also adds constant-sized matpack data and view.  In the key parts, it behaves similar to std::array; e.g., in constructor.  I needed this instead of std::array to ensure the interface was clean for these interpolation routines.  As with the non-constant data, I think it is fine to add any type of feature that std::array has got for the 1D case (as we said we can for std::vector for Vector).  The interface between constant-matpack-data and runtime-matpack-data is just that you can copy over the constant data to the non-constant data at this time.  I think it might be nice to have a way to claim that a view is constant, but it is not there.  Also, there are no strided views allowed for constant data...  but I don't really intend for the interface to go much beyond that.  Just that they should conceptually support most similar things.

I leave this as a draft so people can have a look at naming and interfaces.  It is not entirely ready.  I need to add a few more perf-tests and figure out how exactly the CIA test fails (it really shouldn't).  I also need to compose a perf-table.

This will need some expansion later for the #552 branch.  I want this in first before we can support 2D weights because I still need the altitude for 1-dimensional interpolations and it would be a pain to delay this restructure until then.


erikssonpatrick:
Sounds great! Some quick comments. 
Documentation for developers? Any thoughts? It does not need to be very detailed, but it would be good to have a clear list of options. For example, do we have anything like spline interpolation? When to use interpweigts or not?
So, now much quicker to not involve interpweights for a single interpolation. A big difference? Do you adjust the code where interpweights cause an overhead? Or that's up to others?
With this giving us more options, we will need be clearer in the method descriptions what interpolation that is applied. Is it still the case that equidistant grids give faster interpolation? Useful information for users.

riclarsson:
> Sounds great! Some quick comments. Documentation for developers? Any thoughts? It does not need to be very detailed, but it would be good to have a clear list of options. For example, do we have anything like spline interpolation? When to use interpweigts or not? So, now much quicker to not involve interpweights for a single interpolation. A big difference? Do you adjust the code where interpweights cause an overhead? Or that's up to others? With this giving us more options, we will need be clearer in the method descriptions what interpolation that is applied. Is it still the case that equidistant grids give faster interpolation? Useful information for users.

I must take back what I said about it being a lot faster.  It is 20% faster if you have compile-time known polynomial size.  It is actually 20% slower if you don't have that, so there is something weird going on.  I don't think I want to recommend it at this point so I will leave the old documentation around for that.

Note that compile-time known, I mean that the user cannot change the polynomial order.

So I change nothing and this needs more testing.  

The entire idea of spline interpolation is polynomials on parts of the grid, no?  If so we have had that since forever and this still supports it.

Anyways, I will have alook at updating names in the documentation

riclarsson:
So I won't put any exact numbers here on changes in execution time.  I cannot even get the old gridpos to have the same execution time so there's something about this problem that really messes with my settings.  In average it seems the new code has about 70-50% the runtime of the old.  But since I cannot get gridpos to run within 20% of the runtime of old, it is weird.

But yes, it is still the case that uniform grids are faster.  I tried making that independent of grids, and it is relatively easy to do so, it's just that it is a lot slower across the board to do that.  Do you want this to be a feature?  I could add a customization point so you can send in whatever grid-position function you want to have.

erikssonpatrick:
If the question were directed to me, the answer would be no. Just asked to know how things work.

PR #599, opened 2023-02-20
A bug fix

erikssonpatrick:
The first commit is an algorithm improvement.
The second commit fixes yet another numerical problem in the geometrical calculations. That is a bug fix. This should, in principle, be patched to older versions, at least back to 2.5.6. We noticed it by using 2.5.6. For now, we can avoid the bug. I leave it to you, Oliver to decide about patching.


erikssonpatrick:
Sorry that I had missed to run make check. I had started to fix methods.cc. There were several bad sentences in the description of dlosGauss. So improved that as well.

olemke:
> This should, in principle, be patched to older versions, at least back to 2.5.6.

Since the function doesn't exist in 2.4.x, no need to port it back there. As for 2.5.x, it will naturally be included in the next pre-release (2.5.12) once we release it. Users are supposed to upgrade to the latest pre-release to get the latest fixes for the development version.

For 2.5.6, it's your call make if you think it is necessary to make a branch available in the arts repository that includes the fix. If people should be able to checkout the exact version used in the study, it would make sense to provide a dedicated branch in the main repository for that. Do you already have a branch for that study in your fork that contains other modifications deviating from the original 2.5.6 version? If so, I would use that as the basis for creating a branch in the main repository and call the branch something like `2.5.6-eumetsat` to make clear that it's a special purpose version.

erikssonpatrick:
Maybe I have misunderstood when you apply bug fixes backwards. 
Anyhow, we have no other changes in v2.5.6. And I just got a message from Bengt explaining that his Python code does not handle observations around nadir perfectly. So we need to stay away from nadir, and no need for the bug fix.
No open ends here from my side then.

PR #605, opened 2023-03-09
Adding ppath plotting

riclarsson:


olemke (changes requested):
Very nice. Two requests: Currently all functions from `ppath.py`, including all helper functions, are exposed in `pyarts.plots`. To make it easier for users to discover the relevant plot functions, I'd suggest to modify `__all__` after the import in `ppath.py` like this:
```python
__all__ = [
    'polar_ppath',
    'polar_ppath_list',
]
```

My second request is to add the user facing functions to be listed in the sphinx documentation by adding them to `python/doc/source/pyarts_plots.rst`.

riclarsson:
@olemke I was merely following the existing practice of `__all__` :)

Anyways, this should be good now.  I also added the ppath stuff to the rsts in source/ so that these functions are documented.

olemke:
@riclarsson Maybe there is a misunderstanding here. I was not asking for changing how `__all__` in `__init__.py` is handled, only for listing the public functions explicitly in `__all__` in `ppath.py`.

olemke (approved):

PR #608, opened 2023-03-24
[arts3-dev] Make propagation matrix use atm point

riclarsson:


riclarsson:
@olemke and @m-brath , will all the gas scattering stuff use atm_point?  It uses it basically if it is available and intended to be used mostly on the level of the propagation matrix.

@erikssonpatrick Same question for you about refractive indices.  Are these also used only at the level where the propagation matrices are required?  Because they seem to use the same kind of input.  There are currently only 7 methods that uses rtp_pressure, so it would be nice to remove it (funnily enough, it is the gas-scattering and refr-index stuff, which of course are doing similar things so of course they both need it).

erikssonpatrick:
I think the answer is yes. That is, refractive indexes should be extracted at one point at a time. However, I want to change it from how it works now. refr_index_air_agenda takes e.g rtp_pressure and rtp_vmr. I want it to take a position (z,lat,lon). And methods that actually need pressure, temperature and some vmrs should "ask" for these values at the point. For refractive index you normally don't need VMR of minor gases, just P, T and H2O. So by default returning VMR of all gases would be unnecessary here.

riclarsson:
@erikssonpatrick Excellent!  I think you should then design these functions to take the entire atmospheric field.  It's anyways from there that a point value must be extracted.  We can look at optimising the extraction routine to just get the numbers you need out, either as an atmospheric point struct or just pure numerics

PR #610, opened 2023-03-31
Add IGRF to atm_field possible settings

riclarsson:
This adds a function to extract mag_u, mag_v, mag_w from IGRF.

To speed up calculations, only one call to compute is made

### DIFF
--- a/src/m_atm.cc
+++ b/src/m_atm.cc
@@ -343,15 +343,23 @@ void atm_fieldAddNumericData(AtmField &atm_field, const QuantumIdentifier &key,
   atm_field[key] = data;
 }
 
-void atm_fieldIGRF(AtmField &atm_field, const Time &time, const Verbosity &) {
+void atm_fieldIGRF(AtmField &atm_field, const Time &time, const Index &parsafe,
+                   const Verbosity &) {
   using namespace IGRF;
+
+  //! We need explicit planet-size as IGRF requires the radius
+  //! This is the WGS84 version of that, with radius and flattening
   static const Vector ell{6378137., 0.081819190842621};
 
+  //! This struct deals with the computations, it's internally saving a mutable
+  //! state
   struct res {
     mutable Numeric u{}, v{}, w{};
     mutable Numeric alt{}, lat{}, lon{};
     const Time t;
+
     res(const Time &x) : t(x) {}
+
     void comp(Numeric al, Numeric la, Numeric lo) const {
       if (alt != al or lat != la or lo != lon) {
         alt = al;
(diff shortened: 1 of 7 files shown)


src/m_atm.cc
```diff
@@ -340,3 +342,57 @@ void atm_fieldAddNumericData(AtmField &atm_field, const QuantumIdentifier &key,
                      "Cannot add numeric data to regularized atmospheric field")
   atm_field[key] = data;
 }
+
+void atm_fieldIGRF(AtmField &atm_field, const Time &time, const Verbosity &) {
+  using namespace IGRF;
+  static const Vector ell{6378137., 0.081819190842621};
```
olemke, line 348: Instead of using "magic" numbers, it would be nice to use Earth radius from our constants. And what is the second number?

erikssonpatrick: @riclarsson If the second number is flattening, please note that I have changed the definition of refellipsoid to instead have polar radius as second element. The new version is so far called refellipsoidZZZ:

*-------------------------------------------------------------------*
Workspace variable = refellipsoidZZZ
---------------------------------------------------------------------

Reference ellipsoid.

This vector specifies the shape of the reference ellipsoid. The
vector must have length 2, where the two elements are:
  1: Equatorial radius.
  2: Polar radius.

Usage:  Set by the user.

Size:   [ 2 ]


---------------------------------------------------------------------
Group = Vector
*-------------------------------------------------------------------*

riclarsson: @erikssonpatrick Then I will update the code when we make the switch.  I think it is relatively easy in the underlying code, but it is used from parts that takes `refellipsoid` so I cannot do that update before  :)

@olemke It is not a magic number per se.  IGRF works in spherical coordinates.  I need the exact radius to a point.  We only allow two components in ellipsoid, so this is the closest I can get to give a good "radius" as input.  Since IGRF is a model, it works on its own idea of Earth and not on the user's idea of Earth

olemke: In that case a comment explaining what those two number are would be appropriate.

riclarsson:
@erikssonpatrick One more thing.  Since we are going towards using more explicitly names types, I think it would be very good if `refellipsoidZZZ` and other relevant types that has to have a fixed size to function actually do have a fixed size in the future.

I added a templated type to allow for that: `matpack::matpack_constant_data<T, DIM1, DIM2, ...>` which for `refellipsoidZZZ` would look like `matpack::matpack_constant_data<Numeric, 2>`.  (Of course, we will be using this in the code perhaps as `Vector2` or something like that to keep the name short and easy, but it is generic enough you can have any dimension data of any type represented by it.)

It would help a lot if we can use more explicitly named types like this when we make the new `Ppath` also.  Especially, `ArrayOfVector` is not continuous memory but just a bunch of vector-pointers laid out continuously in memory so that they then need to be de-referenced one-at-a-time before they're used.  However, `Array<Vector2>` is continuous memory layout because the data is actually in the `Vector2` and not in a pointer in the object (as for `Vector`).

Philosophically, it would also fit to make these types explicitly sized.  A `pos` is always 3D, a los is always 2D, and so on.  Because then we don't have to care about sizes and stuff.  I don't even think we have to care about copying the data either at that point, but that's another issue.

Just as a small example.  There is a problem in `Ppath` today that `los` is a `Matrix`.  This is very good for the memory layouts in `Ppath` but `propmat_clearsky_agenda` requires `rtp_los`, which then has to be a `Vector` since we use that as our interface type to `Agenda` calls.  If `Ppath` 'los' was an `Array<Vector2>`, we would need no copies for the call to `propmat_clearsky_agenda`.

erikssonpatrick:
@riclarsson I am in favor of seeing a pos and los as a Vector3 and Vector2, respectively. You discussed Ppath.pos. But notice that we also have e.g. sensor_pos.
But feels a bit odd to discuss this in a PR about IGRF. And in general, I prefer to discuss basic changes like this in a structured way. To make sure that we are fairly consistent. One question here is if there are parallel cases, where we also should use fixed size types. What about Stokes vectors? Should we make them Vector4? Should WSVs that are "flags" be booleans?

PR #611, opened 2023-04-05
License update

riclarsson:
This PR makes most of ARTS LGPL compatible

It updates the license to a dual license of choice: LGPL or GPL.  Note that the choice depends on compilation options and not just personal feelings.

There are two CMAKE flags introduces "ENABLE_ARTS_LGPL" and "ENABLE_ARTS_PURESCIENCE".  By default they are OFF.  If both are on, the compilation will fail.

If the former is true, code paths that are not LGPL compatible are disabled.  That means no cdisort and no legendre-gauss quadratures.  @olemke will contact cdisort and ask for license clarity.

When the second flag is off, this means that all AER code is disabled.  I hope to re-enable the latest version of the code when the theory guide is published on Hitran in the future so I can base my implementation on it.  It also means that RT4, TMATRIX, FASTEM, and REFICE are disabled.  FASTEM should not be in ARTS at all because we are not even allowed to distribute it as far as I can see.  REFICE and RT4 both have no license attached to them, so we cannot compile them in.  TMATRIX has the same clause as AER, that the code is only allowed for scientific purposes.

The idea here is that the LGPL flag will be on for the pyarts binary package we distribute, but of course not the science flag.  This is a very big limitation but it is one we have to make.  Otherwise we are not allowed to distribute the binary for educational purposes, for instance (since neither AER or TMATRIX allow that).

Note that it would have been a lot more parts of ARTS disabled if we did not also update the random number generation library to use standard code paths instead of GSL.  As an added bonus, we can now sample any random number distribution supported by the [C++ standard](https://en.cppreference.com/w/cpp/numeric/random) (including normal, log-normal, uniform, and even several integer distributions among others).  We basically no longer have to limit ourselves to uniform numbers between 0 and 1, though that's of course what we mostly use.

I also tried to fix the legendre-gauss quadrature code to move away from GSL, but the definitions of that code does not seem to agree with Wikipedia about what they are actually doing, so it is beyond me.  @m-brath , could you explain what that code is doing?  It is not doing Legendre Gauss quadrature, because that's defined between -1 and 1 but the code seem to go only from 0 to 1.

Lastly, I will mark this as a draft PR for now because @stefanbuehler and @erikssonpatrick , you need to review this change.  I also need your explicit (as in write it here) approval to remove all the remaining and relevant statements about GPL in our src/ directory.  Before those are gone, those files being compiled of course means that the binary is still GPL.


erikssonpatrick:
A small remark. REFICE can be removed. It's old code and I don't think anyone is using it. At least I would go for a newer model.
(It's on my ToDo list to bring some order to functions producing refractive index. And some new ones.
Going back to the discussion initiated by Vector3. Shall we have a special type for WSVs that are complex numbers?)

erikssonpatrick:
But I don't have much to say when it comes to the overall license issue. Here I am lost and will trust your judgement.
One question. Did we agree on what licence we would wish ARTS should have? The above sounds as we are just following external forces.

riclarsson:
@erikssonpatrick , I would be happy if we remove unused code.  I am not sure who added REFICE though, so I will leave it up to this person to decide if it can be removed, and they should actually do the work of removing it too.  There's probably some more parts of ARTS/src that can be simplified if it is removed...

About reacting to licenses, I am not sure what you mean?  We have to follow external software's licenses.  We haven't so far.  Is this the reaction?

We decided triple licence choice of GPL / LGPL / NoMilitary last meeting.  The latter is not really a license, but a limitation that is incompatible with GPL / LGPL.   I added the flag `ENABLE_ARTS_PURESCIENCE` for the NoMilitary purpose.  If you want to make use of it, I can give you a short guide.  I personally think it is a horrible idea to use that flag because we cannot put those things into the python package since other parts of ARTS are still GPL / LGPL by licenses.  Notably the new `mdspan` code is all only GPL3 / LGPL3 compatible.

Note, as a warning, that code that uses the `ENABLE_ARTS_PURESCIENCE` is therefore not allowed to use `matpack` at all, because this will always be GPL / LGPL.  So if you have anything `matpack` related in there (e.g., use the `Vector` / `Matrix` classes) then the `ENABLE_ARTS_PURESCIENCE` flag is void and has no meaning because you have then relicensed your code to GPL / LGPL automatically.

PR #613, opened 2023-04-11
Implement recent Rosenkranz absorption models

stuartfox:
This PR implements two recent releases (2021 and 2022) of Phil Rosenkranz's complete absorption models for H2O, O2 and N2 in ARTS. I've based the implementation on the FORTRAN codes available on http://cetemps.aquila.infn.it/mwrnet/lblmrt_ns.html

A couple of points/questions that people who know more about ARTS and/or C++ might want to comment on:
- I called these models PWR2021 and PWR2022. Would it be better to name them PWR21 and PWR22 (potentially causing confusion in the year 2098)? Alternatively they could be labelled as "MPM" models as @riclarsson did with "MPM2020" which is a partial implementation of the O2 model (although with parameters taken from the papers rather than the Fortran codes).
- I included both releases in a single PWR20xx.cc source file since it's only the line parameters that are changed between them, the calculation method remains identical. Is it intended that each "predefined" source file should only contain a single model, or is it OK to include multiple releases like this?
- I used std::valarray to store the line parameters to allow mathematical manipulation of all lines simultaneously. However, this seems to be a little-used C++ feature and caused unexpected errors in my initial implementation on the clang (LLVM) compiler when I combined it with "auto" type definitions. I've removed the "auto" types now so that it works, but I could probably re-write using std::vector and std::transform (or loops) similar to MPM2020.cc if that would be better, although I find the resulting code harder to understand.


riclarsson:
Hi Stuart, neat code!

The one thing I would like you to change in the code before merge is your use of `std::complex<Numeric>`.  We have `Complex` in ARTS for that type, and it is nicer to stick with it because we might do small changes in the future (e.g., we changed `Index` from being a `long` to a `std::int64_t` recently; and future versions of C++ will have `std::float64_t` so depending on factors beyond our control, we might want to use that type [or not], so it is nice to be safe and just use `Complex` when possible).  The rest here is just my thoughts and answers to your points above.  Another small thing is that I think that all of your `const vararray` that holds pure numeric data for the model should be marked `static` in addition to `const`.  These data are anyways `static` as they are held by the binary so it is a good idea to mark it as such.  (Note that you cannot mark the non-const data this way as you modify them by reference later on.)

I think the name of your implemented model should be PWR as this is different from the MPM code I added.  PWR is Rosenkranz version using the MPM (now v2) model.  It makes sense in my mind that new MPM code is for keeping research questions about best practices alive and PWR code is for using the best practice according to Rosenkranz.  These two things must be done in parallel.  I would keep the longer name (20xx) because it is a century spanning model :)

It is more than OK to have the code in the same file.  The main point of these files is to keep similar things together and not end up with a behemoth file like we had before.  The namespace naming is a bit unfortunate, I do not like `xx` part.  But that's fine also for now, when more models are added small things like that can be adjusted.  (Perhaps I should have put all PWR code in a PWR namespace...)

About `std::vararray`, it is fine.  It has the big problem that it is not a predictable type.  The [cppreference](https://en.cppreference.com/w/cpp/numeric/valarray) article literally writes "functions and operators that take `valarray` arguments are allowed to return proxy objects to make it possible for the compiler to optimize an expression such as v1 = a*v2 + v3;".  What this means is that you can never know what the return type of any `valarray` operation is, and that this proxy type may change with your compiler.  It does make your code neat and compact to read though, so that is very nice!

I would have used `std::array` or `matpack::matpack_constant_data` to hold the data were I writing this code myself.  They both have the advantage that they do not allocate but use the stack.  They can both therefore also be marked as `constexpr`.  The speed of the executed code would probably also be a bit faster but if you are happy with the execution speed, then that's good.  Note that especially `matpack::matpack_constant_data` is useful for this kind of code as it can put all the data in a single structure, `matpack::matpack_constant_data<Numeric, NROWS, NCOLS>`.  It then works very well with the `transform` paradigm, because a `begin()` and `end()`, or `for(auto&& vec: mat)` returns a view of all the values in the right-most column(s).  So you can more easily write the `transform` statement for things like line-by-line manipulation.  (`matpack::matpack_constant_data` is a very recent addition to ARTS I made that uses an `array` in the background with a reference implementation of the new multi-dimensional standard C++ type and my own iterator-type for it's interface functions.)

stuartfox:
Thanks @riclarsson. I've replaced the `Complex` types and added `static` as requested. I've left the other things for now but am happy to revisit in future.

PR #614, opened 2023-04-19
Arts3 Make Standardized IY

riclarsson:


riclarsson:
This is a draft for making new RTE calculations for discussion at upcoming Zoom meeting.  The idea is that the main calculations are split up.  In ARTS-2.5, we have the quality-of-life problem that each of our `iy*` methods behave mostly the same but are independent methods rather than just wrappers around concepts such as `Radiative Properties` and `Background Radiation Scheme`.  By "mostly the same", I mean that `iyEmissionHybrid` and `iyClearsky` are practically `iyEmissionStandard` but with additional propagation path effects or additional source radiation. 

This draft attempts one way to solve this, with some still rough edges, by splitting the core methods up into smaller parts.  The future idea for ARTS RTE, in my opinion and as I think I have talked to @olemke about several months ago, should be that the selection how the radiative background behaves should come from an `rte_background_agenda` and the way the radiative properties of the atmosphere behaves should come from an `ppvar_rtprop_agenda`.  The rest should just be core RTE calculations.  This is the split we already make in several functions.

This PR is a rough draft for that idea, where basically you can replace `iyEmissionStandard` in the call to `iy_main_agenda` with a set of functions, such that these two behave the same (using pyarts and `print(ws.iy_main_agenda.value)`):
```
Arts Agenda iy_main_agenda with methods:
   ppathCalc
   ppvar_atmFromPath
   ppvar_fFromPath
   RadiativePropertiesCalc
   background_transmittanceFromBack
   RadiationBackgroundCalc
   ppvar_radCalc
   iyCopyPath
   diy_dxTransform
   iyUnitConversion
   iy_auxFromVars
   VectorSet(geo_pos)
   Ignore(iy_id)
```
and
```
Arts Agenda iy_main_agenda with methods:
   ppathCalc
   iyEmissionStandard
   VectorSet(geo_pos)
   Ignore(iy_id)
```

I think that there is a design error in how background radiation is dealt with.  In my view, I see no reason why we must pass a background transmittance along to the background radiation.  It would be much more intuitive to instead say that the background radiation and, more crucially, the partial derivatives of the background, are both passed along "as is".  That is, they should be the values as seen by a "sensor" infinitesimally close to the background that are looking at the background.

riclarsson:
@erikssonpatrick There is a bug in your new `Ppath` calculations.  `ppath.pos` can contain values below the surface altitude and values above the atmosphere.  I have had to hardcode a clamp on these values in `ppathGeometric` to work around this for now.  I think it is very important that you clamp these values somewhere else, because we want to be able to check things like "is this point underground" and "is this point above the atmosphere".

erikssonpatrick:
I fully support making the RTE calculations more modular and trying to avoid repetition. But I do not understand how you plan to solve this from the text above.

erikssonpatrick:
Regarding ppathGeometric. Can you be more specific? Are there points clearly outside? Or just due to numerical issues? I did some tests and I got a point at -9.31323e-10 (with surface at 0m). This is due to numerics and we have to live with. This should anyhow be inside the tolerated range of extrapolation. And note that the bottom of the atmosphere and the surface can deviate.
I noticed that you had replaced z_toa with atm_field as input. I am against that. This means that you always need to create an atm_field to calculate ppaths. This makes things less modular. There are applications when you just want to calculate ppaths.

riclarsson:
@erikssonpatrick I am completely against letting a `GIN` deep in a nested `Agenda` control any important aspect of the atmospheric radiative transfer calculations.  You are designing away defaults for the `ppath_agenda`.

The atmospheric field is the only workspace variable that should own an atmospheric altitude.  We decided this in January.

ARTS needs another workspace variable so that your new use-case does not interfere with the use-case of atmospheric radiative transfer.  I suggest we call this either `atm_path` to replace the current main usage of `ppath`, or perhaps `raypath` to name your new use of the `Ppath` struct.  The method and agenda names would change accordingly.

I thin it would be a very bad idea to have your new use-case interfere with the old use-case.

I also do not understand your comment about extrapolation.  It is an error if you try to get values outside of the atmosphere.  The surface value is not so important from this aspect, but it is extremely ugly to allow values under the surface and claim they are part of the surface.

src/rte.cc
```diff
@@ -1887,11 +1887,13 @@ void rtmethods_jacobian_finalisation(
...
         Tensor3 t_data(ppvar_atm.nelem(), 1, 1);
         for (Index ip = 0; ip < np; ip++) {
           t_data(ip, 0, 0) = ppvar_atm[ip].temperature;
         }
-        water_p_eq_agendaExecute(ws, water_p_eq, t_data, water_p_eq_agenda);
+        ARTS_ASSERT(false)
+        water_p_eq_agendaExecute(ws, water_p_eq, atm_field, water_p_eq_agenda);
         for (Index ip = 0; ip < np; ip++) {
           diy_dpath[iq](ip, joker, joker) *= water_p_eq(ip, 0, 0) / ppvar_atm[ip].pressure;
         }
```
riclarsson, line 1899: @erikssonpatrick This is the part where we essentially break the idea behind `water_p_eq_agenda` that I mentioned at the meeting.  Had the `Agenda` only taken the atmospheric temperature or (better in by branch) the atmospheric point values, then this code would probably work better.  But the `Agenda` took `t_field`.  This is a field variable.  It is then using that field variable as if it is just a list of temperatures.  This is why the data has to be mangled so much here that we actually send in something that has nothing to do with it.

One example of breaking this is that you could easily imagine a function also taking the old VMR field and saying that the equivalent pressure is actually changed because we have a lot if IWC in the atmosphere.

I think the cleanest way to redesign this is that `water_p_eq_agenda` only takes a single `AtmPoint` in the future.  Of is the equivalent water pressure different in layer 0 as a function of layer 1?

erikssonpatrick: The saturation pressure of water only depends on the local temperature, and the cleanest solution would be that the agenda takes a single temperature as input. We then need a WSV for it. Should we introduce a_temperature?

PR #616, opened 2023-05-03
Add convenience functions for reading ARTS types in PyARTS

olemke:
Initializing an ARTS type variable from an XML file currently requires two lines of code, creating the ARTS type variable and then calling `readxml`:

```python
v = pyarts.arts.Vector()
v.readxml("my_fgrid.xml")
```

This PR adds a convenience function to create an ARTS type variable directly from a file:
```python
w = pyarts.arts.Vector.fromxml("my_fgrid.xml")
```

The PR also adds a `readxml` method to the `WorkspaceVariable` type in Python to intialize a WSV from an XML file:

```python
ws.f_grid.readxml("myfgrid.xml")
```

### DIFF
--- a/src/python_interface/py_macros.h
+++ b/src/python_interface/py_macros.h
@@ -67,7 +67,7 @@ constexpr Index negative_clamp(const Index i, const Index n) noexcept {
             return x;                                                      \
           },                                                               \
           py::arg("file").none(false),                                     \
-          py::doc("Create :class:`" #Type "` from file\n"                    \
+          py::doc("Create :class:`" #Type "` from file\n"                  \
                   "\n"                                                     \
                   "Parameters:\n"                                          \
                   "    file (str): A file that can be read\n"              \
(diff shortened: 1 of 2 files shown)


riclarsson:
@erikssonpatrick It is not possible to stick to a single name.  You cannot override names in python depending on if you use them statically or not.  At least not in a way that would be acceptable for ARTS.

Basically, the differences between these two is the same as saying 1) `Numeric x = 0;` or 2) `x = 0;` in C++.  A construction takes place when `fromxml` is called (first case in my C++ example) and a reassignment takes place when `readxml` is called (the second C++ example).  These two are not the same feature so they cannot have the same name.

I would recommend you always stick to `readxml` if you cannot remember this difference, this feature just saves you a line of code

riclarsson (approved):
Maybe fix the extra spaces in the macro?  Not so important though, this can just be merged and the macros reformatted next time a change happens in this file.

olemke:
@riclarsson I could add a `return x` statement to `readxml` in the `PythonInterfaceFileIO` macro. Then the chained call
```python
y = pyarts.arts.Vector().readxml("my_fgrid.xml")
```
 would work. Or do you see any negative sideeffects by that?

riclarsson:
@olemke That would probably end up creating copies quite a lot more often than we want it to do.

erikssonpatrick:
OK.

PR #626, opened 2023-05-25
Use configure_file for wigner headers

olemke:
Prevents unnecessary recompilations when re-running cmake.

### DIFF
--- a/3rdparty/wigner/wigxjpf/CMakeLists.txt
+++ b/3rdparty/wigner/wigxjpf/CMakeLists.txt
@@ -12,7 +12,7 @@ check_c_source_compiles("
         return 0;
     }
     "
-    HAS_LONG_DOUBLE
+    WIGXJPF_IMPL_LONG_DOUBLE
 )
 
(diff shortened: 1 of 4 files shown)


3rdparty/wigner/wigxjpf/CMakeLists.txt
```diff
@@ -47,34 +47,12 @@ check_c_source_compiles("
         __uint128_t b;
     }
     "
-    HAS_UINT128
+    MULTI_WORD_INT_SIZEOF_ITEM
```
riclarsson, line 50: Why not make this follow the `WIGXJPF_HAVE_` pattern?

olemke: That would be a question to the wigner developers because that's the name of the constant they use in their code.

riclarsson: Ah, right, it is a one-to-one translation.  Oh well, then this is good.

PR #629, opened 2023-06-14
Add MPM93 nitrogen (dry air) continuum as predefined model

stuartfox:
This re-implements the MPM93 nitrogen continuum as a predefined model, based on a port from the legacy continuum code. I think this is a useful model to have for comparison purposes as it is currently used by AMSUTRAN to generate RTTOV coefficients for ICI.


src/predefined/MPM93.cc
```diff
@@ -0,0 +1,85 @@
...
+  constexpr Numeric fac = 4.0 * pi / speed_of_light;  //  = 4 * pi / c
+
+  const Numeric th = 300.0 / t;
+  const Numeric strength =
+        S * pow((p_pa * (1.0000 - h2o)), 2.0) *
+        pow(th, xT);
+
+    // Loop frequency:
+    for (Index s = 0; s < f_grid.nelem(); ++s) {
+      // FIXME Numeric f = f_grid[s] * Hz_to_GHz; // frequency in GHz
```
riclarsson, line 75: FIXME?  Is it no longer necessary?

riclarsson (changes requested):
The FIXME should be removed if no longer applicable.  If it is a remaining "fixme", it would be nice if it was fixed before merging this :)

Otherwise, nice code!

riclarsson (approved):

PR #632, opened 2023-06-25
Show methods and variables in python docs

riclarsson:
This adds documentation about:

- Methods that can set/take a type to the workspace variable macro
- Variables of a type on the workspace to the same macro
- Methods that operate on/with a WSV to the WSV doc

There is still a bug in the documentation tactic.  It seems sphinx only selects the first method when generating the documentation.  So the pyarts doc now says that methods like `Print`, `ReadXML` and so on only takes `AbsorptionLines`.  I consider this an independent issue.


riclarsson:
Besides segmentation faults on clang...

There is a bug in my `*`-finding algorithm: In for instance `propmat_clearsky_agendaAuto`, the fact that `propmat_clearsky_agendaAuto(use_abs_lookup=1)` should be called is not represented well...

PR #634, opened 2023-07-06
Rename in and out

olemke:
Since `in` is a Python keyword, we need to rename it. For consistency, `out` is renamed to `output`.


riclarsson (commented):
I think this PR should enforce the new rule.  Perhaps we need a file somewhere with a set of forbidden names?  So that new "in" are not introduced

olemke:
From Zoom discussion: We'll introduce precompile tests that will do these consistency checks in a separate PR.

PR #636, opened 2023-07-11
Adaptive columns in sphinx docs

olemke:
Use one column for lists with less than five elements, otherwise 2.


riclarsson (approved):
If it works, I am all for it.  I had problems getting `:columns:` working.  I see you have added more newlines, I wonder if that was my issue.

Anyways, nice!

olemke:
It was indeed. Between options and contents an empty line is required.

PR #640, opened 2023-07-13
Make repr and str print full value

riclarsson:
This makes pyarts print values upon repr and print calls.

Before commit
```python
>>> ws.aa_grid
Workspace Variable
```

After commit
```python
>>> ws.aa_grid
42
```


olemke:
I think we determined at the time that changing `__getattr__` was responsible for the random segfaults we saw in several builds depending on Python version and OS combinations.

PR #641, opened 2023-07-14
Getting started

riclarsson:
This adds some basic information to the getting-started section.

Before full reviews (as we want for examples/), @olemke , I need your help knowing why my "README.rst" is not respected in the folder `1-basics` but in the folder `getting-started`.

I would also like it if the `getting-started` example block was guaranteed to be first in the block of examples.  It is now sorted by names.  Before it was sorted by whatever `os.listdir` returned.  Also, I am not a fan of the numbering of the Listings.  Perhaps we should remove these?


olemke:
All issues you mentioned are resolved.

riclarsson:
Ok, now we have more examples to review.  I am not going to touch this branch for merge before there are two reviews.

@stefanbuehler , @erikssonpatrick , @m-brath , @olemke Please have a look through this.  If you find something minor, please fix it inline (using the "." operation on this page and using the vscode environment github sets up for you).  If you find something major you want to change, please do so.

Since @olemke so excellently fixed that we now have a preview of the webpage, please have a look at the way the webpage will look like by going to "Checks" tab above and download the "Artifact" called something with "sphinx".  You can also find this via the "Actions" tab of the main GitHub top-banner --- this is useful if there has been a small update that changes a file and "Checks" is still running.  In "Actions", you can select the latest green-tick action that represents this branch.

python/doc/source/getting.started.rst
stefanbuehler, line 1: The link seems not to work (on the website there is just plain text, so I would not know how to get to the getting started part. Suggested text:

We recommend starting with the "Getting-started" section under "Examples"
to get started with pyarts.  These guides are run as part of our continous
integration process, so they should always be up to date.

examples/getting-started/1-basics/5-create-an-agenda.py
stefanbuehler, line 1: I didn't understand how the first three methods to set an agenda differed. Is the assignment at the end still needed for example two and three?

examples/getting-started/2-clearsky-radiance/1-basics.py
stefanbuehler, line 1: Perhaps one sensor position is enough? Then add a small plot at the end, the simulated Tb versus frequency?

stefanbuehler (approved):
Looks great! And being able to directly look at the website also is great!

I suggested some cosmetic improvements. One comment is an individual one, outside this review, because I pressed the wrong button.

My only substantial comment is that we could discuss how input files are handled.

olemke:
I tweaked the example headings a bit. `1-basics` is now displayed as `1. Basics`. Also `-` is replaced by spaces so that `1-create-a-workspace` becomes `1. Create a workspace`.
@riclarsson @stefanbuehler: Only downside is that `arts-cat-data` is now shown as`Arts cat data`. Maybe nicer to rename that folder to `arts-catalog-data`?

examples/getting-started/1-basics/3-loading-data-from-file.py
```diff
@@ -0,0 +1,82 @@
...
+ws.example_line_list = pyarts.arts.ArrayOfAbsorptionLines()  # Create an empty object
+ws.example_line_list.readxml("lines/O2-66.xml")
+
+# Call the Workspace Method "ReadXML" to load data from file
+ws.example_griddedfield3 = pyarts.arts.GriddedField3()
+ws.ReadXML(ws.example_griddedfield3, "planets/Earth/Fascod/tropical/tropical.t.xml")
+
+# Call the pure python function "pyarts.xml.load" to load data from file
+ws.example_griddedfield2 = pyarts.arts.GriddedField2()
+ws.example_griddedfield2 = pyarts.xml.load("star/Sun/solar_spectrum_May_2004.xml")
```
olemke, line 68: Instead of offering the user three ways of doing the same thing without giving clear reason when and why to use each version, we should give one example with the recommended way to do it. Therefore I propose to keep the first one as it is the most concise one and remove the other two.

riclarsson: I have removed the other three

examples/getting-started/2-clearsky-radiance/1-basics.py
riclarsson: I think two is good.  I have added a plot at the end, it looks OK to have both

examples/getting-started/README.rst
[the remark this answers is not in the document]
riclarsson: ok

examples/getting-started/1-basics/2-setting-workspace-variables.py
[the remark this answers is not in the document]
riclarsson: I'll open an issue for another branch for this.

python/doc/source/getting.started.rst
riclarsson: It should be fixed now.

stefanbuehler:
I agree.

m-brath (approved):
Looks very good!

PR #655, opened 2023-08-10
Update to pybind 2.11.1

olemke:
New pybind release.


riclarsson:
A clarification that makes `CallbackFunction` and others a bit difficult is:

```
When writing C++ code that is called from other C++ code, if that code accesses
Python state, it must explicitly acquire and release the GIL.
```

I'll merge this as is though we should remember that

PR #656, opened 2023-08-16
Arts3 dev workspace

riclarsson:


riclarsson:
@olemke This can be squash-merged now I think, if it compiles on all the compilers/buildmodes.  Can you have a look and see if there's any big things missing?

(It is still not passing tests, but that's not the state of arts3-dev yet anyways, I think the next push is to fix that problem.  This will be very difficult before arts-xml is converted to arts3 mode though...)

olemke:
There are still some `arts.h` include statements hiding in the shadows.

riclarsson:
> There are still some `arts.h` include statements hiding in the shadows.

Not anymore, but I am not sure what else I broke refactoring some of the netcdf stuff :/

PR #685, opened 2023-10-04
Distribute and document arts_convert script

olemke:


riclarsson:
Good example.  But it is not really confirmed that the conversion still works, no?  Would it be possible to have this confirmed somehow?

riclarsson:
@stefanbuehler , @erikssonpatrick , @m-brath

My review is above.  I am not too concerned if it can be confirmed.  However, as per policy, at least one more of you guys have to look at this and be happy with the example.  @stefanbuehler , especially for you as you requested this in the sharelatex document.

olemke:
> Good example. But it is not really confirmed that the conversion still works, no? Would it be possible to have this confirmed somehow?

The conversion is actually checked as part of the `check-all` target which converts all controlfiles to Python.

m-brath:
I am fine with it.

stefanbuehler:
Instead of "an existing controlfile", better write "a legacy ARTS controlfile (.arts file)". Otherwise looks great from my side.

PR #710, opened 2023-12-01
[ARTS3] Use external pybind11

olemke:
Remove the pybind11 source code from ARTS and use installed version from conda instead.


3rdparty/CMakeLists.txt
```diff
@@ -6,4 +6,4 @@ add_subdirectory(tmatrix)
 add_subdirectory(wigner)
 add_subdirectory(Faddeeva)
 add_subdirectory(gui)
-add_subdirectory(pybind11)
+#add_subdirectory(pybind11)
```
riclarsson, line 9: just remove?

riclarsson (commented):
If it compiles and the tests run through, I think this is an excellent thing to merge

3rdparty/CMakeLists.txt
olemke: Good catch, that is was not supposed to be left there...

PR #711, opened 2023-12-08
add surfaceFlatRvRhEvEh

Shaofei-Wang224:
Hi, all

In this PR, i add a WSM `surfaceFlatRvRhEvEh`.

This method can calculate `surface_rmatrix `and `surface_emission` when emissivity and reflectance are provided externally.

The reason to add this method is that `e+r≠1` in some sea surface emissivity model (such as Fastem and SURFEM-Ocean). 

This method is similar to `surfaceFlatRvRh`, but `surface_ev_eh` is also need specified.

### DIFF
--- a/controlfiles/CMakeLists.txt
+++ b/controlfiles/CMakeLists.txt
@@ -67,6 +67,7 @@ arts_test_run_ctlfile(fast artscomponents/transmission/TestTransmissionWithScat.
 arts_test_run_ctlfile(fast artscomponents/surfacetypes/TestSurfaceTypes.arts)
 
 arts_test_run_ctlfile(fast artscomponents/surface/TestTessem.arts)
+arts_test_run_ctlfile(fast artscomponents/surface/TestSurfaceFlatRvRhEvEh.arts)
 
 arts_test_run_ctlfile(fast artscomponents/faraday/TestFaradayRotation.arts)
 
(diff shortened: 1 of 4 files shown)


riclarsson:
@erikssonpatrick or @m-brath  One of you need to review this.

@Shaofei-Wang224 I like it.  It is a clean implementation.  Thanks for contributing the code :)

I don't do much with the surface though, so I will only merge this after Patrick or Freddy has left me some information.

One thing I lack here is a simple test.  Tests are very important right now and going forward because we are doing quite a lot of changes in ARTS at the moment, and if we do not have tests then things will break.

I think you should have a look at the files in the `${ARTS}/tests` folder and create a short snippet that ensures that this works as intended.  Since the `e+r=1` ratio is what's seemingly the goal and also achieved, I would recommend the test also checks against this.  The simplest possible test is probably something that just sets up the workspace variables required to call your `surfaceFlatRvRhEvEh` method and then asserts the values to 1.  I recommend looking at `${ARTS}/tests/core/propmat/rescale_species.py` for a very simple example.  I imagine yours can be even simpler as your output variables are completely set by the method itself (so no need to wrap any agendas).  I recommend a new folder under `core/` called `surface` and the file-name to be something like `test_flat_rvrh_eveh.py`.

erikssonpatrick:
I had a quick look. 
@riclarsson This method is when e+r do not sum to 1. The existing surfaceFlatRvRh is for the e+r=1 case

As check I suggest to compare output with surfaceFlatRvRh when surface_rv_rh and surface_ev_eh are set with values summing to 1. Sorry, I don't have time to put this in place myself.

Shaofei-Wang224:
@riclarsson I have added a test according to the suggestion of @erikssonpatrick .

olemke:
The macOS Clang build is broken because it tries to compile LLVM from source. Once that is fixed by #712 I'll update this branch accordingly.

PR #714, opened 2023-12-13
Update fastem

Shaofei-Wang224:
Hi all,

I updated the `Fastem` code. Now ARTS `FASTEM` can work like `rttov13.2`. 

Mainly for T3 and T4 emissivity and reflectivity calculation.


erikssonpatrick:
Hi,

Can you elaborate on what has been changed? If I get it right, you have made an update to latest version of FASTEM, and added coefficient file for v1, 2 and 3 of FASTEM. Correct? You have changed the angles used for the test. If using the old angles, are the old values obtained? In short, the status of FASTEM is now as described here?:
https://nwp-saf.eumetsat.int/site/software/rttov/documentation/release-notes/

By the way, have you noticed SURFEM? Would be great to have also SURFEM in ARTS. But nothing I have time for now ...

/P

Shaofei-Wang224:
Hi, @erikssonpatrick 

 Yes, the latest version of `FASTEM` was updated with the following:
(1) the coefficient file for `FASTEM-3` was added to calculate the azimuthal component for Stokes 3/4. For `FASTEM-6`, Stokes 3/4 emissivity calculations are based on the `FASTEM-3` parameterisation. As for `FASTEM-4/5`, Implementation is based Liu et al. (not changed).
(2) updated the Stokes 3/4 reflectivity calculations for `FASTEM-6`. In the old version of ARTS `FASTEM`, the t3/t4 reflectivity are always 0. Now the t3/t4 reflectivity can be calculated by `FASTEM-6`.
(3) updated the angles (including za and rel_aa) in `artscomponents/surface/TestFastem.arts`. With the old angles, the latest `FASTEM-6` can obtian same result as the old fastem. But the emissivity and reflectivity of t3/t4 could not be checked because they are always 0.

As for SURFEM, i have noticed this. In the future, `SURFEM-Ocean` will replace `FASTEM`. Therefore, I think it is a good idea to integrate `SURFEM-OCEAN` into ARTS. I could try to integrate `SURFEM-Ocean` into ARTS, but it might take some time.

erikssonpatrick:
OK. Thumb up from me, after sorting out why some builds failed.

Would be great if you would consider integrating SURFEM. If needed, I could provide guidance on the mapping to ARTS. Presumably, TESSEM can be used as a raw model, that's also an ML model.

Shaofei-Wang224:
Hi, @erikssonpatrick 

After updating FASTEM, the check for TestSurfaceSinglePol.arts failed at `local_stokes_dim=2`.

Can i just add the maxdiff?

erikssonpatrick:
Hi, 
This is a test of internal consistency, so basically this should still work. What difference do you get? Could you print the values of y2 and y0? To see if anything stands out?
/Patrick

Shaofei-Wang224:
With the old FASTEM, the result are:
```
y0: 272.138 265.226 253.054
y2: 272.138 265.226 253.055
y3: 272.138 265.226 253.055
yd: 271.947 264.958 252.65
```

With the new FASTEM, the result are:
```
y0: 272.138 264.966 253.054
y2: 272.138 265.226 253.055
y3: 272.138 264.965 253.055
yd: 271.947 264.958 252.65
```

PR #718, opened 2024-01-10
Propagation path points for geometric views

riclarsson:
This is an implementation of the propagation path point system we discussed in the ARTS design document.  

It also implements three algorithms I have discussed with @stefanbuehler about: appending altitude/latitude/longitude crossings on-the-fly.

The test_path_point.cc showcases usecases (and is run as a test that it all works).

This is mostly based on the existing geometric implementation but I found a few bugs that had to be fixed.   For instance, the old `enu2los` method would make Zeeman/Faraday calculations fail for 180/0 degrees zenith (these will always be very small numbers or 0, so the angles in a path jumps around a lot because of the `atan2` call).

Since the new code uses fixed-size vectors, it should be faster than the old code but I have not confirmed this.

Still to-do: add documentation to new header file.


riclarsson:
@erikssonpatrick This implements mostly what we talked about last ARTS meeting now.  Only for geometric paths so far.  It is about a factor 3 faster than the old `ppath` implementation `ppathGeometric`, and it seems to fail less often.  No tests have been added except for some bare minimum `.cc-test`.

There is one change from the discussion document because I did not understand what we meant.  This path point implementation contains a `pos-type` and a `los-type` instead of `starting background` and `ending background`.

Other than that there is one more thing we did not discuss at all but must be considered:  Should the first position in the propagation path be the origin of the radiation path or should it be at the end of the radiation path?  The `los` should definitely point along the radiation path, this was decided, but the other part I do not know.  I left it as is today since that made porting things a lot easier.  Anyways, it is something we should decide later.  My feeling is that it is probably a good idea to leave it as it is, with the last path-point representing the origin of the radiation, because that will make it more natural to deal with situations such as "stopping" at some distance because of the opacity.

Anyways, here is a script showing how it works now.  It uses both old and new propagation path calculations so that the results can be compared:

```python
import pyarts
import numpy as np

ws = pyarts.Workspace()
ws.jacobian_targets = pyarts.arts.JacobianTargets()


# %% Species

ws.abs_speciesSet(species=[f"O2-Z-66-118e9-119e9"])
ws.abs_lines_per_speciesReadSpeciesSplitCatalog(basename = "lines/")
ws.Wigner6Init()
f0 = 118750348044.712
ws.f_grid = np.linspace(-2e6, 2e6, 1001) + f0

# %% Atm and surface
ws.atm_fieldInit(toa=100e3)
ws.atm_fieldAddGriddedData(key=pyarts.arts.String("t"),
                           data=pyarts.arts.GriddedField3.fromxml("planets/Earth/afgl/tropical/t.xml"))
ws.atm_fieldAddGriddedData(key=pyarts.arts.String("p"),
                           data=pyarts.arts.GriddedField3.fromxml("planets/Earth/afgl/tropical/p.xml"))
ws.atm_field[pyarts.arts.SpeciesEnum("O2")] = 0.21
ws.atm_fieldIGRF(time="2000-03-11 14:39:37")
ws.PlanetSet(option="Earth")

# %% Use the automatic agenda setter for propagation matrix calculations
ws.propmat_clearsky_agendaAuto()

# %% Checks

ws.lbl_checked = 1
ws.Touch(ws.jacobian_targets)

@pyarts.workspace.arts_agenda(ws=ws, fix=True)
def space_radiation_agenda(ws):
    ws.background_radCosmicBackground()
    ws.background_dradEmpty()

@pyarts.workspace.arts_agenda(ws=ws, fix=True)
def surface_radiation_agenda(ws):
    ws.background_radSurfaceFieldEmission()
    ws.background_dradEmpty()

ws.stop_distance_radiation_agenda
ws.surface_field[pyarts.arts.options.SurfaceKey("t")] = 295.

# %% Sensor

pos = [100e3, 0, 0]
los = [97, 0]
#los = [180, 80]
dist = 1e3

# %% Path

ws.rad_pathGeometric(pos=pos, los=los, max_step=dist)
ws.ppathGeometric(rte_pos=pos, rte_los=los, ppath_lstep=dist)


ws.radStandardEmission2()
newrad = ws.rad * 1.0

ws.radStandardEmission()
oldrad = ws.rad * 1.0
```

erikssonpatrick:
Hej,
You are fast! I have not looked at the code; just answering based on your comments here:
- I was happy about separating the equal-distance part and adding grid crossings. It makes things more modular, and easier to add other features later. Any strong reason going back to the old approach?
- "to fail less often", does this mean that there are fails, or that you haven't yet observed any?
- The order of the points is not critical, but it would, in fact, make more sense to have the one closest/at the sensor last. A bit of work to rearrange. You decide. 
- It would be great to remove start_pos and the similar structure members. If we introduce a domain variable, that is given to each point, then we can remove those variables. If the sensor is space, there is one point marked as space, and the integration starts at the next one marked as space-atmosphere boundary.
- Sorry, but don't like the name rad_path. "Propagation path" is a well-established, common name, so would be stupid to abandon that. Following propmat, a fitting name is proppath.
- Out of scope for this PR, but let's discuss the naming strategy at some point. Should we aim for short names, eg propmat, or shall we now aim for more clear names, eg propagation_matrix? I thought we said the latter. And mention that "rad" is ambiguous, radiation or radiance? I suggest using radiance. Even more correct is spectral_radiance. We cleaned up around the radiation fields and introduced spectral_radiance_field, irradiance_field etc. I liked that and then the natural new name for iy is spectral_radiance.

riclarsson:
Hi Patrick,

I will answer inline.

> * I was happy about separating the equal-distance part and adding grid crossings. It makes things more modular, and easier to add other features later. Any strong reason going back to the old approach?

Yeah, it helps methods that requires a gridded field to use the propagation path object more easily.  @stefanbuehler wanted this for these methods.  One example it could help with is if you have a prepared atmosphere, you can guarantee that the point-values of that atmosphere are used in your calculations (helping, if, for instance, you want to do model inter-comparisons.)  Stefan can perhaps give other examples as well?

Note that this is not "going back" to the old approach.  The main code I've added will just do fixed path lengths and nothing else.  There are 4 additional methods that may be used in other algorithms.  All these algorithms do is to append alt-grid, lat-grid, lon-grid, or tangent point crossings to the path (at appropriate positions).

> * "to fail less often", does this mean that there are fails, or that you haven't yet observed any?

Well, there were two failing cases I observed.

There was an issue with `enu2los` in your approach.  For some observation geometries, I managed to observe problems causes by values close to 0 in its use of atan2.  My current solution is to fix the azimuth to the original azimuth angle if all azimuths angles of a computed path are at -90, -0, 0, and 90 degrees (these are the possible 0-atan2 outputs).  This is done by an `apply_options` function that comes with the path calculation methods.

The `ppathGeometric` is also causing fails now in the tests because it generates path-points that are above the atmosphere.  Since the methods using the atmosphere cannot extract data above the top-of-the-atmosphere altitude limit, this means that you cannot use the atmosphere with those path objects.

> * The order of the points is not critical, but it would, in fact, make more sense to have the one closest/at the sensor last. A bit of work to rearrange. You decide.

I will leave it as it is now then.  There must be a computational benefit to change it and I don't see it in the way we do RTE today.  In fact, even for a potential "opacity-cut" method, this is the order we would want anyways.

> * It would be great to remove start_pos and the similar structure members. If we introduce a domain variable, that is given to each point, then we can remove those variables. If the sensor is space, there is one point marked as space, and the integration starts at the next one marked as space-atmosphere boundary.

The `start_pos` is gone here.  We have of course a "start position", and for the radiation that is the last point of the `ArrayOfPropagationPathPoint`.  I've defined that point to have a `pos_type` of `atm` and a `los_type` of `space` or `surface` now.  The "end position" is the first point of the `ArrayOfPropagationPathPoint`.  I did not see a reason to add a `sensor` type as its position.  It is of course a sensor if that's what you have but quite often it is not.  It's `pos_type` is therefore `atm`, `surface`, or `space` and its `los_type` is `surface`, `space` or `atm`.  I've opted to remove the path point describing the transfer from the top-of-the-atmsophere to the "true" start position given by the user because it just introduces a lot of overhead when extracting atmospheric data to keep it.

> * Sorry, but don't like the name rad_path. "Propagation path" is a well-established, common name, so would be stupid to abandon that. Following propmat, a fitting name is proppath.

I don't care about this name.  `proppath` works.  I'll leave it like this for now until we have a proper discussion on names.

erikssonpatrick:
One question about what we discussed today. How do you set the tangent point? From what I remember, we don't have any working function for that in the general case of a spheroidal reference geoid.
I can add that it is not critical to include the tangent point. Even with a step length of 10 km, you miss the lowest altitude with just 2 m in the worst case.

riclarsson:
The limb point is numerically determined.  The algorithms are relatively slow but should be accurate down to numerical levels.  I say algorithms because there's a different one for finding the zenith angle for the tangent altitude for a given position, and one for finding the tangent point in an existing path.

Neither accepts paths that don't have a tangent point.  Both bisects the distance to find the tangent point.

The "find-the-zentih-angle-for-limb-at-altitude" method (`geometric_tangent_zenith` in the code) increases and reduces two zenith angles until the method that determines the two ellipsoidal intersection points is valid for one zentih angle but not the `std::nextafter` zenith angle.  It returns the valid one.

The "find-the-limb-in-path" algorithm finds the distance between two points where the zenith angle is either exactly 90 or where the `std::nextafter` distances make further iteration meaningless.

Also, note that these may just be accurate down to numerical levels, I mean in the context of the rest of our system.  The core path calculations still rely on many more methods that are not accurate to numerical levels and I can only keep at their level of accuracy.

erikssonpatrick:
"find-the-limb-in-path" sounds just fine.
For "find-the-zentih-angle-for-limb-at-altitude", don't you not also need to consider azimuth angle. It varies along proppath. So I get it to a 2D search to be done.
If you have not noticed, there is an old, non-actibe function for this in geodetic.cc, geomtanpoint. I would at least suggest to do as in that function, to have special code for a spherical planet, to save time. The general code works fine in most cases, but can fail around the poles. Deviations up to 30 km, if I remember correctly. That is why I don't use it. But maybe you can fix it?
And feel free to ask before you start on things like this. Could save time for both you and me.

riclarsson:
You have to give a position that's above the tangent altitude and an azimuth angle to choose what part of the great circle you intersect with.

If you want a special function for spherical planets, you can add some paths to the code.  The default in ARTS3 is not spherical but ellipsoidal.  I want the default case to work, if you add extra paths a lot more testing-code must be added.

PR #719, opened 2024-01-24
Mocking scattering data interface.

simonpf:
This PR aims to sketch out the ARTS integration of the new scattering data handling.


simonpf:
@riclarsson  As per out discussion from last week, I started replacing ParticulatePropertyTag with ScatteringSpeciesProperty as defined in src/core/scattering/properties.h. Could you have a look a that as well as the changes in atm.h? Thanks!

riclarsson:
Nice!

Two things.  The first serious, the second a minor design choice (that also has a bug).

The big problem first:
How does `scattering`, the CMake library, depend on `artscore`?  I really think it is a bad idea to do this.  We want less cross-dependencies to make all this work nicely in the end.  (At some point, `artscore` might also depend on `scattering` since we want some internal functionality using multiple parts of core ARTS functionality.  So it is expected to cross-pollinate, which could cause issues.)

The second minor thing:
I think there's a bug in `ParticulateProperty toParticulatePropertyEnumOrThrow(const std::string_view x);`  The `.cc` file defines `ParticulateProperty toParticulatePropertyOrThrow(std::string x);`.  The `ENUMCLASS` macro gives you `ParticulateProperty toParticulatePropertyOrThrow(const std::string_view x);` already.  For `SpeciesEnum`, we also want short names, like you have in the `.cc` file.  We definitely want short-names for `SpeciesEnum`, but are they really a requirement here?  (Several people of RTE community will much more easily recognize `C2H2` than `Acetylene`, but I cannot make `C2H2` the `enum` name because we must support things like `NO+`, which isn't a good variable name [and workarounds like `NOplus` will add more confusion].  I do not see this as such a big deal in the `ParticulateProperty` `enum`, at least not with the names you have there today.)

riclarsson (dismissed):

riclarsson:
@simonpf This does not compile so merging main arts3-dev is not good.  There seems to be some problems with duplicate `psd.cc` files.  We have to hold merging until we can compile it.  (Or we will probably introduce even greater bugs.)

simonpf:
So, this should *in principle* build. Could you add fftw3 to the build environment @olemke ?

riclarsson:
@simonpf I spent some time fixing up so that it works on Mac and the documentation is generated.  I hope it works online now.

We don't see an easy way to make shtns work on Windows.  Lots of unix-commands with regex in the build process...

We need to encapsulate the use of shtns to pure CC-files.  This is because we want to be able to build ARTS in LGPL mode.  So the same is probably true for FFTW3.  We should just call those libraries from CC-files and not include them in headers.  I don't know yet what limitations that actually imposes on your code.

We are thinking how to work around it here.  If there's another spherical harmonics library we can use. But for now we must encapsulate things to make the things that may work, work on Windows.

riclarsson:
@simonpf It seems it does not build online.  @olemke will have a look how this notebooky thing is done in typhon.

I also noticed that there are some other problems with the documentation.  If you build the `pyarts_docs` target, you will get a lot of text `WARNING: [ARTS DOC] No short doc found:  some-type-or-method`.  These short docs are generated *manually* by a script of mine.  (Nanobind types were not recognized correctly by sphinx.)

Anyways, if you do manage to build the docs themselves, you can go to one of your classes, say `PhaseMatrixDataTROGridded3` and see what a missing documentation looks like.  You have two types of missing documentation, one more severe than the other.  Generally, if you see a `->` arrow, the signature of the method rather than the documentation is displayed.  This just means that there's no documentation.  A few short words on what things like `to_spectral` do would be nice.  The more severe problem comes from methods like `get_za_scat_grid()`.  You see a named type in its signature return `scattering::LatitudeGrid`.  This means the type is not defined at all on the python side of things.  It must be.  Calling that method will otherwise just result in an error.  (The local link to `PhaseMatrixDataTROGridded3` is `arts/build/python/doc/build/pyarts.arts.PhaseMatrixDataTROGridded3.html`, but it is generally a good idea to look at all the classes you bind yourself.  Note that if a C++ `std::`-link is shown, it probably means you forgot to include something like `nanobind/stl/variant.h` or whatever the type that is missing.)

riclarsson:
@simonpf I've now encapsulated FFTW3 and SHTns.  I had to move some code away from headers and into CC-files.

Doing this, I noticed there is a lot of code in headers for non-templated classes.  I think these should to some degree be moved into CC-files to limit the number of recompilations.

We are discussing if we should merge this as-is.  Let us know before Monday morning CEST if you object.  The remaining issues must of course be solved but at least we will stop your stuff diverging from main.

PR #760, opened 2024-04-23
Cpp - disort

riclarsson:
This is a reimplementation of the PythonicDISORT solver.

It still lacks clear citation of the original software.  This must be done as part of the user-facing interface, which is not yet part of this draft.

Remaining work:
1) Make user-facing test.  Method and all.
2) Run speed comparison versus cdisort in the 2.6.x branch.
3) Remove cdisort from ARTS3.
4) Remove the GSL Legendre Gauss Quadrature as a new open-source implementation that is not GPL is used.
5) Add user-facing citation for the new Legendre Gauss Quadrature scheme.
6) Get approval that the inclusion of Boost in ARTS3 is OK.


riclarsson:
As a note for testing. In this iteration the following things limit what is tested:

1) `b_pos` and `b_neg` are always single-values
2) The Fourier error is not tested and will throw if used
3) `f_arr` is never empty in TMS-corrections. (Honestly, TMS correction should perhaps not be called then.)
4) There are no BDRFs when there's a source term.

The test follows what PythonDISORT tests as of April 2nd 2024 and should be updated before merge to be with the unreleased version.

riclarsson:
This now contains enough improvements that it can be merged in my opinion.

I cannot continue with this in an effectively.  It requires integration with the overall ARTS still, which is much more your expertise for DISORT-y stuff, @m-brath .  Of course, I am around for helping.  But I cannot effectively do the integration myself.  It's a little bit of a chicken-and-egg problem with the scattering update branch since so much code might have to be switched later.  As the test-file suggests, this at least produces the same results as the original algorithm and the same results as cdisort.

@olemke , I also removed Eigen with this update and made us use the distributed version instead (same version).  This removes quite a few errors as we don't report on bad code warnings in external libraries.

PR #770, opened 2024-05-27
V2.6.x addsungeo

m-brath:
Add a WSM to change the geometric varaibales of a sun.

### DIFF
--- a/src/m_sun.cc
+++ b/src/m_sun.cc
@@ -192,6 +192,7 @@ void sunsAddSingleFromGridAtLocation(
 
 void sunsChangeGeometry(ArrayOfSun &suns,
                          // Inputs:
+                         const Index &suns_do,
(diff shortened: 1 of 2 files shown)


riclarsson:
First of all, this is not a bug fix but a new feature in 2.6.  I do not think it should be added to 2.6.  Are you or anyone else really using `.arts` files for 2.6 simulations?  Because all that this is doing here can already be done via the python interface of 2.6, and none of it depends on any workspace variables so it is not useful from workspace-flow either.

m-brath:
I agree that the geometric properties can changed using python, but it it is not recommended to use python code inside an agenda. This function is needed, if you want to change the sun position or any other geometric property of the sun within an agenda like dobatch_calc_agenda or ybatch_calc_agenda.

PR #785, opened 2024-07-02
[ARTS-2.6] Fix wigner getting stuck

olemke:
Wigner gets sometimes stuck in an infinite loop in `pexpo_prime_factor` due to a negative `fpf` being passed in. We don't know why/how these negative values come into existence, but enabling `PRIME_LIST_USE_VECTOR` to use vector types / instructions such as SSE / AVX for the exponents, fixes it. So we go with that for now to fix `TestMolOxyAdaptation`.


olemke:
Superseded by #786

PR #814, opened 2024-09-05
Put the wrapper for the WSV in a separate header seen only by a few cpp-files

riclarsson:
The idea is that there should be no way now to set a WsvValue directly. We want this because we want the dtor to not be created.  If it is created, every file that somehow need to access auto_wsg.h is getting a dtor, or 4-5 seconds longer compilation time.  This includes wokrspace.h, which is used extensively.


riclarsson:
@olemke , can you run your script and see if the long dtor for the Wsv Variant is gone from the artsworkspace target on this branch?

PR #820, opened 2024-09-13
Bug fix: the sun should be in spectral irradiance, not as documented …

riclarsson:
…in spectral radiance


m-brath:
Nothing to add

PR #852, opened 2024-10-24
Prototype of the scattering species interface.

simonpf:
Adds a HenyeyGreensteinScatterer that can be used as a scattering species and provides functionality to calculate bulk scattering properties.


olemke:
@simonpf A note on notebooks: Don't commit the notebooks including their output. They need to be clean before committing to keep the size small and more diff friendly. Select "Edit"->"Clear output of all cells" in Jupyter before committing.

PR #854, opened 2024-10-28
Partly port old Lookup calculations

riclarsson:


riclarsson:
@stefanbuehler When you read this after your holiday, we should have a quick talk on what kinds of tests are required to make this port work as expected.  The original ARTS2 code was so over-the-top complicated (it seems to have merged several grids into a single one even) that I just put together the smallest possible interpolation routine I could here.  This means I am probably missing some edge cases, but as I do not know them, this is the state I can get to.

There's just one test running here, and it's for a simple CO2 absorber.  More tests must be added because it tests almost nothing of the code to run it in this manner.

stefanbuehler:
Please don’t mess with the interpolation, there is a lot of evolution in it. AbsLookupExtract() or whatever the function was called exactly is reasonably self-contained, I believe. I would just use it as is. Same for the setup functions, I believe.

riclarsson:
That's not possible.  The old code is not useful directly.  Concepts have to be ported.  For two examples, we neither have SpeciesTag nor do we have the concept of VMR indices in ARTS 3 absorption calculations.  So the idea of computing a species grid of absorption data of ARTS 2 is not going to work exactly the same way in ARTS 3.  In short, that function is not directly portable.  It is not self-contained.

Please point out how I "mess" up the interpolation by providing reasonable tests that I can put in the CI to reproduce old results.  A test that would be useful is if I can know that LUT calculations are expected to provide a "1%" error cf LBL when trained on AFGL tropical atmospheres when run with "7-order" polynomial interpolations in an AFGL subarctic winter atmosphere offset by "20 K" in both directions with a pressure profile adjusted for hydrostatic equilibrium. The quoted numbers are the numbers I do not know: 
- 1%: I do not know what the expected error for this is in ARTS2 (I never used LUT in ARTS2, so I cannot reproduce it).  What is the expected errors?
- 7-order: I do not know the polynomial orders of the individual components in ARTS2.  I do not know how dense the training should be for this to work. This and the first number limits that.  What are the 4 interpolation orders?
- 20K: I do not know what a reasonable training range in temperatures is considered in ARTS2.  Several runs with sampled +-20K up and down should covers this range?

PR #862, opened 2024-11-11
Enable and run some simple test cases for a Henyey-Greenstein scatterer in Disort

riclarsson:
The testfile henyey-simple.py runs the test scenario.

I almost expect this to fail on the LGPL branch because Heyney-Greenstein depends on SHT features.  I am quite sure this is not required since all you need for this is `pow(g, l)`, so hopefully the code works anyways.

The branch is not in a good-state as is at time of creating this draft-PR.  Several things need to be addressed wrt the design and implementation of the scattering species to improve things.  I will list them in no particular order:

1. Using multiple core with SHT in the main branch is not supported.  I added here a line `#pragma omp critical` in sht.h to work around the problem for now.  However, the entire design that requires this workaround should be fixed.  The problem is that we have a global `std::map` to begin with.  For other global data that must exist in ARTS (e.g., `PREDEF` and `Wigner`) we either have static data that is constant, or specialized workspace method that initializes the data block and allows clearing them.  The second approach seems what we should go for for SHT global data.  Or perhaps even better, to not have any global data.  Having data just appended to a global variable in deep code is not a clean approach.  What should we do to fix this?
2. The normalization of the Henyey-Greenstein phase-matrix is not numerically invertible.  I need the vector `[g**l for l in range(degree+1)]` for Disort.  But the normalization gives `[x * g**l * sqrt((2*l+1) / (4pi)) for l in range(degree+1)]`, where `x` is the scattering cross-section.  Anyways, getting `g**l` back is not possible for large `l` in this normalization (numerically, it is not stable).  I am not sure this is a problem we can solve.  For now I just `v = std::clamp(v, -1.0, 1.0)`, but that's not good if the underlying scattering species has bad data.  So we need to be careful about how other species are added here, because we must be able to somehow guarantee that `v = std::clamp(v, -1.0, 1.0)` is not fixing bad data, just numerical errors.  Is it possible to just return this without the `sqrt(2*l+1)` factor?
3. There seems to have been some math problem in the Henyey-Greenstein code.  I have changed a division by 4pi that seems to have repeated itself for too many indices.  Did it mess up the code?
4. ArrayOfScatteringSpecies now inherits publicly from the vector.  This allows `ws.scattering_species = [spec1, spec2, ...]` in python initialization.  Was there any reason not to inherit publicly?  I am also a bit sceptical it is a good idea to inherit instead of just having the vector as a member variable.  I don't see the benefits.
5. Why is it that each inidividual species has a temperature and frequency grid?  To me it seems that the grid should live on the bulk-properties instead, and that the inner parts should be kept simpler.  Also, why temperature grid instead of atmospheric point grid?
6. I had to change `ExtSSACallback` to be of the templated type `CallbackOperator`.  This is because passing a python operator into ARTS is not thread-safe.  By wrapping it in a `CallbackOperator`, we can enable the GIL as required during the call if the type was initialized from python.  If you initialize the callback operator from C++, no GIL reactivation is required.
7. I removed some instances of iostream.  We should avoid having it in headers, it adds a lot of compile time.
8. The compile time warnings are insanely numerous.  I am not sure how or when they are created, but somewhere in the code somethings are not initialized properly.  This fills page-wise of GCC warning reports (though clang believes all is good).  The core must be identified so we know which compiler is wrong.
9. I've added the required 6 spectral workspace variables as ARTS types.  The code just runs for stokes_dim=1, but that ought to be easy to fix.  The names, however, are quite long to follow with what we want these things to be named as (long explicit names).  I think we should reconsider this naming scheme but only once we know what the 24 new workspace variables should be called.  Using TLA:s like TRO and ARO in WSV:s creates a really annoying to understand idea, like this sentence, so the naming is important.
10. The agenda `propagation_matrix_scattering_totally_random_orientation_spectral_agenda` and its defaults shows how to execute the Scattering Species code for Disort.
11. The agenda `disort_settings_agenda` with the option `ScatteringSpecies` shows how to setup Disort using clearsky and scattering species calculations.


riclarsson:
Addition:  This PR also leaves ` cpp.fast.scattering.test_phase_matrix` broken.  I am not sure how.  It "works" in release build but an assert is activated in reldeb.

simonpf:
The thread safety is an issue because SHTns --- at least as far as I understand --- is not thread safe. SHTns pre-computes coefficients for a certain configuration  such as degree (l) and order (m) in a configuration step after which it is ready to use. However, all of these are inside SHTns and not isolated.

I introduced the SHTProvide as a way of managing the current configuration in order to avoid reinitialization of the SHT library if the configuration between consecutive calls isn't changed. I not the biggest fan of this design either so if you have a better suggestion please let me know.

simonpf:
Regarding 5: The reason for the temperature grid is that the scattering data varies with temperature. Therefore, prior to calculating the bulk properties for a specific atmospheric point, this dependency is represented using a temperature grid.

riclarsson:
@simonpf I merged this since it didn't have anything crazy in it.

The multithreading issue this time was that you were writing to a std::map (of some kind) in a situation I wanted to run in a parallel loop (extracting data at multiple atmospheric grid points).

We have two options to solve this issue: 1) make the global data impossible to modify in extraction methods, or 2) make the `ScatteringSpecies` own/share their SHTns data.

For 1), we need workspace methods like `SHTnsAdd(degree, order)` and `SHTnsInit()`.  Then the extraction inside deeper calculations would just access the global map purely as if it is constant (read-only).

For 2), we need workspace methods like `scattering_speciesPrepareTROSpectral` to compute the required SHTns data and make each individual scattering species element own/share a SHTns pre-computed object.

I prefer 2) since a key element everyone keeps talking about is how you can only consume properly prepared data.

Anyways, if SHTns is fundamentally thread-unsafe, all calls to it must be done in some kind of mutex-block.  I am not sure where it is called, but this must be encapsulated in a very stringent manner.

Regarding point 5), it really sounds again like the reason to have a grid for these fields confuse the consumption of the scattering data with the storage of it.  We do not need to have a temperature grid if we just want to consume the data at a single atmospheric point.  If I am not misunderstanding this, we should more strongly disconnect the output of the combinatoric {aro,tro,general}/{spectral,gridded,double_spectral} methods from how the data needs to be kept in memory or file to allow consumption.

PR #866, opened 2024-11-20
Extend nb stubs to all arts submodules

olemke:
Adds completion support for all nb submodules inside the arts module.
.pyi files are now written to an `arts` subfolder. Getting access to the new completions requires deleting either the whole build directory or removing `build/python/src/pyarts/arts.pyi` manually.

### DIFF
--- a/src/python_interface/CMakeLists.txt
+++ b/src/python_interface/CMakeLists.txt
@@ -164,11 +164,19 @@ set_target_properties(
     pyarts_cpp PROPERTIES LIBRARY_OUTPUT_DIRECTORY
     "${ARTS_BINARY_DIR}/python/src/pyarts/")
 
-nanobind_add_stub(
-    pyarts_cpp_stub
-    MODULE arts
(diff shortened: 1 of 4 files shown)


src/python_interface/CMakeLists.txt
```diff
@@ -164,11 +164,19 @@ set_target_properties(
...
+  file
+  globals
+  hitran
+  interp
+  lbl
+  math
+  physics
+  predef
+  rtepack
+  zeeman
```
riclarsson, line 181: This is going to be a nightmare.  Can you add a small pytest somewhere also that the list is exhaustive?  We can loop over all of pyarts.arts and below to check for if the named thing is a module, I think.  Well, I am doing it in the documentation file.

olemke: I've added a test that outputs an error like this when a module is missing a `.pyi` file:
```
E       AssertionError: Missing stub file(s) for arts submodule(s):
E           hitran
E         Add them to the `arts_add_cpp_stubs` call in src/python_interface/CMakeLists.txt
```

riclarsson (approved):

PR #881, opened 2024-11-27
Make it possible to fit an additional error vector on a custom grid

riclarsson:
This reenables the idea that F(X) = Y + E in OEM.  Custom functions for E must be provided.


riclarsson:
With the additional benefit that we now fully support multiple sensors in the same manner as ARTS2 only supports using a single sensor (for E, ARTS2 can only use one fit)

PR #882, opened 2024-11-27
Add Complex Muelmat and the two matpack versions needed by TRO spectral

riclarsson:


riclarsson:
@simonpf There's now ComplexMuelmat as a 4x4 matrix for dealing with the spectral shtns-stuff.

The name should probably be fixed since Mueller calculus is inherently real, but I'll keep it as this for now so it is consistent with the other type.  Name suggestions are appreciated!  I am not yet sure if you keep absolute phase information in your spectral matrices or not.

PR #883, opened 2024-11-28
Add missing test directories

olemke:
Tests from new subdirectories were not collected.

@riclarsson The test in the hitran subdir fails as it's using a local path. Please fix.


olemke:
Will refile this with a more fundamental rework of how we collect tests.

PR #886, opened 2024-11-29
General TRO spectral scatterer

riclarsson:
This adds the most generic TRO spectral scatterer to Scattering Species.  It is must be fully implemented by the user.  I also add it as a test to the existing simple-test case to show that it is possible to fully pass python functions to it


riclarsson:
@simonpf This branch shows what I was talking about during the meeting.

The new files contain the structs for more limited return types I was after, and also the full implementation of the general scatterer for tro spectral.

It also wraps the visiting functions in scattering_species.cc in a way that you don't have to implement all of the interface for every type.  E.g., a type that cannot be TRO does not need the TRO interfaces.

riclarsson:
Had to make the mutex change as part of this branch because there was a kernel panic somewhere (only in CI, I cannot see it locally, so it might be a very narrow problem)

PR #908, opened 2025-01-14
Speed up pyarts startup when all you need is XYZ

riclarsson:
This mimics numpy's import style for submodules.  Startup cost of pyarts scripts that use only pyarts goes down from 0.75 s on my machine to sub 0.1 s.  Mileage may vary.


olemke (approved):
Really nice speed up. :-)

PR #944, opened 2025-05-13
Rename and add openings

riclarsson:
Renames some document files to better reflect results of March discussion.

Add openings where missing, including for the example folder.

We need to consolidate the documentation somehow.  It is very confusing what things go into what folder (top level doc, examples, or python docs).  I don't think this is a long term good idea.  It works for now.


olemke:
I agree about consolidating the files. At least the `doc/arts` stuff should be relocated to `python/doc/source`. Not sure about the examples though.

riclarsson:
@olemke should we do this as part of this PR or not? I'm honestly not sure how much work it is to do this refactor.

olemke:
Nah, let's do that in a separate PR.

PR #953, opened 2025-06-04
[ARTS-2.6] Force openmp openblas in linux dev environment

olemke:
The `environment-dev-linux.yml` didn't explicitly depended on the openmp version of libopenblas. The pthread version, which is installed by default, causes significant performance degredation.

Other fixes:

Added `LIBCPP_DISABLE_AVAILABILITY` to fix compilation error on macOS with newer libc++.

Add `conda-remove-defaults` flag in CI to completely disable the Anaconda channels. There is still a warning about the `defaults` channel in the build logs, but that's due to the order the `setup-miniconda` action does things. It first adds the `conda-forge` channel (which generates the warning about defaults being present), then removes the `defaults` channel.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -91,6 +91,7 @@ jobs:
         with:
           miniforge-version: latest
           channels: conda-forge
+          conda-remove-defaults: "true"
           channel-priority: true
           activate-environment: arts
           environment-file: ${{ matrix.devenv }}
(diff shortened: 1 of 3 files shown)


olemke:
> @olemke I would want the same environment updates in ARTS3 if you can port it? I just earlier today noticed that my system was experiencing this issue as well.

Yes, of course.

PR #956, opened 2025-06-05
Make sparse sensor possible

riclarsson:


olemke:
As a quick and dirty test, I modified the `xml_read_from_stream` function for `SparseStokvecMatrix` to read each value separately (lines 2840) and added code to reset the stream flags if a subnormal number was read:
```cpp
      is_xml >> irow;
      is_xml >> icol;
      is_xml >> data.I();
      if (std::fpclassify(data.I()) == FP_SUBNORMAL) is_xml.clear();
      is_xml >> data.Q();
      is_xml >> data.U();
      is_xml >> data.V();
```
Then the file can be read without error. The question now is how to deal with this properly...

olemke:
Let's see if that works around it for now. It is a shame the there is [a bug report open for libc++](https://www.github.com/llvm/llvm-project/issues/38360) since 2018 about this but nothing has been done to fix it.

riclarsson:
@olemke it probably works because of reading to string and then parsing the results now.  I'll merge. It would be nice to delete the double istream from the standard so we always can go via imanip.

(The low numbers of course means that they should be removed.  But that's a separate issue I don't think we should be spending time on. So I'll merge now)

PR #963, opened 2025-06-19
Change XFIT so reldeb builds work again in CI.  Add reldeb to CI.

riclarsson:


riclarsson:
@olemke I need you to confirm the changes in xsec_fit.cc.  I think there was a classic off-by-one in the checks.

olemke (approved):
Thinking through how `lower_bound` and `upper_bound` are used to determine the overlap between `f_grid` and `data_f_grid`, I agree that the comparisons were incorrect before.

PR #967, opened 2025-06-24
Fix missing doc

riclarsson:


olemke:
It seems they're currently updating the runners. On my fork, I get a newer runner image (20250617.1.0) with MSVC 19.44. But even restarting the job here still got me an old runner image (20250602.1.0). Guess we'll just have to wait a bit and retry.

PR #975, opened 2025-07-14
Make clang20 work on Mac

riclarsson:
This makes CLANG 20 work on mac-arm.

There were two problems.  I simply brute-forced the solution.

1. They seem to have broken ``std::from_char`` for integers.  I get errors all over the place for it.  I did not want to bother with this, so I switched to using ``boost/charconv.hpp``.  Including ``charconv`` at all breaks this.  I am not sure who broke it, since the error message links to conda-forge stuff.  Anyways, it is broken and so I removed it
2. They did break how they defined ``std::max``.  I think this is purely on clang itself.  ``std::max`` is defined to just call ``operator<`` and do the right thing.  However, clang-20 no longer does that it seems.  In the line-mixing code, I use this to find the greatest quantum number to allocate a table in the Wigner calculator.   Because of the clang-20 changes, this silently uses something other than the ``operator<`` for ``Rational``.  This error may happen elsewhere in ARTS, or perhaps we have used ``std::max`` somewhere else where they break things, but for now the tests work locally for me (well, bar the fact that the XML code must merge.)


olemke:
My fix/workaround in ARTS 2.6 was to add this (as proposed by conda) to CMakeLists.txt for macOS:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 92368b17b..d9b5f9320 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -412,6 +412,7 @@ set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${ARTS_CXX_FLAGS} ${ARTS_FLAGS}")
 if (APPLE)
   set (OS_NAME "Mac OS")
   set (OSX 1)
+  add_definitions(-D_LIBCPP_DISABLE_AVAILABILITY)
 elseif ("${CMAKE_SYSTEM_NAME}" STREQUAL "CYGWIN")
   set (OS_NAME "Cygwin")
   set (CYGWIN 1)
```
Applying the same to ARTS 3 makes it compiles, but then I get an endless list of wigner errors in lbl.ecs_co2, ecs_o2 and lbl.ecs_o2:
```
wigxjpf: More iterations (13) than allocated in temp array.
wigxjpf: More iterations (13) than allocated in temp array.
```
Let's see if your fix works better than my workaround. :-)

riclarsson:
@olemke I tried adding that but kept getting weird error.  If conda is fighting clang with those options, we should perhaps be neutral and not use what they are making difficult.

The endless loop comes from the fact that someone messed with ``std::max``.  It returns random numbers.  It is simply broken in clang20.  We should perhaps switch to manual ``max`` functions everywhere because of this, but this was the one that caused those loops.

riclarsson:
@olemke Ok, it seems to still not be working.  But it is impossible to know before the XML branch is merged since most tests - perhaps even all - are failing because of that.  So I will leave this here for now.

PR #979, opened 2025-07-17
Quantum revision

riclarsson:
This revision of the quantum numbers make it a lot easier to use them in python and in C++.  The type-safety comes from a very small variant that keeps track of things as compared to the old code that had a very complicated manual union keeping this information.

I wanted to test the quantum numbers handling a bit more and the old code was a bit of a mess.  There is something else fundamentally broken in clang-20 on MacOS.  The tests here might pass, but I can randomly see the bug where I use 60+Gb data for the CO2 ECS test.  I think that's what is causing the problem on Mac-Intel.

I ran all three of the address, memory, and undefined sanitizers to see if it would give anything, but they don't give any information about an error.  I think copying the data helps, but there is something very weird going on.

### DIFF
--- a/src/core/lbl/lbl_data.cpp
+++ b/src/core/lbl/lbl_data.cpp
@@ -83,7 +83,7 @@ std::pair<Size, std::span<const line>> band_data::active_lines(
 Rational band_data::max(QuantumNumberType x) const try {
   Rational out{std::numeric_limits<Index>::lowest()};
   for (auto& line : *this) {
-    auto& qn = line.qn.at(x);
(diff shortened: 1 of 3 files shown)


olemke:
> There is something else fundamentally broken in clang-20 on MacOS. The tests here might pass, but I can randomly see the bug where I use 60+Gb data for the CO2 ECS test.

I propose that we put back the `<20` limitation into the `environment-dev-mac.yml` to avoid this random behavior until the clang-20 compiler issues are sorted out upstream.

PR #993, opened 2025-08-15
Lagrange interpolation refactoring

riclarsson:
I came across an issue with the cyclic interpolation.  I deemed it easier to reimplement the whole Lagrange library to fix the issue that way since the old code was a bit messy.

I removed the interpolation derivatives to simplify the code.  (These can be added back but I do not believe that to be a good idea.)

I removed the named approach to scaling and cyclicity and instead am using a functional approach.  This should allow extending it quite freely.

The new code is also quite a lot faster than the old (probably because of the simplifications to the code itself).  The new test_laginterp.cpp file even has its linear interpolation faster than GridPos, which was definitely not the case before.  You can see that in the timing information when running that file if you compare the output that reads ``res``, ``itw``, and ``gridpos`` from the test-file.

- ``reinterp dynamic all-in-one linear order 1, identity-identity`` and ``make_lags dynamic all-in-one linear order 1, identity-identity``, or
- ``reinterp fixed all-in-one linear order 1, identity-identity`` and ``make_lags fixed all-in-one linear order 1, identity-identity``

The old approach takes about 320 microseconds, the dynamic method takes 340 microseconds, and the fixed takes about 250 microseconds.  I think it might be worth to have a look at other machines if it is worth removing the old code.


erikssonpatrick:
Great. Thanks.

riclarsson:
I'll merge this then.

On an independent note, we should fix how we deal with data at some point so that we can choose data version in the tests somehow

PR #998, opened 2025-08-27
Add cdisort interface

olemke:
Add cdisort in addition to cppdisort for testing and validation.

`ws.disort_spectral_radiance_fieldProfileCdisort` can be used as a drop-in replacement for `ws.disort_spectral_radiance_fieldProfile`.

While overall agreement is very good, there are still some rare cases that can produce 10% relative difference at very shallow (close to 90°) angles. Our best guess currently is that this is inherent to numerical differences in the two algorithms. Since shallow angles are the least favorable case for the disort method, this may further contribute to amplifying the differences of the two implementations.

The python script in `tests/core/cdisort/` is provided as a starting point for comparisons between the two implementations of disort.


riclarsson:
Never mind, I ran the code for increasing number of quadratures.  The new code is faster (and gets relatively faster the more quadratures for some reason).

For Friday: Are the relative differences for these angles acceptable?

olemke:
> Nice! But I think you should rename `cdisort_spectral_radiance_fieldCalc`. Perhaps `disort_spectral_radiance_fieldCalcCdisort` (so that tab-completion shows both options). We claim we should name methods after their main output if they have one, and that's still `disort_spectral_radiance_field`.

You're absolutely right! Just adding the `c` was so tempting that I forgot about our naming scheme. ;-)

> Is it still true that they are about as quick?

cppdisort is about 6.5x faster than cdisort.

> Have you tried identifying what happens at high zenith or are we simply fine with the differences?

We looked through the code but it's hard to guess where the differences come from. We're are pretty certain that it is not coming from errors in the setup of cdisort. We thought we leave it for now and move on to see how it impacts flux and scattering calculations once we have a setup for those, keeping this problem in mind.

> Edit: Also a point for Friday: should meta-methods really have authors? That's your method but you are using my methods within it. Should the author-list just be composed of the authors of the other methods? To be clear, I am already appending other authors to the method, so your documentation here will read `Richard Larsson, Oliver Lemke, Richard Larsson, Oliver Lemke, Automatically Generated`. (It is still a vector, so using comma-separation in the string leads to more authors. And I put `Automatically Generated` for methods whose generation I automated - which perhaps also should be changed.)

Good point. I just saw the author field and added my name, wasn't aware that it is already so smart to accumulate the authors of the called methods. Maybe then there is indeed no need for explicit authors for meta methods.

PR #1000, opened 2025-09-01
Fix error in template instantiation - std::abs<floating_point> is illegal

riclarsson:
Fix for update in tool-chain.

### DIFF
--- a/environment-dev-win.yml
+++ b/environment-dev-win.yml
@@ -9,6 +9,7 @@ dependencies:
         - ipykernel
         - lark-parser
         - libcxx
+        - llvm-openmp < 21
         - matplotlib
         - nanobind < 2.5.0
         - nbsphinx


riclarsson:
@olemke Something has gone wrong on Windows over the weekend.  I have a red-tick for Windows even just pulling main.  We need to merge both fixes at once.  I cannot test Windows.

olemke:
I'll test it on the Windows machine here. The notebook failure seems similar to a problem we had in the past, maybe some Python package update broke it.

riclarsson:
Sounds really weird.  Good luck.  It's not like we can require clang<21 then either and merge this so I'll just leave it here.

olemke:
That was more painful than necessary. 🙁 Got sabotaged by my browser when copying the Python package lists from the logs for the version comparison. Turned out that it only copied the beginning and the end of the list, cutting out the middle part for some reason so that I didn't see the version change in llvm-openmp on Windows. 🤬 In the end it was all llvm's fault. The new `llvm-openmp` version led to silent crashes in `scipy.linalg.expm` (and possibly other places) on Windows.

PR #1004, opened 2025-09-04
Use full package name in stubgen

olemke:
Generate full typenames in stubs.
Resolves stubgen error with Python != 3.12. nanobind's stubgen throws error `AttributeError: module 'math' has no attribute 'floor'` when importing `numpy`. Likely an import naming confusion between the `math` module inside `arts` and the global Python `math` module.I don't know why the error does not occur with Python 3.12 but consistently with 3.10, 3.11 and 3.13.


riclarsson (approved):
If it works, it's OK.

Issue #1005, opened 2025-09-04
List continuum models in built-in documentation of absorption_speciesSet


stefanbuehler:
Help user figure out how to add the CKC4.0 water vapor continuum. Or, if there already is a complete list of options for absorption species somewhere else then refer to it here.

PR #1012, opened 2025-09-05
Upon request, give all extra information that is available about the …

riclarsson:
…species tags and give a link to let people know that they can easily download the data.

This should address #1005 since all available tags are printed and links to the data is added.


riclarsson:
The segfault here before was weird.  It would be caught better if we could run clang-asan automatically in the CI.

olemke:
> The segfault here before was weird. It would be caught better if we could run clang-asan automatically in the CI.

I was about to remove the `linux-nofortran-gcc` build from the CI because we don't depend on fortran anymore. I can replace that one with an `mac-asan-clang` build instead.

riclarsson:
Sounds good.  But we have few of those, I think it is better if we use a linux runner.

PR #1016, opened 2025-09-09
Improve ASAN option

olemke:
Added a cmake option for building with address sanitizer. The necessary compiler flags are now set in `CMakeLists.txt` instead of `CMakePresets.json`. The correct path for the asan shared library is determined automatically and preloaded when running the tests.

On linux, while it technically works, it's not very useful at the moment since most of the tests fail, often with failures not related to ARTS but other python modules (e.g. matplotlib). These do not occur on macOS.

@riclarsson: To speed things up a bit, I changed the base build preset for asan builds from debug to reldebug. Let me know if you prefer it to be a full debug build.


riclarsson:
This is solved with #1014 now.  The error was three ``Range{a + 1, N - a}`` which all should have read ``Range{a + 1, N - a - 1}``, for some start ``a < N`` and size ``N``.  I didn't need to change any results because of this (the removal of some checks to just allow nan-values is because I want to have a faded lower limit (i.e., no lower edge input)).

riclarsson:
So am I understanding correctly that the tests are now run even when ASAN is on?  If so, once #1014 is merged we should add a linux test to this PR that runs ASAN.

I also think the mechanism around the stubs should be more generic as some sort of ``DISABLE_PYARTS_STUBS`` parameter that the ASAN one turns on if it is off.  I think the largest part of my time compiling is spent on the ``__init__.pyi``.  It's over 3Mb large of structured ascii...

riclarsson:
One last question on the stubs:  since 2.7, nanobind should be able to handle their generation.  Why remove their generation rather than bump the nanobind version to 2.9 minimum?  Because we will start using newer nanobind stuff at some point.

PR #1042, opened 2025-10-21
Small fixes

riclarsson:
This contains many small fixes and additions.

- It adds a Zeeman geometry example.
- It adds a quick-look way to plot AbsorptionBands.  This allows simple "plot my absorption" style plots, with some level of automatic frequency selection.
- It fixes how ArrayOfSensorObsel is quick-look plotted (it used to copy a sparse matrix to a dense one, and one of my recent use-cases had that at 50 Gb, so sparse maths seems warranted).
- It adds a quick-look way to plot Zeeman geometry.
- It adds a simplified ``arts_omp_parallel(bool)`` method to keep open-mp pragmas shorter.
- It quits later in atmospheric reading routines so more errors can be gathered.
- It adds a simplified Zenith resolution sensor opening.
- It uses dynamic scheduling in ``measurement_vectorFromSensor``.  Some geometries are simply too much more expensive that it often costs more to use simple scheduling.  (Or perhaps it is because efficiency cores get scheduled.)
- Added a simple planetary data interface.  It can be used to get the ellipsoid and gravity operator.
- Moved my binary tangent algorithm to the geodesy bindings.
- Fix some documentation inconsistency in py_lbl.  Also added ``extract_species`` as a member to the ``AbsorptionBands`` and direct ``__getitem__`` to ``AbsorptionBand`` for easier access to the lines.
- Expose ``SparseStokvecMatrix`` and some helper reductions.  Also speed up collect mechanisms significantly when multiple sensors are collected into one.
- Add a few magnetic geometry stuff to Zeeman interface.

### DIFF
--- a/doc/arts/dev.workspace.groups.rst
+++ b/doc/arts/dev.workspace.groups.rst
@@ -40,7 +40,6 @@ of the map and the object is a struct with the following fields:
 
 - ``file`` - the main header file that must be included to use the workspace group.
 - ``desc`` - a description of the variable as a string.
-- ``array_depth`` - an integer defining the depth of the array.  You can recursively access data of this type of array using the ``[]`` operator using an integer or range type.  Remember that the ``value_type`` of the array should also be a workspace group.
 - ``value_type`` - a boolean that defines whether a workspace group instance of this type is copyable in python.  You cannot copy python types such as ``int`` and ``str``.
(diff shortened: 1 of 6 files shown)


olemke:
Interesting. It works fine when I build locally. But I can see them missing in the generated artifact. I'll check on our linux server.

PR #1044, opened 2025-10-23
Plots for gui

riclarsson:
This is a simple (AI generated) GUI proposal for ARTS 3.  It is mainly here to enforce a design.  The design is centered around:

1. Files included by ``plots/__init__.py`` should have a common set of interfaces.  In the init file, a plot function just calls one of the other functions, returning the common fig, ax pair.
2. Files in ``gui/edit/__init__.py`` should offer an edit interface.  The edit interface should be possible to call standalone but will often be called by selecting a variable in a list somewhere else by which it means it is not standalone but has a window parent.  The init file contains a generic interface that calls existing interfaces elsewhere, and will otherwise resort to eval options.
3. The core GUI feature is based on plotting a single variable.  It functions by passing a plot-routine to it that should accept a set of kwargs and a executed function that also (may) accept a kwarg.  This will then be passed as simulations_settings.  I have also added an "Additional Settings" where you can overwrite the context of the results kwarg.

So in short: If you want some plots done, add a plots/Typename.py file and include it in the init.  If you want to edit some value create a gui/edit/Typename.py file and include it in the init.  That's it.  You now have to call your own "generate data" method that is passed to the plotter.  For those not familiar, doing ``function_call(**a_dictionary)`` is equivalent to ``function_call(key1=a_dictionary[key1], key2=a_dictionary[key2], ...etc)``, which is how this all works.

This sounds perhaps complicated, so here is an example where it works directly on a builtin group:

```python
import pyarts3 as pyarts
from pyarts3.gui import start_gui


def plot_kwarg_func(*, los=None, pos=None, t: str = "2000-03-11 14:39:37", N: int = 50, **_):
    if los is None:
        los = pyarts.arts.Vector2([45.0, 180.0])
    elif not isinstance(los, pyarts.arts.Vector2):
        los = pyarts.arts.Vector2(los)

    if pos is None:
        pos = pyarts.arts.Vector3([50e3, 0.0, 0.0])
    elif not isinstance(pos, pyarts.arts.Vector3):
        pos = pyarts.arts.Vector3(pos)

    ell = pyarts.arts.planets.Earth.ellipsoid
    mag = pyarts.arts.igrf(pos, ell, t=t)
    ang = pyarts.arts.zeeman.MagneticAngles(mag, los)

    # Return kwargs for plot; plot takes 'ang' as first param but also accepts keyword
    return {
        'ang': ang,
        'N': N,
    }


# Initial simulation settings (editable via GUI)
simulation_settings = {
    'los': pyarts.arts.Vector2([45.0, 180.0]),  # zenith [deg], azimuth [deg]
    'pos': pyarts.arts.Vector3([50e3, 0.0, 0.0]),  # ECEF position [m]
    't': "2000-03-11 14:39:37",  # IGRF time string
    'N': 50,  # Plot density
}

# Start the GUI using the MagneticAngles plotter
start_gui(plot_kwarg_func, pyarts.plots.MagneticAngles.plot,
          simulation_settings=simulation_settings)
```

And here is an example for manipulation of the workspace for disort:

```python
import os

import matplotlib.pyplot as plt
import numpy as np
import pyarts3 as pyarts
from pyarts3 import plots
from pyarts3 import gui

# Download catalogs
pyarts.data.download()

toa = 120e3
lat = 0
lon = 0
NQuad = 40

ws = pyarts.Workspace()

# %% Sampled frequency range
line_f0 = 118750348044.712
ws.frequency_grid = np.linspace(-20e9, 2e6, 101) + line_f0

# %% Species and line absorption
ws.absorption_speciesSet(species=["O2-66"])
ws.ReadCatalogData()

# %% Use the automatic agenda setter for propagation matrix calculations
ws.propagation_matrix_agendaAuto()

# %% Grids and planet
ws.surface_fieldPlanet(option="Earth")
ws.surface_field[pyarts.arts.SurfaceKey("t")] = 295.0
ws.atmospheric_fieldRead(
    toa=toa, basename="planets/Earth/afgl/tropical/", missing_is_zero=1
)
ws.atmospheric_fieldIGRF(time="2000-03-11 14:39:37")

# %% Checks and settings
ws.spectral_radiance_transform_operatorSet(option="Tb")
ws.spectral_radiance_space_agendaSet(option="UniformCosmicBackground")
ws.spectral_radiance_surface_agendaSet(option="Blackbody")

# %% Core Disort calculations
ws.disort_settings_agendaSetup()

ws.disort_quadrature_dimension = NQuad
ws.disort_legendre_polynomial_dimension = 1
ws.disort_fourier_mode_dimension = 1


def plot_kwargs(ws):
    """Generate plot kwargs from workspace."""
    ws.disort_spectral_radiance_fieldProfile(
        longitude=lon,
        latitude=lat
    )
    return {
        "disort_radiance": ws.disort_spectral_radiance_field,
    }


gui.start_gui(
    plot_func=plots.DisortRadiance.plot,
    plot_kwarg_func=plot_kwargs,
    simulation_settings={"ws": ws},
)
```

To be clear, most of the stuff in this PR is AI generated and untested.  I am only letting the AI fix the GUI by modifying the design.

### DIFF
--- a/tests/python/plots/test_plot_signatures.py
+++ b/tests/python/plots/test_plot_signatures.py
@@ -12,7 +12,6 @@ import importlib.util
 
 
 def get_plot_modules():
-    """Get all plot module files from pyarts3.plots."""
     plots_dir = Path(__file__).parent.parent.parent / "src" / "pyarts3" / "plots"
     
     # Get all .py files except __init__.py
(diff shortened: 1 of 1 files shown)


riclarsson:
I am merging this as WIP

PR #1047, opened 2025-10-29
[ARTS-2.6] Relax ELL07 error condition

olemke:


erikssonpatrick:
Yes. There should not be any liquidcloud below about 235K. But to avoid getting errors too quickly, the old limit was 210K. Now when instead returning zeros, 230K makes more sense.

PR #1056, opened 2025-11-14
Shorter names

riclarsson:
This is a draft for reducing the size of names in ARTS3.  The original set of names made this grow out of control

### DIFF
--- a/tests/core/jupiter/read-atm.py
+++ b/tests/core/jupiter/read-atm.py
@@ -2,6 +2,6 @@ import pyarts3 as pyarts
 
 ws = pyarts.Workspace()
 
-ws.atmospheric_fieldRead(toa=500e3,
-                         basename="planets/Jupiter/MPS/",
-                         deal_with_field_component="Ignore")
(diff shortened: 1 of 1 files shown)


erikssonpatrick:
In general thumbs up. But maybe a few details:
- Just x for "cross-section" is not very clear. I suggest to go for xsec (Your table indicates that we here already use this short name!)
- I am using za for zenith angle. And then it will be aa for azimuth angle. OK inside code, but not very clear for workspace variables. What about zen and azi? That would better follow the pattern of lat and lon.
- By the way, what is the short name for pressure? pre?

erikssonpatrick:
Another point. You wrote here specular_propmat. A typo? You mean spectral_propmat I assume.

Is there a consistency on this point, for similar variables. That if not having spectral_ in front, they contain data for multiple frequencies?
Adding spectral_ is not very compact. What about a scheme like propmat and propmats, where the lack of the s means a single element, while if the s is there the variable contains multiple values?

PR #1063, opened 2025-11-25
Add linux arm build to CI

olemke:


riclarsson:
@olemke Excellent!  It builds quite fast.  I guess few people are using these cores :)

The std::abs thing...   what actually happens?  is this related to arm or a separate bug?

olemke:
@riclarsson The abs thing is an arm thing. Without `std::`, the compiler uses the int instead of float version of abs. Which surprised me since we're using the same GCC compiler version on x86 and don't get this issue:

```
/home/runner/work/arts/arts/src/core/scattering/phase_matrix.h:96:13: error: no matching function for call to 'equal(int, double)'
   96 |       (equal(abs(aa_scat - aa_inc), 2.0 * pi_v<Scalar>))) {
      |       ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
https://github.com/olemke/arts/actions/runs/19571385116/job/56045515247#step:7:457 (not sure if github let's you view this)
Similar here: https://github.com/atmtools/arts/actions/runs/19664055131/job/56316610489#step:7:70

There is another bunch of warnings, would be interesting to hear your opinion on what we should do about those:
```
/home/runner/work/arts/arts/src/core/lbl/lbl_data.h:229:69: warning: the result of the conversion is unspecified because '-1' is outside the range of type 'LineShapeModelVariable' [-Wconversion]
  |   LineShapeModelVariable ls_var{static_cast<LineShapeModelVariable>(-1)};
  |                                                                     ^~
```
https://github.com/atmtools/arts/actions/runs/19694341913/job/56416288579#step:8:313

And also for this one, why doesn't GCC show the same warning on x86? 🤷‍♂️

PR #1094, opened 2026-02-10
Update for HITRAN2024 new quanta and for TIPS2024

riclarsson:
This update uses TIPS 2024 for partition functions, adds some species that are new in HITRAN2024, and changes some print and store options so that maps are printed in a sorted manner - or should be.


riclarsson:
Yeah, arts-cat-data would be a good place to place it.  It is required for compilation though, so there's a few extra steps that need to happen.

PR #1099, opened 2026-02-18
Change upstream location of testdata to UHH Gitlab

olemke:
If a download fails, output the actual RequestException error message.
Reduce outer retry attempts from 30 to 2 (each of those already retries up to 5 times inside the `download_file` function). Past experience has shown if the download fails once, it'll keep failing no matter how many retries are done.


riclarsson:
Hehe, I just saw the  ❌

olemke:
Superseded by #1100

PR #1110, opened 2026-03-19
Port of single scattering radar of ARTS2

riclarsson:
This is a port of single scattering radar from ARTS2 to ARTS3.

It is entirely AI-guided.  So I cannot merge this without review from @erikssonpatrick and @m-brath .

There are two parts to this PR:

1. The port of the scattering data
2. The addition of a RayleighScattering ScatteringSpecies.  I designed the latter to work like species tag for absorption models so that both the water droplet model that the AI ported from ARTS2 and the air-model that was used in the ARTS2/3 solar scattering calculations can be used.

Anyways, this needs a proper review because I am not sure how to test the radar other than trying to recreate the test in ARTS2, and I am not sure that that has been done correctly.

### DIFF
--- a/examples/2-clearsky-radiative-transfer/2-zeeman/6-zeeman-sun-scattering.py
+++ b/examples/2-clearsky-radiative-transfer/2-zeeman/6-zeeman-sun-scattering.py
@@ -41,6 +41,9 @@ ws.spectral_rad_surface_agendaSet(option="Blackbody")
 ws.ray_path_observer_agendaSetGeometric()
 ws.spectral_propmat_scat_agendaSet(option="AirSimple")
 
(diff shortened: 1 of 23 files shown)


riclarsson:
@m-brath I need your help reviewing the scattering implementation to ensure it is in line with your understanding of how the new system should work.

@erikssonpatrick I need your help to understand whether or not the radar tests and implementation comply with your understanding from ARTS2.

Generally, I am also a bit unhappy with how many WSVs were added, but I do not have the experience required with radar stuff to make a class of this I am happy with (i.e., a RadarSensor class).

riclarsson:
No interest.  So I will close this.

PR #1111, opened 2026-03-19
Use Eigen 5.0

olemke:


olemke:
@riclarsson I just saw that you also made a change in #1110 to fix the Eigen dependency. What do you reckon, should we just depend on whatever version is installed or keep the dependency fixed to a version we have "tested"?

riclarsson:
I don't like this approach.  We have no reason to fix the Eigen version and will just lose out on potentially new functionality.  They didn't break anything, they just renamed their versioning.

When something breaks, we fix the version of that thing to the previous version.  The tests either covers it, or we don't care about the code breaking.

Of course, we don't test performance yet (i.e., we don't care about it).  So it might be we actually want to lock version to 3.4, but I don't think this update is in that spirit.

olemke:
Ok, fine with me. I'll update this PR accordingly.

riclarsson:
Excellent.  They also explain their versioning [here](https://gitlab.com/libeigen/eigen/-/releases/5.0.0).  As you can see, it is still Eigen3 (like pyarts is pyarts3), and they claim that this is a breaking change because of the changes below (which we do not use).  But I do think that 5.1.0 will work from their explanation.

olemke:
I just cherry-picked your change. Then we can merge this PR already and don't have to wait for the radar-port PR which might take longer until merge.

PR #1112, opened 2026-04-03
Refactoring of species and isotopologues

riclarsson:


riclarsson:
@olemke I would like it if you check this stuff.  I have made many changes to the species layout.  We are now downloading a lot more data.  But with this, it is possible to extend ARTS species without modifying C++ files.

I kept the split-format in arts-cat-data so that adding individual species is relatively straightforward.  I think.

riclarsson:
@olemke, I will merge this because I need it.  The short of this is that for any release of ARTS, you need to have a modern copy of arts-cat-data ready.  Data from there is used for compilation now.

olemke:
@riclarsson I'm back from Easter vacation and looking at the code now. I'm happy with the overall implementation but I think handling access to the data in the build system needs some improvements. It's currently not obvious that when compiling locally, one has to set ARTS_CAT_DATA_PATH explicitly. The warning about no partition functions being available during the configuration should be an error and clearly state that ARTS_CAT_DATA_PATH needs to be set.

We could even consider downloading the whole partition-functions and isotopologues folders during the build process. There not that large anyway. I see some risk that people accidentally end up with a half-broken arts build because only the partition functions present in their testdata directory were compiled in.

PR #1116, opened 2026-04-15
Fix arts_omp_parallel condition

olemke:
When n = -1 (passed as unsigned long long) to `arts_omp_parallel`, it becomes ULLONG_MAX. The comparison `arts_omp_get_max_threads() < static_cast<int>(n)` evaluates `static_cast<int>(ULLONG_MAX)`, which truncates to -1 on most platforms. Since max_threads < -1 is always false, parallelisation is always disabled if n = -1 is passed and lnot the indented behavior.

[Example on godbolt](https://godbolt.org/z/9Peo4Ydo3)

### DIFF
--- a/src/core/util/arts_omp.cc
+++ b/src/core/util/arts_omp.cc
@@ -14,6 +14,7 @@
 */
 
 #include "arts_omp.h"
+#include <limits>
 
 //! Wrapper for omp_get_max_threads.
 /*! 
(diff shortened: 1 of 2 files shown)


olemke:
@riclarsson Since you had an OMP guard in the `low_memory` version, I also added guards to the `high_performance` version. Let me know if you intentionally didn't put them there for some reason.

riclarsson:
This is fine!  Thanks for the catch.

PR #1117, opened 2026-04-16
Make decorator that executes an agenda in parallel

riclarsson:
Now you can do

```python3
@pyarts.in_parallel(ws=ws)
def create_agenda_parallel(ws: pyarts.Workspace):
  ws.abs_speciesSet(species=["H2O", "CO2"])
  ws.atm_pointInit()
```

The two methods are executed in parallel.  Note that this will come with overhead, and the example above is clearly faster to run serial.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,4 @@ heaptrack.*.gz
 qdrant_storage/
 .roo/
 tests/core/sensor/sensor_count.xml
+tests/core/xml/test.xml
--- a/python/src/pyarts3/__init__.py.in
+++ b/python/src/pyarts3/__init__.py.in
@@ -2,7 +2,7 @@
 """This module contains functions to interact with ARTS."""
 
 from pyarts3 import arts  # noqa
-from pyarts3.workspace import Workspace, arts_agenda  # noqa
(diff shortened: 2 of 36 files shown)


riclarsson:
@olemke, there is a bug in the code here.  I am not sure how to solve it, so I have disabled "Any" outside of agendas.  I did not know how to find the actual cause here.

The issue: casting is nanobind to a py::object that is invalid for "Any" might cause segfault.  To reproduce under current ARTS3: ``ws.ReadXML(ws.abs_cia_data, filename="cia/O2-CIA-O2.xml")``.  This will sometimes cause segfaults but cause bad-cast all the time.  I thought I had tested this enough.  The error is entirerly in nanobind looking code by the debug info, and I cannot figure it out...

olemke:
So `py::type_name(*_output)` is causing the segfault when outputting the runtime error message in ReadXML. Taking your ReadXML example and passing it an xml file with the correct type (CIARecords instead of CIARecord), the data is read correctly into the WSV. Segfault only occurs when the runtime error is triggered. Does `pyAny` cause problems in other places? If it's just an issue in the runtime error message, we could maybe just work around it (like not printing the type of _output)? Maybe also something we can discuss in the meeting tomorrow.

PR #1122, opened 2026-04-24
Rework testdata download system to support folders and download at configure time

olemke:
This PR improves the ARTS testdata acquisition system by:
1. Adding support for downloading entire folders (not just individual files)
2. Moving testdata download from build time to CMake configure time

### Changes

| Component | Change |
|-----------|--------|
| `cmake/modules/ArtsTestdata.cmake` | **New file** - CMake module with `setup_arts_xml_data()` and `setup_arts_cat_data()` macros |
| `CMakeLists.txt` | Refactored to use the new macros; downloads now happen at configure time |
| `testdata/get_testdata.py` | Major rewrite to support folder downloads via ZIP archives and SVN directory listings |
| `testdata/cat_data_files.txt` | Download all species info via folder paths (`partition_functions/`, `isotopologues/`, etc.) instead of individual files |
| `testdata/CMakeLists.txt` | `get-testdata`, `refresh-testdata` |

### Key Features

**Folder Download Support**: The `get_testdata.py` script now:
- Downloads folders as ZIP archives from GitLab (primary source)
- SVN fallback uses directory listing with recursive file discovery

**Configure-Time Downloads**:
- Testdata is now downloaded during the CMake configuration phase instead of at build time, ensuring data is available before compilation starts.
- The `refresh-testdata` target can be used to force a redownload at build time (e.g., get new upstream catalog changes without pruning the build directory)
- `get-testdata` is used as a safety net dependency at build time to ensure that the testdata is in order (should probably be removed later), only case where this is useful is to download files that have been added to the catalogs upstream but are missing locally. But with major drawbacks: Upstream file modifications are not detected. Contents of folder downloads is not checked for new files.


riclarsson (approved):
Looks good

PR #1124, opened 2026-04-29
Fix numerical instability in ECEF to geodetic coordinate conversion

olemke:
This PR fixes a numerical instability in the ECEF to geodetic coordinate conversion algorithm. The previous implementation used an exact equality check (`ecef[0] == 0 && ecef[1] == 0`) to detect points on the poles, which was both numerically unstable and failed to handle near-pole cases. These near-pole cases could cause the iterative algorithm to enter an endless loop. To safeguard against any other edge cases that might still be overlooked, additional convergence protection was added.

### Changes
- `src/core/geodesy/geodetic.cpp`: 
  - Added `near_pole_threshold_ecef` constant to detect near-pole points using a threshold-based approach instead of exact equality
  - Added `max_ecef2geodetic_iter` constant to prevent endless loops in the iterative conversion algorithm
  - Reorganized the coordinate conversion logic to first check the horizontal distance from the z-axis and handle near-pole cases using the simplified formula
  - Added user error reporting when the iterative algorithm fails to converge within the maximum number of iterations

- `src/tests/CMakeLists.txt`: Reenabled the `test_path_point` test that was previously disabled (commit c94deff0a part of #773)

### DIFF
--- a/src/core/geodesy/geodetic.cpp
+++ b/src/core/geodesy/geodetic.cpp
@@ -25,6 +25,8 @@
 #include <lin_alg.h>
 #include <math_funcs.h>
 
+#include <cmath>
+
 using Math::pow2;
 
 inline constexpr Numeric DEG2RAD = Conversion::deg2rad(1);
@@ -135,7 +137,7 @@ Vector3 ecef2geodetic(Vector3 ecef, Vector2 refellipsoid) {
   } else {
     // The general algorithm not stable for lat=+-90. Catch these cases
     // Also catch near-pole cases where numerical instability can occur
-    const Numeric sq = sqrt(ecef[0] * ecef[0] + ecef[1] * ecef[1]);
+    const Numeric sq = std::hypot(ecef[0], ecef[1]);
 
     if (sq < near_pole_threshold_ecef * refellipsoid[0]) {
       // Near-pole case: use simplified formula (same as exact pole)
--- a/src/core/geodesy/geodetic.cpp
+++ b/src/core/geodesy/geodetic.cpp
@@ -45,6 +45,21 @@ inline constexpr Numeric ellipsoid_radii_threshold = 1e-3;
 inline constexpr Numeric POLELATZZZ =
     90 - 1e-8;  // Rename to POLELAT when other one removed
 
+/** Threshold for near-pole detection in ECEF to geodetic conversion
+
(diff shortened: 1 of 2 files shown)


src/core/geodesy/geodetic.cpp
```diff
@@ -117,29 +132,39 @@ Vector3 ecef2geodetic(Vector3 ecef, Vector2 refellipsoid) {
...
     pos[0] -= refellipsoid[0];
 
+  } else {
     // The general algorithm not stable for lat=+-90. Catch these cases
-  } else if (ecef[0] == 0 && ecef[1] == 0) {
-    pos[0] = fabs(ecef[2]) - refellipsoid[1];
-    pos[1] = ecef[2] >= 0 ? 90 : -90;
-    pos[2] = 0;
+    // Also catch near-pole cases where numerical instability can occur
+    const Numeric sq = sqrt(ecef[0] * ecef[0] + ecef[1] * ecef[1]);
```
riclarsson: You should almost always use hypot here.  It is numerically stable when one of ecef[0, 1] is zero.

src/tests/CMakeLists.txt
```diff
@@ -178,8 +178,8 @@ target_link_libraries(test_rtepack PUBLIC artstime rtepack rng)
 # ####
 add_executable(test_path_point test_path_point.cc)
 target_link_libraries(test_path_point PUBLIC artstime path rng)
-#add_test(NAME "cpp.fast.test_path_point" COMMAND test_path_point)
-#add_dependencies(check-deps test_path_point)
+add_test(NAME "cpp.fast.test_path_point" COMMAND test_path_point)
+add_dependencies(check-deps test_path_point)
```
riclarsson, line 182: You add this again.  Was this the reason for the previous failure, then?

src/core/geodesy/geodetic.cpp
olemke: Good advice. I wasn't aware of that function. We should probably use it in other places too.

src/tests/CMakeLists.txt
olemke: It was consistently failing on Windows (not sure if it errored out or got stuck in the endless loop). Since the fix, I haven't seen any failures reoccurring which leads me to believe this was the culprit.

PR #1128, opened 2026-05-13
Change CMB by 1 cK

riclarsson:


riclarsson:
The reported value of CMB is 2.725 K but value from ARTS early days was at 2.735 K.

NOTE: This is in ARTS 2.6 but I will not change it because too many reference file-values change.

olemke:
Noted for further discussion in the next meeting.

PR #1141, opened 2026-07-09
Add DISORT coupling

riclarsson:
This adds a mode for DISORT that couples a subsurface model with an atmosphere model via using eachother as boundary conditions.  It is experimenta but seems to mostly work.

### DIFF
--- a/src/m_disort_settings.cc
+++ b/src/m_disort_settings.cc
@@ -514,7 +514,7 @@ void disort_settingsOpticalThicknessFromPath(
   }();
 
   ARTS_USER_ERROR_IF(stdr::any_of(r, Cmp::le<0.0>()),
-                     R"(Atmospheric layer thickness must be upward.
+                     R"(Atmospheric layer thickness must be positive.
 
 Values:   {:B,}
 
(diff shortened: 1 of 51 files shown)


src/m_disort_settings.cc
```diff
@@ -483,7 +514,7 @@ void disort_settingsOpticalThicknessFromPath(
   }();
 
   ARTS_USER_ERROR_IF(stdr::any_of(r, Cmp::le<0.0>()),
-                     R"(Atmospheric layer thickness must be positive.
+                     R"(Atmospheric layer thickness must be upward.
```
olemke: Accidental replacement of positive -> upward?

src/m_disort_settings.cc
```diff
@@ -509,7 +540,7 @@ ray_path: {:B,}
                            R"(
 Not strictly increasing optical thicknesses between layers.
 
-Check *spectral_propmat_path* contain zeroes or negative values for A().
+Check *spectral_propmat_path* contain zeroes or downward values for A().
```
olemke: Accidental replacement of negative -> downward?

src/m_disort_settings.cc
```diff
@@ -546,7 +577,7 @@ void disort_settingsSubsurfaceScalarAbsorption(
   }();
 
   ARTS_USER_ERROR_IF(stdr::any_of(r, Cmp::le<0.0>()),
-                     R"(Atmospheric layer thickness must be positive.
+                     R"(Atmospheric layer thickness must be upward.
```
olemke: Accidental replacement of positive -> upward?

PR #1142, opened 2026-07-09
fix: restore str->enum implicit conversion in supergeneric args under nanobind 2.13

olemke:
This PR fixes a regression introduced by nanobind 2.13 that silently broke implicit Python-side conversions (such as `str` -> `SpeciesEnum`) for supergeneric workspace-method arguments leading to several tests failing, such as:

```
➜ ctest --test-dir build -R pyarts.arts.tests.core.vmr.log$
Test project /Users/u237002/Hacking/uhh/arts/build
    Start 95: pyarts.arts.tests.core.vmr.log
1/1 Test #95: pyarts.arts.tests.core.vmr.log ...***Failed    0.92 sec
Traceback (most recent call last):
  File "/Users/u237002/Hacking/uhh/arts/tests/core/vmr/log.py", line 64, in <module>
    ws.jac_targetsToggleLogarithmicAtmTarget(key="H2O")
TypeError: jac_targetsToggleLogarithmicAtmTarget(): incompatible function arguments. The following argument types are supported:
    1. jac_targetsToggleLogarithmicAtmTarget(self, jac_targets: pyarts3.arts.JacobianTargets | None = None, atm_field: pyarts3.arts.AtmField | None = None, key: pyarts3.arts.AtmKey | pyarts3.arts.SpeciesEnum | pyarts3.arts.SpeciesIsotope | pyarts3.arts.QuantumLevelIdentifier | pyarts3.arts.ScatteringSpeciesProperty | None = None) -> None

Invoked with types: pyarts3.workspace.workspace.Workspace, kwargs = { key: str }
```

It also relaxes the `nanobind` pin in the platform dev-environment files to `<=2.13` so the fix can be consumed.

### The underlying problem

ARTS exposes supergeneric workspace-method arguments to Python as `const std::variant<std::shared_ptr<pyT1>, std::shared_ptr<pyT2>, ...> * const`. The bindings register implicit conversions like `py::implicitly_convertible<std::string, SpeciesEnum>`, and the stock nanobind `std::variant` caster honors these by probing each alternative twice: first with the `convert` flag clear (exact matches only), then with `convert` set (allowing implicit conversions).

nanobind 2.13 added a single line at the top of `type_caster<std::shared_ptr<T>>::from_python` (in `nanobind/stl/shared_ptr.h`):

```cpp
flags &= ~((uint8_t) cast_flags::convert);
```

The intent of that change is to prevent dangling references to implicitly-converted temporaries owned by the cleanup list. But as a side effect, when the variant caster probes a `std::shared_ptr<T>` alternative during the second (`convert`-allowed) pass, the shared_ptr caster immediately clears `convert` before delegating to the inner `T` caster. The inner `T` caster (e.g. `SpeciesEnum`) therefore runs with `convert` unset and rejects the incoming `str` even though `implicitly_convertible<std::string, SpeciesEnum>` is registered. Since every alternative is a `std::shared_ptr<...>`, *all* alternatives fail, and `key="H2O"` raises a `TypeError` instead of constructing a `SpeciesEnum`.

This is purely a consequence of the `shared_ptr` wrapper interacting with the variant's two-pass strategy; the inner enum caster itself still honors implicit conversions when `convert` reaches it, which is why a non-supergeneric argument of type `SpeciesEnum` continues to work.

### The fix

`src/python_interface/python_interface.h`:

Add `Python::Supergeneric<...>`, a struct that is layout-compatible with `std::variant<std::shared_ptr<T>...>` (it exposes the variant via `operator const std::variant<...>&`) so the existing `select_gin` / `select_gout` / `has_selected_value` / `type` overloads keep working unchanged — forwarding overloads taking `Supergeneric` are added alongside them.

Add a custom nanobind `type_caster<Python::Supergeneric<...>>` that takes ownership of the conversion. It mirrors the stock variant caster's two-pass strategy (try without `convert`, then with `convert`), but instead of delegating each alternative probe to `type_caster<std::shared_ptr<T>>` — which is where the 2.13 regression strips the flag — it invokes the inner `T` caster directly (`make_caster<T>`) with the `convert` flag intact. On success the resulting `T*` (kept alive by the cleanup list for the duration of the call) is wrapped in a `std::shared_ptr<T>` whose deleter increfs/decrefs the Python source object, reproducing the lifetime semantics of nanobind's stock shared_ptr caster. `from_cpp` is delegated to the regular variant caster via the stored `value` member.

`src/python_interface/gen_auto_py_methods.cpp`:

Emit generated supergeneric arguments as `Python::Supergeneric<pyT...>` instead of `std::variant<std::shared_ptr<pyT>...>`, and drop the now-redundant `shared_ptr` wrapping in the alternative list, so the custom caster is selected for these parameters.

`environment-dev-*.yml`:

Relaxed the `nanobind=2.9` pin to `nanobind<=2.13` so the fixed build can use the newer nanobind on all platforms.

### Breaking Changes
None. `Supergeneric` is layout-compatible with the previous `std::variant<std::shared_ptr<T>...>` and the change is internal to the Python bindings; generated method signatures and runtime behavior match the pre-2.13 semantics.

### DIFF
--- a/environment-dev-linux-clang.yml
+++ b/environment-dev-linux-clang.yml
@@ -21,7 +21,7 @@ dependencies:
         - libcxx
         - libcxx-devel
         - llvm-openmp
-        - nanobind=2.9
+        - nanobind<=2.13
         - nbsphinx
         - matplotlib<3.12
         - netcdf4
--- a/environment-dev-linux.yml
+++ b/environment-dev-linux.yml
@@ -16,7 +16,7 @@ dependencies:
         - libboost-devel
         - make
         - matplotlib<3.12
-        - nanobind=2.9
+        - nanobind<=2.13
         - nbsphinx
         - netcdf4
         - ninja
--- a/environment-dev-mac.yml
+++ b/environment-dev-mac.yml
@@ -20,7 +20,7 @@ dependencies:
         - libmicrohttpd
         - llvm-openmp
         - matplotlib<3.12
-        - nanobind=2.9
+        - nanobind<=2.13
         - nbsphinx
         - netcdf4
         - ninja
--- a/environment-dev-win.yml
+++ b/environment-dev-win.yml
@@ -11,7 +11,7 @@ dependencies:
         - libcxx
         - llvm-openmp < 21
         - matplotlib<3.12
-        - nanobind=2.9
+        - nanobind<=2.13
         - nbsphinx
         - netcdf4
         - ninja
--- a/src/python_interface/gen_auto_py_methods.cpp
+++ b/src/python_interface/gen_auto_py_methods.cpp
@@ -65,9 +65,9 @@ std::string method_arguments(const WorkspaceMethodInternalRecord& wsm) {
     } else {
       const auto& v          = wsm.gin.at(i);
       std::string_view comma = "";
-      os << ",\n    std::variant<";
+      os << ",\n    Python::Supergeneric<";
       for (auto& arg : generics[i]) {
-        os << std::exchange(comma, ", ") << "std::shared_ptr<py" << arg << ">";
+        os << std::exchange(comma, ", ") << "py" << arg;
       }
       os << "> * const _" << v;
     }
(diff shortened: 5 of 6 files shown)


riclarsson:
It seems to me that this should really be fixed upstream.  It is very hacky to use other people's detail namespace.  Have you made a post to the nanobind issues-list?

Because I can easily imagine this is another time when the custom-exposure struct/tag they use is ignored by mistake.  What does that flag variable actually do?

It is fine as a workaround for now in our project but we should check with them.  So I will merge it.

olemke:
I agree that it's not a desirable solution. But they intentionally made this change to fix a dangling shared pointer. See wjakob's description under ["Dangling shared_ptr from implicit conversion"](https://togithub.com/wjakob/nanobind/pull/1371). Also see the [commit details](https://togithub.com/wjakob/nanobind/commit/22730697dea69d2b87672efe85661a35ccfc30e5).

riclarsson:
I'll ask on the commit details page if he has any thoughts on this.

PR #1146, opened 2026-07-15
Gcc16

riclarsson:
This commit makes the formatting system in ARTS work with GCC16.

There is a change I cannot identify between GCC15 and GCC16 where raw characters, and their arrays are not treated correctly.  Raw `char` becomes `int`, so `' ' `prints as `32`.  Raw `char[]` arrays (probably also `const char *` but I never tested), also has their problems in that they keep the `\0` in the generated string (so it secretly prints to files, which is only caught when programmatically parsing the file again.)

Anyways, the updated code explicitly makes these illegal.  Your code will not compile when you try to put raw characters into the formatter.  The error for `tags.format(ctx, v.numer, "/", v.denom);` reads (at the bottom):

`• the expression '(... && no_raw_chars<Rest>) [with Rest = {char[2], long long int}]' evaluated to 'false'`

which might sound cryptic but the exact line is marked in your custom formatter.  So hopefully seeing "no_raw_chars" and the exact function call will be enough to understand the bug.  Sadly, all the formatter code is templated so the error message will appear late in the compilation process.


CMakeLists.txt
```diff
@@ -109,11 +109,11 @@ else()
   message(STATUS "Compiling ARTS without time profiling information (-DENABLE_ARTS_PROFILING=ON to enable)")
 endif()
 
-########### C++20/23 Support ##########
+########### C++23/26 Support ##########
 
 if (NOT CMAKE_CXX_STANDARD)
-  if (ENABLE_CXX23)
-    set(CMAKE_CXX_STANDARD 23)
+  if (ENABLE_CXX26)
```
olemke, line 115: Also needs rename in options() list.

src/core/jacobian/covariance_matrix.h
```diff
@@ -405,11 +405,11 @@ struct std::formatter<BlockMatrix> {
 
   template <class FmtContext>
   FmtContext::iterator format(const BlockMatrix &v, FmtContext &ctx) const {
+    tags.add_if_bracket(ctx, "["sv);
```
olemke: This looks like a bug. The opening [ is now added before the early return for non-null matrices.

PR #1148, opened 2026-07-15
fix: relax DISORT test tolerance under LGPL build

olemke:
The LGPL-compatible ARTS build uses a different eigensolver implementation than the default build, which can produce slightly different floating-point results. The existing test tolerance of `1e-6` was too strict for this configuration and caused the `disort-test-4.cpp` to fail. This PR widens the accepted relative-error threshold for the LGPL build to `2e-6` so the test passes without weakening the default build's stricter check.

### DIFF
--- a/src/core/disort-cpp/disort-test.h
+++ b/src/core/disort-cpp/disort-test.h
@@ -25,7 +25,11 @@ inline bool is_good(const auto& a, const auto& b) {
         if (nonstd::isnan(first) or nonstd::isnan(second)) return false;
         if (first == 0.0 and second == 0.0) return true;
         const Numeric ratio = std::abs(first / second - 1);
+#if ARTS_LGPL
+        return ratio < 2e-6;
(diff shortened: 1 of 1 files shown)


riclarsson:
Would it not be better to simply set OpenMP counts to 1?  Because as far as I can see, that the Lapack is multi-threaded should be the only real difference.

olemke:
Have a look at my [change-disort-test4-single-threaded branch](https://github.com/olemke/arts/commits/change-disort-test4-single-threaded/)
It reverts the previous change and adds an [`arts_omp_num_threads(1)` to `main()` for `disort-cpp-test-4`](https://github.com/olemke/arts/commit/02ec2f54b633d6a67ba3c875fad9ff30e611585f) instead.
And still, the [lgpl test failed](https://github.com/olemke/arts/actions/runs/29413346806/job/87345561206)

PR #1150, opened 2026-07-16
Wider format from 80 -> 120 chars

riclarsson:
I think we change to allow more horizontal space to be used in clang-format.  I personally would prefer 120 or even 132 char.  Any thoughts?  I would go through all the code and apply the formatting before merging this.


olemke:
In light of the modern C++ constructs (long template names, qualified types like `std::vector<std::shared_ptr<const Foo>>`), I think increasing the line length is reasonable. But I wouldn't go beyond 120 chars. The longer a line gets, the harder it is to visually navigate from the end of one line to the beginning of the next. I would also like to still be able to see the whole line (with a readable font size :-) ) on a smaller display (e.g. laptop) without being forced to invoke horizontally scrolling. I think 120 chars is a workable solution to have the right balance.

riclarsson:
OK, I'll leave it at 120.

Are you fine with the other clang-format changes I made?  It is mostly about alignments and allowing C++ keywords to not create excessive new lines.  I think they all make code navigation a bit easier.

olemke:
Yes, looks good overall. I'm not a big fan of the function argument alignments though. While it looks clean, it potentially creates messier diffs since changing one argument can require a reformat of the whole argument list. This makes the actual change harder to identify and increases the potential for merge conflicts.

riclarsson:
I prefer the code looking good to the change-log being clean here.  Mostly because I spend a lot more time coding than reviewing.  The arguments aligning is very useful as it makes it much easier to read the names of the incoming parameters.

olemke:
It's fine. I'm not vetoing it.

PR #1151, opened 2026-07-16
Move builtin species to arts-cat

riclarsson:


olemke:
You might have to restart the CI on this because I just updated the git mirror of arts-cat-data with your recent changes. The `isotoplogues/` directory was still missing the new files when `Prepare testdata` was run.

PR #1153, opened 2026-07-22
More outputs

riclarsson:


src/workspace_agenda_class.cpp
```diff
@@ -68,6 +64,60 @@ void Agenda::finalize(bool fix) try {
...
+  auto& [required, can_be_fixed] = out;
+
+  for (const std::string& i : must_in) {
+    if (stdr::binary_search(outs_first, i)) required.push_back(i);
+    if (not stdr::binary_search(ins_first, i)) can_be_fixed.push_back(i);
+  }
+
+  return out;
+}
+
```
olemke, line 105: - Verbatim duplicate of `missing_inputs`?
- Indentation of function parameters don't satisfy clang format

src/workspace_agenda_class.cpp
```diff
@@ -68,6 +64,60 @@ void Agenda::finalize(bool fix) try {
...
+  auto [ins_first, outs_first, in_then_out] = agendas_ins_and_outs(methods);
+
+  const auto [required_input, fixable_input] = missing_inputs(must_in, ins_first, outs_first);
+
+  if (not required_input.empty()) {
+    throw std::runtime_error(std::format(
+        R"(Agenda "{}" first requires {:B,} as input but they are output of the agenda)", name, required_input));
+  }
+
```
olemke, line 126: - `fixable_input` is computed but never used.
- Lines 121-124 (the `outs_first` check inside the `must_in` loop) are now **unreachable**: `required_input` contains exactly those elements and throws at line 116 first.
- The loop's remaining branch re-derives exactly what `fixable_input` already holds.

PR #1162, opened 2026-09-01
Transmission matrix

riclarsson:
This adds some extra ways to compute the transmission matrix.  It also seem to fix a bug in my expm code.  This error might affect ARTS2.


riclarsson:
@stefanbuehler There is a typo in the ARTS 2.6 paper by me.  I messed up some square roots.  Eq. (7) x^2 and y^2 should not have square roots around their numbers.  I implemented it from the paper for ARTS3 (the ARTS2 version does not make the error as far as I can see).

stefanbuehler:
Thanks for finding this and pointing it out. Is the correct equation written down somewhere (for ARTS3)? We could do an erratum for the paper, but I'm a bit afraid of the overhead that that process will likely cost, if we do it through the journal. Or perhaps we could make an errata section in the main .md file for ARTS 2.6?

PR #1163, opened 2026-09-01
DISORT/VDISORT overhaul

riclarsson:
This is a PR written mostly by AI (ChatGPT 5.6 Sol) with my guidance to ensure that existing and known tests are working (and that there is consistency between DISORT and VDISORT).  Below is what it considers done, missing, and requests from a review.

My main gripe is I need the VDISORT test suite.  I have set up some simple tests for known physics, like simple Fresnel-to-Brewster and Rayleigh at 90 deg.  But I do not have a wide enough range of tests to confirm if this is working in all ways we want, so there is no user facing VDISORT code.

Done:
- Validate CPP-DISORT against the supported Fortran DISORT reference suite. Latest version
- Extend VDISORT to reproduce the same 16 plane-parallel tests through its scalar embedding.
- Add arbitrary-angle radiances, unified flux results, DFDT, IMS/TMS corrections, delta-M preprocessing, and scalar delta-M-plus.
- Stabilize conservative and near-conservative scattering.
- Add polarized sources, boundaries, BRDFs, Fresnel and Cox–Munk surfaces, and combined surface models.
- Share common DISORT/VDISORT algorithms and BRDF kernels.
- Improve eigenmode validation, LAPACK error handling, caching, and allocation behavior.
- Expand analytic, reference, Python-interface, BRDF, and polarization tests.
- Document solver mathematics, matrix layouts, numerical behavior, validation scope, and known limitations.
- Reorganize the test and CMake structure.

Missing:
- Independent polarized VDISORT reference data for many-stream Mueller problems, reference-plane rotations, and polarized IMS/TMS.
- User-facing VDISORT workspace methods.
- The limiting solution and warning policy for beam–eigenmode resonance.
- General polarized delta-M-plus peak inference.
- Exact zero-thickness layer support.
- Additional conservative Jordan pairs beyond energy conservation.
- Off-grid ideal-Fresnel direct-beam reflection; Cox–Munk remains the beam-compatible alternative.
- Pseudo-spherical geometry, intentionally out of scope.

Requests:
- Standard polarized VDISORT benchmark cases with complete atmospheric, illumination, boundary, and output data.
- Many-stream Mueller-scattering cases with nonzero \(Q\), \(U\), and \(V\).
- Non-principal-plane cases that verify Stokes reference-plane rotations and the sign conventions for \(U\) and \(V\).
- Polarized IMS/TMS and delta-M cases with strongly forward-peaked scattering.
- Conservative polarized media, especially any physical example with more than one conserved mode.
- Direct-beam cases near beam–eigenmode resonance.
- Polarized surface cases for Fresnel, rough Fresnel/Cox–Munk, and mixed diffuse/specular reflection.
- Cross-platform numerical results for GCC, Clang, MSVC, and alternative LAPACK implementations.
- Realistic ARTS use cases that can guide the eventual user-facing VDISORT workspace interface.


olemke:
## HIGH — delta-M-plus end-to-end path corrupts memory

`DisortSettings::init()` (src/core/disort-cpp/disort.cpp:2191-2197) passes `NLeg = legendre_coefficients.ncols()` and `NLeg_all = legendre_polynomial_dimension` — **inverted** relative to `main_data` semantics (`NLeg` = truncated solve dim, `NLeg_all` = full moments, disort.h:153-157, ctor disort.h:275-281). Harmless on `main` where `ncols == L` always (verified: main's `resize()` allocates both to `legendre_polynomial_dimension`), but this PR:

1. relaxes `check()` to allow `ncols ≥ L` (disort.cpp:2177),
2. adds `disort_settingsDeltaMPlus` producing exactly `ncols = L+2` and `(F,N,L)` peak moments (m_disort_settings.cc:427-436; `delta_m_plus` returns `nleg` columns, disort.cpp:31).

Then `set()` (disort.cpp:2213-2214) assigns `[N,L]` into the `[N,L+2]` `delta_m_peak` view and `[N,L+2]` into the `[N,L]` `Leg_coeffs_all` view. Matpack view assignment only `assert`s shapes (matpack_mdspan_strided_view_t.h:147-150) → **abort in debug, out-of-bounds heap write in release**. Nothing catches it upstream, and no test runs the delta-M-plus agenda end-to-end (`delta-m-plus-settings.py` only checks the transform; the C++ Fortran tests use the direct constructor with the correct orientation).

**Fix:** swap the two arguments in `init()` (`NLeg = legendre_polynomial_dimension`, `NLeg_all = ncols`) and add an end-to-end test that runs `ScatteringSpeciesDeltaMPlus` → `disort_spectral_rad_fieldCalc`.

src/core/disort-cpp/disort.cpp
```diff
@@ -762,31 +1022,25 @@ void main_data::check_input_size() const {
 void main_data::check_input_value() const {
   ARTS_TIME_REPORT
 
-  ARTS_USER_ERROR_IF(tau_arr.front() <= 0.0, "tau_arr must be strictly positive, got {:B,}", tau_arr);
+  dc::check_layer_input_values(tau_arr, omega_arr, mu0, phi0);
 
-  ARTS_USER_ERROR_IF(stdr::any_of(omega_arr, [](auto&& omega) { return omega >= 1 or omega < 0; }),
-                     "omega_arr must be [0, 1), but got {:B,}",
-                     omega_arr);
```
olemke, line 769: The guard is simply gone? Is this deliberate?

riclarsson: The guard is in common.cpp and written as:

```cpp
  ARTS_USER_ERROR_IF(std::ranges::any_of(omega, [](const Numeric x) { return x < 0.0 or x > 1.0; }),
                     "omega_arr must be in [0, 1], got {:B,}",
                     omega);
```

The main difference is that omega close to 1 are solved in a different path.

src/core/disort-cpp/disort.cpp
```diff
@@ -762,31 +1022,25 @@ void main_data::check_input_size() const {
...
 
-  ARTS_USER_ERROR_IF(stdr::any_of(mu_arr[rf(N)], [mu = mu0](auto&& x) { return std::abs(x - mu) < 1e-8; }),
-                     "mu0 in mu_arr, this creates a singularity.  Change NQuad or mu0. Got mu_arr {:B,} for mu0 {}",
-                     mu_arr,
-                     mu0);
+  ARTS_USER_ERROR_IF(
+      stdr::any_of(delta_m_peak | by_elem, [](const Numeric x) { return !std::isfinite(x) || x < 0.0 || x > 1.0; }),
+      "delta_m_peak moments must be finite and [0, 1], got {:B,}",
+      delta_m_peak);
 }
```
olemke, line 790: While the m0 check above has moved to common.cpp, this singularity check is gone. Or is it now covered in some other way?

riclarsson: I do not understand?  The dc::check_layer_input_values call is made in check_input_values.  This is called in update_all.  update_all is where all the calculations happen.  That's where all the updates happened before.

olemke: `dc::check_layer_input_values` checks for tau, omega, mu0 and phi0. But the singularity check of `mu_arr` is not there or are anywhere else.
On main, the range check (old :784) and the coincidence check (old :786-789, std::abs(x - mu0) < 1e-8 against mu_arr) both lived in check_input_value. The new check_input_value (disort.cpp:1022-1044)  has neither the `mu_arr` comparison nor any call site that performs it.

riclarsson: Ah, OK, I misread.  I think it should be fine with overlapping sun and grids now.

olemke: I don't know the depths of the method well enough to judge that, so I leave the judgment call to you. Just want to mention that GLM seems to be vary about it (italic+bold highlight by me):

**Why that guard existed.** The beam's particular solution is obtained by solving with the shifted transport operator and then rescaling by the beam attenuation factor. When the beam direction coincides with a discrete-ordinate stream, the beam *resonates* with an eigenmode: an eigenvalue `K_j ≈ -1/mu0` makes the divisor `1 + K_j·mu0` vanish. The resonance is mathematically removable (L'Hôpital limit) but numerically it's catastrophic cancellation — the formula `jvec[j] *= mu0 / (1.0 + K[j]*mu0)` (disort.cpp:845, unchanged in this PR) amplifies roundoff without bound. The diagonal shift `1/mu0` added to the pair branch (disort.cpp:807) and in VDISORT (vdisort.cpp:1145-1146) has the same exposure.

**What the PR does.** The guard is simply gone. The new `check_input_value` (disort.cpp:1022-1044) validates Legendre coefficients, `I0`, `f_arr`, `delta_m_peak` — nothing checks `mu0` against `mu_arr`; the shared `dc::check_layer_input_values` (common.cpp:239-250) only range-checks `mu0 ∈ [0,1]`. **_So a user who happens to pick a solar zenith angle equal to a Gauss point (entirely reachable — users don't know the quadrature grid a priori) now gets silently wrong/huge numbers instead of the old clear error._**

**Why it's not a clean fix-already situation.** _**The PR *did* add analytic handling of removable singularities — but only in the correction kernels (`downward_tms_kernel` `|w|<1` branch, `ims_chi` `|z|<1` branch, covered by test 11d). The beam particular solve itself got no equivalent special-casing, no replacement validation, and no test pinning the new behavior.**_ The PR's own "Missing" list even concedes "the limiting solution and warning policy for beam–eigenmode resonance" is unhandled and requests "direct-beam cases near beam–eigenmode resonance" — so the authors know the regime is open, but the *removal* of the guard isn't called out in the description, which makes it look accidental rather than deliberate.

**Reasonable resolutions** (in increasing effort):
1. Restore the check (cheap, conservative — blocks legitimate work only in a tiny measure-zero-ish band), or
2. Keep removal but special-case `|1 + K_j·mu0| < tol` with the L'Hôpital limit like the TMS kernel does, or
3. Keep as-is but document the accepted resonance behavior and add a test asserting what currently happens.

riclarsson: OK, I do not have special workaround here so I will simply disable this.  I had to change quite a few test that produced expected results with mu0 in mu but fails because of this, but I do not wish to spend time investigating this

PR #1167, opened 2026-09-10
[ARTS-2.6] fix: correct sinefit baseline indexing in calcBaselineFit

olemke:
When a sine baseline fit (sinefit) retrieval quantity is used with more than one measurement block (`no_mblock_variation=0`), `x2artsSensor` reads the sin/cos coefficients from the wrong positions in the state vector. The second and later measurement blocks land half a block into the previous block's coefficients, so OEM receives a forward-model baseline that is inconsistent with the sinefit Jacobian columns — in practice, the sine baseline is never fitted for those blocks. This was reported and analysed in [discussion #1166](https://github.com/atmtools/arts/discussions/1166), along with a synthetic-data test case demonstrating the failure.

### Changes
- `src/jacobian.cc`: In `calcBaselineFit()`, the per-mblock and per-polarisation offsets into the state vector used the polyfit strides (one coefficient per pol/los/mblock), which is wrong for sinefit quantities that store a sin/cos pair per dimension. The per-los offset already used the factor 2, but the other two did not. The offsets now share a single `stride` (2 for sinefit, 1 for polyfit) for the mblock, los, and pol dimensions, making the layout consistent with `jacobianCalcSinefit`. This also fixes a previously shadowed wrong polarisation stride when `no_pol_variation=0`.

### Verification
- Running the reporter's `arts_sinefit_bug_demo.py` with the standard agenda now recovers the injected 30 mK / 50 MHz sine ripple in both measurement blocks (retrieved coefficients `[16.2, 25.0, 16.7, 25.4]` mK vs. expected `[16.2, 25.2, 16.2, 25.2]` mK, residual RMS 5.9/6.1 mK ≈ noise floor), matching the manual workaround discussed in #1166 to the decimal.
- Polyfit behaviour is unchanged, since its strides are identical to before.

### Breaking Changes
None — retrievals that avoided the bug (e.g. by applying the baseline manually via the Jacobian, or using a single measurement block) see identical results; retrievals affected by the bug now work as intended.


erikssonpatrick:
This looks all OK to me. Tests work and, yes, there was a bug in the code.

